PR #4, opened 2020-11-19
Add metadata presenter gem

tomas-stefano:
## Context

The pod is failing due to assets precompile.

Also there are not logs via k8s so adding the right env vars make work


brenetic (approved):
🚀

PR #5, opened 2020-12-02
Save user data to session

njseeto:
Story: https://trello.com/c/AJe4bfQK/1087-save-text-field-component-user-data-to-session-continue

We are currently not adding the tests as we should have different previews for the Editor that we currently do not have.
So it might be too early for this (happy to have a conversation about this if we think we can do some form of testing).


brenetic (approved):
🚀

PR #43, opened 2021-01-08
Generate service access token through fb jwt auth gem

tomas-stefano:
## Context

We had done the work to generate the access token in the gem PR https://github.com/ministryofjustice/fb-jwt-auth/pull/9 
Now we need to add the gem in the editor app.


brenetic (approved):
🚀

PR #52, opened 2021-01-13
Add encoded keys to config

brenetic:
This adds the encoded private and public keys to the config map and secrets as well as the metadata api url.

We also need to change the issuer name to be just 'editor' as the service token cache prefixes the name with 'fb-' for us.


njseeto (approved):
🚀

PR #62, opened 2021-01-14
Reference the page name locale correctly in form settings

brenetic:
Should reference the locale properly.

## Before

Screenshot 2021-01-14 at 18 56 19

## After

Screenshot 2021-01-14 at 18 56 09


njseeto (approved):
💯

tomas-stefano (approved):

PR #64, opened 2021-01-18
Acceptance tests

tomas-stefano:
## Context

This PR adds the acceptance tests for the editor.

The base idea is to have tests where we can run locally against:

1. local editor
2.  or against any editor environment (given a editor url).

### ENV vars

Two env files are manipulated into one file:

1. one for running with local containers
2. other pointing to the test environment (but it can be changed to any other environment if we like)

### Local editor

Added all necessary containers for the editor

- editor-db
- editor-app
- metadata-db
- metadata-app
- service-token-cache-app
- service-token-cache-redis

This will spin up the containers and run the tests against the editor. A future improvement is to transform the editor into a volume to make us more productive.

### CI

To run this you need to have the user and password setup on CircleCI (I already added there) or if you run on your local machine that points to the test environment you can as well. If you face 401 on the tests it is because the user env var and pass aren't set.

### Caveats

I have to make a specific dockerfile for acceptance test because we need chrome driver and other packages to run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,10 +12,13 @@ jobs:
       - setup_remote_docker:
           version: 19.03.13
       - run:
+          name: setup
+          command: make setup
+      - run:
           name: security
           command: docker-compose run --rm editor-app bundle exec brakeman -q --no-pager
       - run:
-          name: test
+          name: unit tests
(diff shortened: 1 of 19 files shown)


njseeto (approved):
🚀

PR #66, opened 2021-01-19
Add single question page

brenetic:
## Add Single Question Page with text component

This adds the functionality to create a single question page with a text component and add it to an existing service metadata.

At this point we just redirect to a read only edit page.

## Deep duplication of the default metadata in initializer …

Because Ruby is referencing the initialized default metadata objects, these are mutated whenever we create a new page or component object. Obviously this is not what we want.

Therefore we do a deep duplication whenever accessing the default metadata for a required page or component.

## Automatically generate a new component ID

We need to automatically generate a unique component ID and name whenever creating a new component. This will allow it to be referenced by name in the future and for us to be sure that the correct field is being pointed to.

## Add acceptance tests to create page

Removed the dropdown for now. Add unit tests validation for page creation when user add '/'

## Change MetadataApiClient Version parameter to be payload …

`payload` better describes what the actual parameter is going to be used for

https://trello.com/c/5hSTER5h/1168-add-single-question-page-with-text-component


njseeto (approved):
🚀  looks great!

PR #74, opened 2021-01-22
Feature/edit start page

tomas-stefano:
[Trello card](https://trello.com/c/dta3Fs4i/1199-backend-editing-start-page)

## Context

This PR contemplates:

1. Renders the page on the edit page.
2. Add the current properties of the page as hidden inputs
3. The submitted form will be created a new version in the metadata API with the new attributes.

## Observations of the implementation

We add the [Hashie](https://github.com/hashie/hashie#deeplocate) gem to use the deep locate when updating the metadata API. Find the object to be updated passing its id to locate.

We hardcoded the key 'pages' for now but in the future the MetadataUpdater should be responsible to update other nodes as the configuration or others.

## Showing the feature of this PR

The gif shows the page being rendered and the current attributes being created on the API.

edit-start-page

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
           command: make setup
       - run:
           name: security
-          command: docker-compose run --rm editor-app bundle exec brakeman -q --no-pager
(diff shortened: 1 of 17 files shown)


brenetic (approved):
🚀

PR #78, opened 2021-01-25
Use generated _uuid property in pages

brenetic:
When:

1. Create a service (the start page uuid)
2. Create a page (the page uuid)

Use metadata presenter 0.3.0 which uses the UUID in the page path. As a result of this version we no longer need to override the reserved_answer_path as it's just a post to the page URL now.

https://trello.com/c/wbZcslTs/1224-use-generated-page-id-instead-of-name


tomas-stefano (dismissed):
🎸

njseeto (approved):
🚀

PR #80, opened 2021-01-26
Add page uuid call to page creation

tomas-stefano:
## Context

The current test environment is raising an error when creating pages. 
```NoMethodError (undefined method `page_uuid' for #<PageCreation:0x0000558662cdcb78>```

This fixes the test environment.

## Caveats

Page creation requires to redirect the user to edit the page when the page is created. So this PR returns the last page UUID.
This will be a problem when we need to create a page and insert in the middle of the service but it solves the problem now.


brenetic (approved):
🚀

PR #96, opened 2021-02-01
Publisher/part 2 env vars

tomas-stefano:
[Trello card](https://trello.com/c/M8ERFVPZ/1200-add-publish-mechanism-part-2)

## Context

This PR adds:

1. Generate the private-public key pair for a service when is created.
2. Pass the metadata as env var to the runner.
3. Add the service account token as env var
4. Add publishing and post publishing step to the cloud platform adapter.

## Next steps (will be done in different PR)

- Show the hostname after the publish is complete
- Create a worker dockerfile
- Add delayed job or sidekiq


brenetic (approved):
🚀

PR #107, opened 2021-02-05
Publish/username password

tomas-stefano:
## Context

This adds the implementation of the publish a service sending the username and password and injecting the username into the service pods on publishing.

We split the service configuration to generate into two different Kubernetes configuration:

1. secrets.yaml
    - The encoded private key
    - The username
    - The password
2. configmap.yml
   - The public key (so service token cache can authenticate the service)


brenetic (approved):
🚀

PR #111, opened 2021-02-08
Encrypt username and password for basic authentication

brenetic:
This adds the username and password for basic authentication as well as
encrypting all the service configuration values in the DB.

This includes the private and public keys. However the public key which
resides in the config still needs to be base64 encoded when written to
the config_map.yaml. This is because the datastore ServiceTokenCacheClient
does a base64 decode after retrieving the key. The service token cache
just looks for the key 'ENCODED_PUBLIC_KEY' in the config map and passes
back to the requesting app.

The 'ENCODED_' part of both keys relates to the time they are written to
the yaml files, _not_ when they are written to the DB.

https://trello.com/c/9BnRRaD5/1228-publish-mechanism-add-username-and-password

### DIFF
--- a/app/views/publish/index.html.erb
+++ b/app/views/publish/index.html.erb
@@ -17,6 +17,21 @@
 <%= form_for(@publish_service_creation_dev, url: publish_index_path(service.service_id)) do |f| %>
   <%= render 'form', f: f, deployment_environment: 'dev' %>
 
+  <div class="govuk-form-group">
+    <%= f.check_box :require_authentication %>
+    <%= f.label :require_authentication %>
+  </div>
+
+  <div class="govuk-form-group">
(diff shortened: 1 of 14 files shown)


app/views/publish/index.html.erb
```diff
@@ -17,6 +17,21 @@
...
+  <div class="govuk-form-group">
+    <%= f.label :username %>
+    <%= f.text_field :username %>
+  </div>
+
+  <div class="govuk-form-group">
+    <%= f.label :password %>
+    <%= f.text_field :password %>
+  </div>
+
```
tomas-stefano, line 34: The inputs are duplicated. I'll remove it.

tomas-stefano (approved):

PR #113, opened 2021-02-08
Add encrypt env vars to web k8s app

tomas-stefano:
## Context

The env vars were added on the worker deployment, we should also add on the web deployment,


brenetic (approved):
🚀

PR #115, opened 2021-02-09
Fix/runner

tomas-stefano:
## Context

We need to explicitly send to the runner serve the assets when publishing in order to make work. This is only for the MVP as we need a plan to serve the assets in a CDN.

The rails to log to stdout is because when do `kubectl logs` or `kubetail` the runner were not logging.


brenetic (approved):
🚀

PR #122, opened 2021-02-10
Add encryption key and salt for development

tomas-stefano:
## Context

The encryption env vars were added and we need them in the development environment. Just added a random generated one.


brenetic (approved):
🚀

PR #136, opened 2021-02-16
Accept components as a JSON when updating a page

tomas-stefano:
After the conversation with the frontend we agreed that the components
attribute will be a JSON object that the JS will manipulate and send
it back when user click save along with the other properties as non
JSON.


spec/controllers/pages_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      it 'parses components as json' do
+        expect(controller.page_update_params).to eq({
+          'id' => page.id,
+          'latest_metadata' => service_metadata,
+          'components' => [component_params],
+          'service_id' => service.service_id
+        })
+      end
```
brenetic, line 34: is it worth testing with two sets of components at this point or wait until the multiple question page is added?

brenetic (approved):
🚀

spec/controllers/pages_controller_spec.rb
tomas-stefano: Yes I think it is worth to test. Added now

PR #144, opened 2021-02-16
Fix form navigation bar position and focus style

stevenburnell-moj:
**Fix form navigation bar position and focus style**

**Was (with focus)**
Screenshot 2021-02-16 at 23 43 58

**Now (with focus)**
Screenshot 2021-02-16 at 23 44 14


brenetic (approved):
🚀

PR #145, opened 2021-02-17
Feature/1219 frontend editing start page

stevenburnell-moj:
Page elements now editable by user with translation to/from html/markdown on large content areas. 
Save button disabled until some detected change in text (even just a space).

Note: Editing will only work with target classes added to
fb-metadata-presenter/app/views/metadata_presenter/page/start.html.erb
A separate PR will be raised for that.

**Was (without editing ability)**
Screenshot 2021-02-16 at 23 59 18

**Now (with editing)**
Screenshot 2021-02-17 at 00 00 56


acceptance/pages/editor_app.rb
```diff
@@ -40,4 +39,13 @@ def set_component_type_field(value)
     page.execute_script("document.getElementById('page_component_type').value = '#{value}'")
   end
 
+  def submit_new_page
+    #find('#new_page input[type="submit"]', visible: false)
```
brenetic, line 43: do we need to keep this in?

app/javascript/src/editable_components.js
```diff
@@ -0,0 +1,315 @@
...
+    $node.on("click.editablecomponent", this.edit.bind(this));
+    $node.on("blur.editablecomponent", this.update.bind(this));
+  }
+
+  edit() {
+    console.log("edit");
+  }
+
+  update() {
+		console.log("update");
```
brenetic, line 54:
```suggestion
	console.log("update");
```

🙈

acceptance/pages/editor_app.rb
stevenburnell-moj: No, I should have deleted when replacement line was working. Will remove.

tomas-stefano (approved):

PR #162, opened 2021-02-25
Fix incomplete previous attempt to remove errors on close

stevenburnell-moj:
Errors were visible after closing and reopening a dialog.
Previous fix only solved when 'cancel' button was clicked.
This should fix the issue no matter how the dialog is closed.


brenetic (approved):
🚀

PR #180, opened 2021-03-06
Feature/1348 frontend edit single question page radio button component

stevenburnell-moj:
Add Editable Component code to support Single Question Radio fields. Will need update to metadata presenter view for this to kick in.

Screenshot 2021-03-06 at 01 15 17


brenetic (approved):
🚀

PR #186, opened 2021-03-09
Login page and Footer

brenetic:
## Login page

Created the login page layout and content. Slack channel will be injected later. The sign in link just takes the user to the services page currently since they have already gotten passed basic auth.

## Header navigation

The forms link and sign out link are now there. The latter does a faux session destroy before redirecting to the login page. It only appears if not on the login page but will later be changed to check whether a user is logged in or not.

## Add footer

This currently does not contain any accessibility, privacy or cookies links. These will be added later.

## Screenshots

Screenshot 2021-03-08 at 21 49 41

Screenshot 2021-03-08 at 21 49 52

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,2 +1,5 @@
 class HomeController < ApplicationController
+  def show
+    @sign_in_url = '/services'
+  end
 end
--- a/app/helpers/auth0_helper.rb
+++ b/app/helpers/auth0_helper.rb
@@ -15,7 +15,8 @@ module Auth0Helper
   # Is the user signed in?
   # @return [Boolean]
   def user_signed_in?
-    session.try(:[],:user_id).present?
+    request.path != '/'
+    # session.try(:[],:user_id).present?
   end
 
   def identify_user(user_id = session[:user_id])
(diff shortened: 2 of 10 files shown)


app/controllers/home_controller.rb
```diff
@@ -1,2 +1,5 @@
 class HomeController < ApplicationController
+  def show
+    @sign_in_url = '/services'
```
tomas-stefano, line 3: Can we change to `services_path`?

app/helpers/auth0_helper.rb
```diff
@@ -15,7 +15,8 @@ def login_path
   # Is the user signed in?
   # @return [Boolean]
   def user_signed_in?
-    session.try(:[],:user_id).present?
+    request.path != '/'
```
tomas-stefano, line 18: Is this a temporary solution?

brenetic: this is just to separate out the styling from the login functionality. it will change as soon as the auth0 pr is raised

app/controllers/home_controller.rb
brenetic: this is also temporary and will become:

```
    if Rails.env.development?
      @sign_in_url = '/auth/developer'
    else
      @sign_in_url = '/auth/auth0'
    end
```

tomas-stefano (approved):

tomas-stefano (approved):

PR #192, opened 2021-03-09
Email Submission Configuration

njseeto:
This PR relates to this story: https://trello.com/c/utAUt3qG/1253-add-submission-configuration-serviceoutputemail-to-editor

This PR contains changes including:
#### Adds layout of pages
Add html.erb files in `views/settings` folder and associated labels in `config/en.yml`

This allows the user to go to the Settings -> Submission settings -> Send by email -> Send by email forms page

#### Puts every setting into a specific controller
This meant moving logic out of `app/controllers/settings_controller.rb` and splitting it out into different controllers, `app/controllers/settings/email_controller.rb` `app/controllers/settings/form_information_controller.rb` `app/controllers/settings/submission_controller.rb`

The controllers now had their own index.html.erb files:
Settings -> index.html.erb

/settings/form_information
Controller: Settings - Form information -> index.html.erb

/settings/submission
Controller: Settings - Submission -> index.html.erb

/settings/submission/email
 Controller: Settings - Email -> index.html.erb

This meant changes to the `config/routes.rb` `spec/requests/settings_spec.rb` and paths in associated `index.html.erb` files.

#### Sets all 5 possible email setting configurations in both dev and production environments
`app/models/email_settings.rb`
This is because there are 3 fields (subject, heading pdf heading) that require a default, whereas the output and pdf subheading can be left blank.

#### Make inputs return from params, db or default
`app/controllers/settings/email_controller.rb`
`app/models/email_settings.rb`
And add associated tests
`spec/factories/service_configurations.rb`
`spec/models/email_settings_spec.rb`

#### Do not inject email env vars if send email flag is false
If the user unchecks the "Send by email on Test/Live" then a record in the db will be created (SubmissionSetting) then when we publish we should reject the email related env vars for those that doesn't have a send email flag as true.
`app/models/service_configuration.rb`
`app/services/publisher/service_provisioner.rb`
And associated tests:
`spec/models/service_configuration_spec.rb`

#### Add submission setting table
This table will be responsible to handle the flag that the service needs to send email or not.
Usually every service will have two rows on this table.
`app/models/submission_setting.rb`
`db/migrate/20210304150513_create_submission_settings.rb`
`db/schema.rb`
And spec files:
`spec/factories/submission_settings.rb`
`spec/models/submission_setting_spec.rb`

### DIFF
--- a/acceptance/pages/editor_app.rb
+++ b/acceptance/pages/editor_app.rb
@@ -18,6 +18,9 @@ class EditorApp < SitePrism::Page
   element :form_name_field, :field, 'Form name'
   element :save_button, :button, 'Save'
 
+  element :submission_settings_link, :link, 'Submission settings'
+  element :send_by_email_link, :link, 'Send by email'
+
   element :page_url_field, :field, 'The page’s relative url - it must not contain any spaces', visible: false
   element :new_page_form, '#new_page', visible: false
 
--- a/app/controllers/settings/email_controller.rb
+++ b/app/controllers/settings/email_controller.rb
@@ -1,0 +1,50 @@
+class Settings::EmailController < FormController
+  before_action :assign_form_objects
+
+  def create
+    @email_settings = EmailSettings.new(
+      email_settings_params.merge(service: service)
+    )
+
+    if @email_settings.valid?
+      EmailSettingsUpdater.new(
+        email_settings: @email_settings,
+        service: service
+      ).create_or_update!
+
+      redirect_to settings_email_index_path(service_id: service.service_id)
+    else
+      if email_settings_params[:deployment_environment] == 'dev'
+        @email_settings_dev = @email_settings
+      else
+        @email_settings_production = @email_settings
+      end
+      render :index, status: 422
+    end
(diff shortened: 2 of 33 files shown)


acceptance/pages/editor_app.rb
```diff
@@ -29,6 +32,14 @@ def modal_create_service_button
     all('.ui-dialog-buttonpane button').first
   end
 
+  def configure_dev_link
+    find('#configure-dev').click
+  end
```
tomas-stefano, line 37: This method can be deleted. It is not being used anymore :)

acceptance/pages/editor_app.rb
```diff
@@ -29,6 +32,14 @@ def modal_create_service_button
...
+
+  def send_email_to_field
+    all('#email-output-dev').first
+  end
```
tomas-stefano, line 41: Also this method can be deleted too

tomas-stefano (commented):
Awesome work Nat! Just minor comments but everything else looks good to me.

PR #194, opened 2021-03-09
Rename page answer presenter

tomas-stefano:
## Context

When you add a check your answers page the editor returns a 500 error. The reason is that the check your answers page renamed the method needed to render its answers.

So the editor needs to be renamed as well.


brenetic (approved):
🚀

PR #199, opened 2021-03-09
Disable acceptance tests

brenetic:
These will never pass for the moment as we have not updated them to reflect the changes in the UI.

We will re-enable them a little later when the UI is more settled.


tomas-stefano:
I am closing this in favor of the ticket that I am working. We will have a passing acceptance tests soon! 🎸

PR #204, opened 2021-03-11
Re enable Auth0 authentication

brenetic:
## Re-enable the use of Auth0 …

This puts back auth0 and adds some tests which will now look for specific users instead of the generic one.

Also create a PermissionsController which is the thing to check for a logged in user. This is done with the potential for policies and permissions in the future.

## Format username correctly …

The prototype requires the user name to take the form of first initial and last name.

Display it correctly in the header navigation.

## Encrypt user details in DB


tomas-stefano (approved):
🎸

PR #207, opened 2021-03-11
Add delete pages feature

tomas-stefano:
[Trello card](https://trello.com/c/vWF85IUT/1371-user-can-delete-remove-a-page-backend-m)

## Context

The delete pages will delete the 'pages' node in the metadata and
the pages[0]['steps'] in metadata as well.

Example

id => 'page.name'

'page.name' will be deleted entirely from the 'pages' node and
the pages[0]['steps'] node too.

The exception is that we should not delete the start page
(see the discusssion here:
https://mojdt.slack.com/archives/CE78LKCU8/p1615481002253100)


brenetic (approved):
🚀

PR #208, opened 2021-03-11
Add link to ask form builder slack channel

brenetic:
This was previously going to be injected as secret into the app, but actually the provided link requires people to authenticate before they can get to the channel so it's probably fine to just have in the locales.


tomas-stefano (approved):
:shipit:

PR #210, opened 2021-03-12
Removes the "current" underline in the nav on Editing pages

f-marry:


stevenburnell-moj (commented):
This looks like it would also remove the 'current' underlining from the Flow page as well, which I'm guessing is not the intended effect. Assume you just want it removed from navigation when you're editing individual page. Is that right?

f-marry:
Hi, you're correct about what I'm trying to achieve. I thought that code would still leave the current on the flow page,

stevenburnell-moj (approved):
Just pulled down the code to check this one out and it looks to me like it is having the desired effect, despite previous comment. Marking it as 'approved' because we can always fix later if need be.

PR #219, opened 2021-03-15
Tests/acceptance

tomas-stefano:
## Context

This PR adds:

1.  the acceptance tests for adding pages, add a login step (that works for the editor locally).
2. The docker-compose development. I add this because if you change the editor code locally it should reflect to the editor in the container.

See this PR more like: organising the specs and making work with the new frontend.

## Caveats

* The tests are still failing because of the footer. If you delete the footer in both app/views/layouts/application.html.erb and app/views/layouts/form.html.erb then the tests will pass locally.
* The xpath used in the page object in the acceptance tests was added because it tries to imitate the dropdown behaviour of click, hover and click respectively.

## Next steps

- [ ] Add the edit (ALL pages!) feature specs
- [ ] Solve the footer problem
- [ ] Make the authentication work when running against the test environment


brenetic (approved):
🚀

PR #221, opened 2021-03-16
Add all tests for editing single question pages

tomas-stefano:
## Context

This adds the edit the single question acceptance tests

## What these tests do not cover yet?

* The add option on radio and checkboxes is not covered because is in progress at the moment.


brenetic (approved):
🚀

PR #231, opened 2021-03-18
Add session rake task to readme

njseeto:
This has been added to avoid confusion, we want to be explicit in saying we are not using the built in rails trim task.

Story: https://trello.com/c/0CceGvhN/1385-session-duration-for-editor


brenetic (approved):
🚀

PR #238, opened 2021-03-19
Fix lost page type after error

stevenburnell-moj:
(copied from ticket)

On form 'flow' page, the default set type is Single Question (written in hidden form fields on page load). This default is overwrites the chosen type after user error.

Steps to replicate:

1. Visit form 'flow' page.
2. Click 'Add page' (but not Single Question type).
3. Enter incorrect information in Dialog.
4. Submit.
5. Page is sent to server and reloads with error.
6. Correct error and submit.
7. Page is created but of Single Question type, not the alternative you chose in step 2.


brenetic (approved):
🚀

PR #239, opened 2021-03-19
Remove DS_Store

brenetic:
Mac specific file. Shouldn't be checked in


stevenburnell-moj:
Haha! The other day I added a .gitignore line (which I never got round to committing) that must have been influenced by the appearance of this file. Might have to revisit that action now.

stevenburnell-moj (approved):

PR #241, opened 2021-03-19
Add Sentry Configuration 

brenetic:
## Add Sentry configuration

This requires three gems as we also have delayed job queues for publishing.

## Remove SENTRY_CURRENT_ENV 

As a result of having 4 different namespace environments as opposed to the regular 2 we structure our sentry projects differently.

We have 2 sentry projects per app, one for test and one for live. The test one covers anything in the test-dev, test-production and live-dev namespaces whereas live is exclusively for live-production only. Therefore it's fine to let SENTRY_CURRENT_ENV use whatever RAILS_ENV is set to as it does by default.

## Send alerts to sentry 

Currently just set to capture errors when creating services or versions or when an error occurs in the rake task that cleans inactive sessions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,6 +28,9 @@ gem 'rails', '~> 6.1.3'
 gem 'sass-rails', '>= 6'
 gem 'tzinfo-data'
 gem 'webpacker', '~> 5.2'
+gem 'sentry-ruby', '~> 4.3.0'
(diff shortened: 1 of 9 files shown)


mattempty (approved):
🚀

njseeto (approved):
💯

PR #247, opened 2021-03-23
Add service Sentry DSN's during publishing

brenetic:
We need to inject Sentry DSN's into the configuration for all the published services.

test-dev, test-production and live-dev all use the same DSN which sends exceptions to the test project in Sentry.

live-production has its' own project in Sentry.


njseeto (approved):
🚀

PR #248, opened 2021-03-23
Fix/editable date component

stevenburnell-moj:
Date component was not saving updates due to JS error and then saving incorrect value for label.


tomas-stefano:
Approved. We have an acceptance tests for this already in https://github.com/ministryofjustice/fb-editor/blob/main/acceptance/features/edit_single_question_page_spec.rb#L40
Was it failing before this PR?

tomas-stefano (approved):

PR #249, opened 2021-03-23
Fix/acceptance tests

tomas-stefano:
The content of some fields has changed. unfortunately, the acceptance tests are not using the locales and doing this could take a while to be done so changing on the specs instead.


brenetic (approved):
🚀

brenetic (approved):
🚀

PR #250, opened 2021-03-23
Escape service metadata

njseeto:
Story: https://trello.com/c/7FaYyHxY/1406-investigate-editor-publishing-failure

There is a publishing failure when the service metadata has a `'` character.
This PR addresses this issue by escaping the service metadata.

Co-authored by: Tomas Destefi <tomas.destefi@digital.justice.gov.uk>

### DIFF
--- a/app/services/publisher/service_provisioner.rb
+++ b/app/services/publisher/service_provisioner.rb
@@ -20,7 +20,7 @@ class Publisher
     LIVE_PRODUCTION = 'live-production'.freeze
 
     def service_metadata
-      service.to_json
+      service.to_json.inspect
     end
 
     def get_binding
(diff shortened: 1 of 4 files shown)


app/services/publisher/service_provisioner.rb
```diff
@@ -20,7 +20,7 @@ class ServiceProvisioner
     LIVE_PRODUCTION = 'live-production'.freeze
 
     def service_metadata
-      service.to_json
+      service.to_json.inspect
```
brenetic, line 23: do we want to keep the `inspect` in there?

njseeto: I think so, when passed into the deployment.yaml.erb file it doesn't escape properly without the inspect and the test fails :(

brenetic (approved):
🚀

PR #256, opened 2021-03-24
Make sure we are using the right names for editor service

tomas-stefano:
For the service monitor it is best to make sure we are following [CP docs](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/monitoring-an-app/application-metrics.html#applications-configured-to-use-multiple-processes)

We need to make sure that we are using the same pod definition name in other places. 

We do have the `fb-editor-web` and `fb-editor-workers` pods so let's use this everywhere.


njseeto (approved):
🚀

PR #257, opened 2021-03-24
Remove sentry alerting when creating a service

tomas-stefano:
When someone tries to create a service with the same name of a
existing one, the sentry is alerted without need. So removing on this
commit.


brenetic (approved):
🚀

PR #259, opened 2021-03-24
Make acceptance tests to pass again

tomas-stefano:
The radio/checkboxes js component changed name so we
had to change.

The metadata presenter is not included in the spec helper
so we had to add the require


njseeto (approved):
🚀

PR #261, opened 2021-03-25
Deny access to metrics endpoint

brenetic:
We do not want unauthorised access to the metrics endpoint. There is no need for the request to even get as far as the application.

Prometheus seems to go directly to the container instead of using the ingress.


njseeto (approved):
🚀

PR #275, opened 2021-03-30
Fix colour and margin in confirmation panel

stevenburnell-moj:
**Was**
Screenshot 2021-03-30 at 11 03 51

**Now**
Screenshot 2021-03-30 at 11 04 31


brenetic (approved):
🚀

PR #276, opened 2021-03-30
Make add page after works

tomas-stefano:
## Context

Insert a page after a given page.

The way it works is:

Given a params from the form called `page[add_page_after]` with the uuid of the page, the new page will be add after the given page in the params as expected.

### DIFF
--- a/app/generators/new_page_generator.rb
+++ b/app/generators/new_page_generator.rb
@@ -1,11 +1,15 @@
 class NewPageGenerator
   include ActiveModel::Model
-  attr_accessor :page_type, :page_url, :component_type, :latest_metadata
+  attr_accessor :page_type,
+                :page_url,
+                :component_type,
+                :latest_metadata,
+                :add_page_after
 
   def to_metadata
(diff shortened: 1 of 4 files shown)


app/generators/new_page_generator.rb
```diff
@@ -34,4 +38,22 @@ def component
       page_url: page_url
     ).to_metadata
   end
+
+  INSERT_PAGE_IN_LAST = -1
+
+  def insert_page_on
```
brenetic, line 44:
```suggestion
  def insert_page_at
```

app/generators/new_page_generator.rb
```diff
@@ -34,4 +38,22 @@ def component
       page_url: page_url
     ).to_metadata
   end
+
+  INSERT_PAGE_IN_LAST = -1
```
brenetic, line 42:
```suggestion
  INSERT_LAST = -1
```

brenetic (approved):
🚀

PR #277, opened 2021-03-30
Don't show the delete page if is the start page

tomas-stefano:
[Trello card](https://trello.com/c/vFcrTNto/1412-remove-delete-function-from-3dot-menu-for-start-page)

## Context

On the backend we don't allow start pages to be deleted so the "Delete page" on start page is useless.


brenetic (approved):
🚀

PR #278, opened 2021-03-30
Add rubocop

tomas-stefano:
## Context

Add rubocop and fix 80 to 90% of offences. The other 10% is on the rubocop todo file.


app/controllers/settings/email_controller.rb
```diff
@@ -19,11 +19,11 @@ def create
       else
         @email_settings_production = @email_settings
       end
-      render :index, status: 422
+      render :index, status: :unprocessable_entity
     end
   end
 
-  private
+private
```
brenetic, line 26: i actually like this as it's what i've done in other places i've worked. however it's not something we do here on any of our repos, so maybe we shouldn't start now

brenetic (approved):
🚀

PR #285, opened 2021-03-31
Add user authentication to preview feature

tomas-stefano:
[Trello card](https://trello.com/c/Ng5jvVI5/1414-add-auth-to-all-pages-for-using-the-editor)

## Context

This PR adds the authentication check on preview form/preview pages.

## Solution

The preview comes from the metadata presenter gem and has a superclass inside of the editor app (PreviewController).

So by making the preview controller to inherit from permission controller we make the user to be redirect to the login path

I had to chage the `root_path` because it makes the user to redirect to root path in the in the context of the metadata
presenter gem (/services/:id/preview) which creates a infinite redirection.

Adding the FbEditor::Application.url_helpers.root_path makes it return '/' which is what we want.

### DIFF
--- a/app/controllers/preview_controller.rb
+++ b/app/controllers/preview_controller.rb
@@ -1,3 +1,3 @@
-class PreviewController < ApplicationController
+class PreviewController < PermissionsController
   layout 'metadata_presenter/application'
 end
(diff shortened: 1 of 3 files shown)


njseeto (approved):
🚀

PR #287, opened 2021-03-31
Make readme a little clearer

njseeto:
Changed as it is slightly confusing as it can be read as running the commands in the metadata-api rather than in the fb-editor.


brenetic (approved):
🚀

PR #291, opened 2021-03-31
Fix duplicated Editable Collection Item text by using default text

stevenburnell-moj:
Fixes the current clone issue (duplicated text) when adding Radio options and Checkboxes.

Previously we saw duplicated text from first (cloned) element:
Screenshot 2021-04-01 at 10 10 50

New items will look like this (checkbox example, radios are same text):
Screenshot 2021-03-31 at 23 31 48

### DIFF
--- a/app/views/partials/_properties.html.erb
+++ b/app/views/partials/_properties.html.erb
@@ -5,6 +5,8 @@
       controller:  "<%= controller_name %>"
     },
     text: {
+      default_option: "<%= t('default_text.option') %>",
+      default_option_hint: "<%= t('default_text.option_hint') %>",
       dialogs: {
         button_cancel: "<%= t('dialogs.button_cancel') %>",
         button_delete: "<%= t('dialogs.button_delete') %>",
--- a/app/javascript/src/editable_components.js
+++ b/app/javascript/src/editable_components.js
@@ -493,19 +493,23 @@ class EditableCollectionFieldComponent extends EditableComponentBase {
  * because it would alert us to something very wrong.
  **/
 EditableCollectionFieldComponent.createCollectionItemTemplate = function(config) {
-  var $item = this.$node.find(config.selectorCollectionItem).eq(0);
+  var $clone = this.$node.find(config.selectorCollectionItem).eq(0).clone();
   var data = mergeObjects({}, config.data, ["items"]); // pt.1 Copy without items.
   var itemConfig = mergeObjects({}, config, ["data"]); // pt.2 Copy without data.
   itemConfig.data = mergeObjects(data, config.data.items[0]); // Bug fix response to JS reference handling.
 
   // Filters could be changing the blah_1 values to blah_0, depending on filters in play.
   itemConfig.data = EditableCollectionFieldComponent.applyFilters(config.filters, 0, itemConfig.data);
-  $item.data("config", itemConfig);
+
+  // In case we need some custom actions on element.
+  safelyActivateFunction(config.onCollectionItemClone, $clone);
+
+  $clone.data("config", itemConfig);
 
   // Note: If we need to strip out some attributes or alter the template
   //       in some way, do that here.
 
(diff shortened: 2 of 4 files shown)


app/views/partials/_properties.html.erb
```diff
@@ -5,6 +5,8 @@
       controller:  "<%= controller_name %>"
     },
     text: {
+      default_option: "<%= t('default_text.option') %>",
+      default_option_hint: "<%= t('default_text.option_hint') %>",
```
tomas-stefano, line 9: I think we have already one of them available `MetadataPresenter::DefaultText[:hint]`.
Maybe we should put the other value in the gem? @brenetic do you agree?

brenetic: not sure i'm following why this was an issue?

Screenshot 2021-04-01 at 09 44 39

it's ok for the text to be the same as the editor is going to change them or remove them

stevenburnell-moj: If we can alter the template that outputs the radio and checkbox items so that we have something like this:

```
<div class="govuk-checkboxes" data-module="govuk-checkboxes">
    <div class="govuk-checkboxes__item EditableComponentCollectionItem">
        <input id="answers-checkbox-checkboxes-1-option-2-field" class="govuk-checkboxes__input" aria-describedby="answers-checkbox-checkboxes-1-option-2-hint" type="checkbox" value="Option 2" name="answers[checkbox_checkboxes_1][]" disabled="disabled">
        <label for="answers-checkbox-checkboxes-1-option-2-field" class="govuk-label govuk-checkboxes__label EditableElement" contenteditable="true" data-fb-default-text="[Required option text]">Option 2</label>
        <span class="govuk-hint govuk-checkboxes__hint EditableElement" id="answers-checkbox-checkboxes-1-option-2-hint" contenteditable="true" data-fb-default-text="[Optional hint text]">[Optional hint text]</span>
    </div>
    <div class="govuk-checkboxes__item EditableComponentCollectionItem">
        <input id="answers-checkbox-checkboxes-1-option-field" class="govuk-checkboxes__input" aria-describedby="answers-checkbox-checkboxes-1-option-hint" type="checkbox" value="Option" name="answers[checkbox_checkboxes_1][]" disabled="disabled">
        <label for="answers-checkbox-checkboxes-1-option-field" class="govuk-label govuk-checkboxes__label EditableElement" contenteditable="true" data-fb-default-text="[Required option text]">Option</label>
        <span class="govuk-hint govuk-checkboxes__hint EditableElement" id="answers-checkbox-checkboxes-1-option-hint" contenteditable="true" data-fb-default-text="[Optional hint text]">[Optional hint text]</span>
    </div>
</div>

```

Then I should be able change the JS to pick up from those element. Not sure on effort as untested though.
(see required data attributes on label and span)
```
data-fb-default-text="[Required option text]"
data-fb-default-text="[Optional hint text]"
```

Alternative is to add these attributes (with different names, obviously) to the outer component, but I think that might be harder and is still deviating from the method other elements are using, so we might not be gaining much.

E.g.

```
<div class="govuk-checkboxes" data-module="govuk-checkboxes" data-fb-default-label-text="[Required option text]" data-fb-default-hint-text="[Optional hint text]">
    <div class="govuk-checkboxes__item EditableComponentCollectionItem">
    ...
    </div>
</div>
```

stevenburnell-moj: Note: Lines 73 + 74 of app/javascript/src/page_editable_content.js can be adapted to find values from data-fb-default-text attributes of the <label> and <span> elements, if we figure out how to add them through the fb-metadata-presenter/app/views/metadata_presenter/component/_radios.html.erb (and _checkboxes.html.erb) templates.

stevenburnell-moj:
Closing because time spent in PR has caused conflicts which are easier to resolve by creating a new branch instead.
See PR https://github.com/ministryofjustice/fb-editor/pull/305

PR #304, opened 2021-04-07
Format user display name based on connection type

brenetic:
The google and azure user claim set have slightly different structures
to their names. The latter uses surname first.

However both their nicknames have similar structures of
first_name.last_name. So swap to use that and then fallback to the
email.

Also remove any numbers as some of the azure email addresses have
numbers at the end.


tomas-stefano (approved):
🚀

PR #305, opened 2021-04-07
Fix/1429 default text for optional fields 2

stevenburnell-moj:
This is work that was in PR#291 but that work became outdated due to time spent in PR. 
Rather than force-push the branch, have simply created new branch, updated to latest main and raised this new PR.

See https://github.com/ministryofjustice/fb-editor/pull/291 for description of this work.


app/views/partials/_properties.html.erb
```diff
@@ -5,6 +5,8 @@
       controller:  "<%= controller_name %>"
     },
     text: {
+      default_option: "<%= t('default_text.option') %>",
+      default_option_hint: "<%= t('default_text.option_hint') %>",
```
tomas-stefano, line 9: I would suggest to change this lines to:

```
default_option: MetadataPresenter::DefaultText[:option],
default_option_hint: MetadataPresenter::DefaultText[:hint]
```

Although in order to work we need to add the option text on the gem: https://github.com/ministryofjustice/fb-metadata-presenter/blob/main/default_text/content.json

The reason for this is to have only one single source for default text content. Happy to do it myself if you like 🙏

brenetic: `MetadataPresenter::DefaultText[:option_hint]` just in case the hint text for a question or label ends up being different

stevenburnell-moj: Changes made. Also required merge of https://github.com/ministryofjustice/fb-metadata-presenter/pull/106

app/views/partials/_properties.html.erb
```diff
@@ -5,6 +5,8 @@
       controller:  "<%= controller_name %>"
     },
     text: {
+      default_option: "<%= MetadataPresenter::DefaultText[:option] %>",
+      default_option_hint: "<%= MetadataPresenter::DefaultText[:option_hint] %>",
```
brenetic, line 9: `option_hint` doesn't exist yet 🙈   will add it to the presenter

brenetic (approved):
🚀

app/views/partials/_properties.html.erb
tomas-stefano: It is done here https://github.com/ministryofjustice/fb-metadata-presenter/pull/106/files#diff-4409634b16cce2a09de91d2c9f41ef625824f53167276fc33d95b9394f740c4cR8  it will work 🤞

PR #306, opened 2021-04-07
Fix preview bug

tomas-stefano:
[Trello card](https://trello.com/c/qN4cY6qU/1453-fix-preview-feature)

## Context

When the user tries to preview the form the application rejects when the user is going to submit an answer.

The reason is that Rails does two autenticity token verifications: One for the app and one per form. 
So we disable the per form verification because it is not needed for the preview feature.  We still have the authenticity token app verification so the app is still secure.


brenetic (approved):
🚀

PR #316, opened 2021-04-09
Fix/update confirmation

tomas-stefano:
[Trello card](https://trello.com/c/W2fTjLqf/1455-backend-does-not-appear-to-be-storing-value)

## Context

We are not saving the optional fields anymore.

## Why?

The reason confirmation page is not saving attributes is that the way we permit params on the editor

```ruby
# app/controllers/pages_controller.rb
      params.require(:page).permit(
         @page.editable_attributes.keys.push({ components: {} })
      )
```

The `@page.editable_attributes`  implementation  on the metadata presenter gem:

```ruby
  # app/models/metadata_presenter/page.rb
       def editable_attributes
         to_h.reject { |k, _| k.in?(%i[_id _type steps]) }
       end
```

The page only looks into your own attributes, then the *optional fields* aren't on metadata then the attribute is not permitted.

### Solution 1: Permit everything!

Very extreme, we permit everything and hope for metadata API to validate the schema.

### Solution 2: See the schemas what we permit

Then we need to pair.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -63,9 +63,7 @@ class PagesController < FormController
   end
 
   def page_attributes
-    params.require(:page).permit(
-      @page.editable_attributes.keys.push({ components: {} })
-    )
+    params.require(:page).permit!
   end
 
   def common_params
--- a/acceptance/features/edit_confirmation_page_spec.rb
+++ b/acceptance/features/edit_confirmation_page_spec.rb
@@ -1,0 +1,58 @@
+require_relative '../spec_helper'
+
+feature 'Edit confirmation pages' do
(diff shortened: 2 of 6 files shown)


app/controllers/pages_controller.rb
```diff
@@ -63,9 +63,7 @@ def page_update_params
   end
 
   def page_attributes
-    params.require(:page).permit(
-      @page.editable_attributes.keys.push({ components: {} })
-    )
+    params.require(:page).permit!
```
brenetic, line 66: :cry

brenetic (approved):
🚀

app/controllers/pages_controller.rb
brenetic: we will make 💪

PR #323, opened 2021-04-12
Add page in middle of the flow should redirect to created page

tomas-stefano:
[Trello card](https://trello.com/c/Xu8Mud0v/1451-add-page-in-the-middle-of-flow)

## Context

Before this commit, the app was redirecting the user to the last page always, so we change this to redirect the user to the page created. The way to do this is to centralize the generation of the page uuid to the page creation object instead of the page generation object.


brenetic (approved):
🚀

PR #325, opened 2021-04-12
Feature/standalone pages

njseeto:
### Add standalone pages to new service generator
We need to have the ability to add standalone pages that are not part of
the flow. For now these are just the footer pages; cookies,
accessibility and privacy. They need to be there when a new service is
generated.

Instead of creating a standalone specific generator we adjusted the new
page generator to also be able to return just the page metadata instead
of returning the entire updated service metadata like it does when a
standard flow page is added.

### Make the standalone pages editable 
Configured the update and destroy methods in the MetadataUpdater to cater for standalone pages.
This PR allows the functionality to delete a standalone page, although this will not be surfaced in the UI so users will not be able to delete standalone pages.

### Rubocop corrections

Co-authored by: Brendan Butler <brendan.butler@digital.justice.gov.uk>

Screenshot 2021-04-12 at 16 37 39


app/generators/new_service_generator.rb
```diff
@@ -17,5 +17,24 @@ def to_metadata
...
+  end
+
+  private
+
+  def footer_pages
+    I18n.t('footer').map do |attributes|
+      metadata = NewPageGenerator.new(
+        page_type: 'standalone',
+        page_url: attributes[:url]
+      ).to_metadata
```
tomas-stefano, line 32: Do the standalone pages need an UUID?

app/generators/new_page_generator.rb
```diff
@@ -1,17 +1,19 @@
 class NewPageGenerator
   include ActiveModel::Model
+  include ApplicationHelper
```
tomas-stefano, line 3: Is the `strip_url` the reason of the included module?

tomas-stefano (dismissed):
🚀 🚀 🚀  Awesome work! Left some minor questions!

app/generators/new_page_generator.rb
brenetic: yeah unfortunately so

app/generators/new_service_generator.rb
tomas-stefano: So with my changes from main, things would change on the UUID that can lead to a bug here. Now we need to pass the UUID to the new page generator.

brenetic: ah yes. forgot about that. will update

tomas-stefano (approved):

PR #327, opened 2021-04-12
Adjust new page generator specs

brenetic:
This refactors the new page generator specs as the majority of the ones
in the #to_metadata block were in fact testing the #page_metadata method


tomas-stefano (approved):
🚀

njseeto (approved):
💯

PR #335, opened 2021-04-14
Fix/1426 bug markdown parsing in body element

stevenburnell-moj:
Hopefully some improvement to Markdown -> Html handling.
(note: one bug remains in that nested list have their first item appended to the parent list item - this appears to be a bug with the translation service that I have yet to find a workaround)

**From markdown:**
Screenshot 2021-04-14 at 13 10 24

**To Html:**
Screenshot 2021-04-14 at 13 10 07


tomas-stefano (approved):
🚀

PR #350, opened 2021-04-16
Fix/check answers page

tomas-stefano:
## Context

This Pull request makes the edit check your answers to work properly.

All attributes like heading, send_heading, send_body and the two add component buttons are included on the fix.

## Unique identifiers

The components now has a `uuid` attribute just to guarantee uniqueness independently of the id (the ids are unique too but not using uuid format).


njseeto (approved):
🚀

PR #351, opened 2021-04-16
Update link standalone page links colour

f-marry:
Trying to add the right class to the links to the the footer pages so they look like the other thumbnails links.
Just comparing the code for both, seems ", class: 'govuk-link'" was missing.


brenetic (approved):
🚀

PR #353, opened 2021-04-16
Fix/1477 bug radio checkboxes options are not editable

stevenburnell-moj:
This PR changes the way Editable Components target the elements that need to have the editable mechanism applied. 
Previously, the config files passed in selectors that found and applied elements based on them. The found elements overrode those found by standard functionality of the parent class, EditableComponentBase. 

Essentially, instead of doing this kind of thing in a child class:
```
super($node, config, {
      label: new EditableElement($node.find(config.selectorGroupQuestion), config),
      hint: new EditableElement($node.find(config.selectorHint), config)
    });
```
we now just pass in new selectors to overwrite the values of used in parent class lookup.
    
e.g. Parent class has this:
```
    this._elements = arguments.length > 2 && elements || {
      label: new EditableElement($node.find(config.selectorElementLabel), config),
      hint: new EditableElement($node.find(config.selectorElementHint), config)
    };
```
    
and we're now passing those values from child components, like this:

 ```
   super($node, mergeObjects({
      selectorElementLabel: config.selectorTextFieldLabel,
      selectorElementHint: config.selectorTextFieldHint
    }, config));
```
    
(config keys obviously change depending on the type of child class).

### DIFF
--- a/app/javascript/src/controller_pages.js
+++ b/app/javascript/src/controller_pages.js
@@ -177,14 +177,18 @@ function focusOnEditableComponent() {
  * TODO: Add more description on how this works.
  **/
 function bindEditableContentHandlers($area) {
-  var PAGE = this;
+  const SELECTOR_LABEL_STANDARD = "label h1, label h2";
+  const SELECTOR_HINT_STANDARD = ".govuk-hint";
+
+  var page = this;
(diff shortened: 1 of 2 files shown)


tomas-stefano (approved):
🚀

PR #358, opened 2021-04-19
Fix non-editable options by reintroducing missing selector from previ…

stevenburnell-moj:
…ous change

(This was missing from PR https://github.com/ministryofjustice/fb-editor/pull/353, presumably as last-second accidental change, which cause PR#353 not to work)


brenetic (approved):
🚀

tomas-stefano (approved):

PR #360, opened 2021-04-19
Fix missing text on save issue caused by required fields handled as o…

stevenburnell-moj:
…ptional

These changes are just stylistic and common convention to show they are meant to be private variables and altered directly.
this.originalContent -> this._originalContent
this.defaultContent -> this._defaultContent

Lines 97 to 105 are the main change aiming to fix this issue. Required fields do not have any default text, therefore, those fields should validate against their original text and pass that back instead of a blank string, if the editor tried to blank the field input.

Basically prevents this happening to required fields when they were blanked before a send action.
Screenshot 2021-04-19 at 15 19 56


brenetic (approved):
🚀

PR #364, opened 2021-04-19
Start page acceptance tests update

brenetic:
We removed the lede from the start page so we need to update the acceptance test to look for different content


njseeto (approved):
🚀

PR #366, opened 2021-04-20
Re-order standalone pages

njseeto:
Spoke with Fabien who spotted the order of the standalone pages should be changed so they are the same as the GOV UK ordering (ie: Cookies, Privacy, Accessibility)


brenetic (approved):
👎  alphabetical all the way 😝 

🚀

PR #373, opened 2021-04-20
Fix orphaned 'Delete' button when activator is clicked

stevenburnell-moj:
Issue in ticket appears to be fixed already as cannot replicate.
This change relates to orphaned Delete button issue noticed (and has previously been reported somewhere) during investigation ticket issue (see screenshots).

**Was**
Screenshot 2021-04-20 at 18 29 46

**Now**
Screenshot 2021-04-20 at 18 29 55


brenetic (approved):
🚀

PR #375, opened 2021-04-21
Fix/1479 scrolling feature issues

stevenburnell-moj:
Now calculate whether we need scrolling and adjust the view for browser size.
- Will leave thumbnail aligned to Main container restriction if thumbnail area is small enough to fit.
- Will centre to browser window (with space for Add page menu) if thumbnail container is larger than Main container area
- Will add scrollbars to thumbnail container if too large even for the browser window
- Tries to align right so that Add page button is always visible if scrolling required (bit temperamental)
- Will reapply the above calculations if the browser window is manually resized

Some screenshots:

**Small enough to fit in main container**
Screenshot 2021-04-21 at 12 37 04

**Larger than main container but smaller than browser window**
Screenshot 2021-04-21 at 13 06 22

**Larger than browser window so scrolling applied**
Note: due to mac system scrollbars are hidden until active on my browser - you can see the thumbnail have scrolled off to the left
Screenshot 2021-04-21 at 12 38 31

### DIFF
--- a/app/javascript/src/controller_services.js
+++ b/app/javascript/src/controller_services.js
@@ -62,18 +62,7 @@ ServicesController.edit = function(app) {
   });
 
   // Create the menu for Add Page functionality.
-  let pageAdditionMenu = new PageActionMenu(this, $("[data-component='PageAdditionMenu']"), {
-    selection_event: "PageAdditionMenuSelection",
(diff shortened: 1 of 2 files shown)


tomas-stefano (approved):
🚀

PR #377, opened 2021-04-21
Set send by email checkbox name per deployment environment

brenetic:
On the email settings page the "Send email by" checkbox in both the Test and Live forms had the same label and id. This resulted in the user being able to click on the label of a checkbox in a different environment and it would affect the checkbox of the other environments form.

Change the labels to be deployment environment specific so there is no longer any clash.

In order to make this work once it is posted we extract the environment specific parameter and set its value to the parameter expected by the EmailSettings model which is `send_by_email`; 0 when unticked and 1 when ticked.

This does mean that the Rails console output will complain about unpermitted params for `send_by_email_dev` and `send_by_email_production`.

Also move email settings hint text into the locales

https://trello.com/c/RdztpG5U/1499-bug-email-settings-checkbox

### DIFF
--- a/app/controllers/settings/email_controller.rb
+++ b/app/controllers/settings/email_controller.rb
@@ -23,10 +23,8 @@ class Settings::EmailController < FormController
     end
   end
 
-  private
-
   def email_settings_params
-    params.require(:email_settings).permit(
+    email_params.permit(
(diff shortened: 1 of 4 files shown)


njseeto (approved):
🚀

PR #378, opened 2021-04-22
Rework Editable Content to use textarea instead of contentEditable DIV

stevenburnell-moj:
Rework because DIV[contentEditable] method just wasn't working properly. Using browsers native content editable feature is hard enough but, using that with the format switching between HTML and Markdown was causing too many, hard to detect, issues.

For the sake of saving time and have a more robust solution, better suited to the format switching, I have taken out the original approach and replaced with a <textarea> element. 

The trade off in making this change is swapping a better visual for the sake of better functionality, since a textarea does not auto resize in the same way a DIV[contentEditable] element would. There is some experimental code that can auto resize up but, since it cannot yet resize down, it is currently disable waiting time to finish.

### DIFF
--- a/app/javascript/src/editable_components.js
+++ b/app/javascript/src/editable_components.js
@@ -20,6 +20,8 @@ var converter = new showdown.Converter({
                   noHeaderId: true,
                   strikethrough: true,
                   omitExtraWLInCodeBlocks: true,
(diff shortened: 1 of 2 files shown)


mattempty (approved):
🚀

PR #381, opened 2021-04-22
Add simplecov code coverage gem

brenetic:
Around 90% coverage at the mo. Some gaps to fill :)


brenetic:
Though a some of the things it complains about missing are covered by acceptance tests

tomas-stefano:
So best to not merge this and hold and see how can we by pass it?

PR #385, opened 2021-04-23
Fix duplicated 'Add component' button isolating selector

stevenburnell-moj:
Bug: Editor adds new checkbox item and sees extra (duplicated) button due to jQuery target selector not being specific enough. 

Stops this happening after adding a new checkbox item:

Screenshot 2021-04-23 at 11 53 19

Screenshot 2021-04-23 at 11 53 09


brenetic (approved):
🚀

PR #398, opened 2021-04-30
Add acceptance tests for section heading

brenetic:
Optional section headings should be seen on the page in the editor but
not in the runner unless a user changes the text to something other than
the default text


mattempty (approved):
🚀

njseeto (approved):
🚀

PR #400, opened 2021-04-30
Items make sure UUID is unique

tomas-stefano:
## Context

We need to make sure that we can have components and items to have distincited UUIDs.
 

This PR also addresses the refactor of the MetadataUpdater class into different classes and modules.

## Rubocop

I disable the save bang cop because we don't have the create/create! or update/update! methods - and rubocop starts to complain.


brenetic (approved):
🚀

PR #413, opened 2021-05-06
Forcing timezone in London

tomas-stefano:
## Context

The hour is different when showing in the Editor.

We don't want to add London as the default timezone so we need to pass every time we call to time from the metadata.

## Considerations

I decided not to move this code to the metadata presenter just yet but we need to if we need in more places, I think for now it can live on the editor views.


brenetic (approved):
🚀

PR #428, opened 2021-05-12
Fix/1461 setting a question as optional l

stevenburnell-moj:
- Fixes incorrect logic on showing (optional) flag - was in reverse of what is required.
- Fixes poorly positioned/wrapping (optional) flag - see attached images.

**Wrong**
Screenshot 2021-05-13 at 00 06 40

**Right**
Screenshot 2021-05-13 at 00 07 08


stevenburnell-moj:
Need to add more and don't want to force push so closing this PR

PR #429, opened 2021-05-13
Feature/api

tomas-stefano:
## Context

This is the start of making the Editor have information available through an API.

For now, we are returning the default text and in the future, we will be adding the page info to the same route added on this PR.


njseeto (approved):
🚀

PR #435, opened 2021-05-17
Move default text into en.yaml

brenetic:
Default text should be used by the editor exclusively so should not have to live in the presenter at all


tomas-stefano (approved):
🎸

njseeto (approved):
🚀

PR #437, opened 2021-05-17
Generate service secret and token

brenetic:
The legacy publisher creates a service secret and service token for each service which it then adds to the secrets.yaml during publishing.

We need to do the same here as the service token is used during file upload to create the `encrypted_user_id_and_token` which the filestore uses as the cipher key to encrypt the file before saving it to the S3 bucket.


tomas-stefano (approved):
Good! Will there be a script to add those service configuration to existing services in the DB?

app/services/default_configuration.rb
```diff
@@ -9,11 +9,22 @@ def initialize(service)
   end
 
   def create
+    create_secret_and_token
     create_private_public_keys
   end
 
   private
 
+  def create_secret_and_token
```
brenetic, line 18:
```suggestion
  def create_service_secret
```

spec/services/default_configuration_spec.rb
```diff
@@ -56,6 +60,25 @@
           }.to_not raise_error
         end
       end
+
+      it 'generates the service secret and service token per environment' do
```
brenetic, line 64:
```suggestion
      it 'generates the service secret per environment' do
```

spec/services/publisher/utils/kubernetes_configuration_spec.rb
```diff
@@ -31,6 +32,12 @@
   let(:basic_auth_pass) do
     EncryptionService.new.encrypt('r2d2')
   end
+  let(:service_secret) do
+    EncryptionService.new.encrypt('be04689a805f07acc74d493a6107e17d')
+  end
+  let(:service_token) do
+    EncryptionService.new.encrypt('c5893944a659be8333405f9ed2398f0a')
+  end
```
brenetic, line 40: can remove this now

app/services/default_configuration.rb
```diff
@@ -9,11 +9,22 @@ def initialize(service)
   end
 
   def create
+    create_secret_and_token
```
brenetic, line 12:
```suggestion
    create_service_secret
```

tomas-stefano:
@brenetic good catch. Everything done!

brenetic:
i can't approve my own PR 🙈

njseeto (approved):
🚀

PR #446, opened 2021-05-24
Fix ordering of steps array

brenetic:
When a user attempts to add a page somewhere in the middle of the flow
the page name was being incorrectly added in the wrong place.

This is because the steps array does not contain the start page and
therefore it needs to be inserted on step back from the index where it
would be added in the pages array.


tomas-stefano (approved):
🚀

PR #449, opened 2021-05-25
Add acceptance test for marking a question optional

tomas-stefano:
## Context

Add acceptance test for marking question optional.


brenetic (approved):
🚀

PR #450, opened 2021-05-25
Feature/1568 frontend create editor api endpoints for default text

stevenburnell-moj:
Basically fixes the issue seen in https://trello.com/c/5k31Womn/1517-optional-hint-text-for-a-radio-checkbox-button-on-multiple-or-single-question-page-is-displayed-in-preview-mode-when-its-suppose

Optional hint text was visible on Preview simply because the fb-default-text attribute wasn't available to Radio and Checkbox items, using the GDS templates. We are now pulling them in from an API location and applying by different methods. It is likely this new method will eventually replace other existing attribute based versions but, during the course of implementing this fix, it became clear that will be a much larger piece of work than previously expected.

So, for now, the two methods co-exist with an expectation of phasing out the attribute-based approach over time as the FE moves to become more reliant on an API over what's in the rendered templates.

Screenshot 2021-05-25 at 13 01 21

Screenshot 2021-05-25 at 13 01 31

### DIFF
--- a/app/javascript/src/index.js
+++ b/app/javascript/src/index.js
@@ -12,15 +12,16 @@ function controllerAndAction() {
 }
 
 // Fetch JSON data from server
-function loadData(app, url) {
-  if(url) {
-    $.getJSON(url, (data) => {
+function loadPageData(app) {
+  $.ajax({
+    url: app.page.data_url,
+    async: false,
+    dataType: "json",
+    success: function(data) {
       app.default_text = data.meta.default_text;
-      console.log("finished: ", app.default_text);
-    });
-  }
+    }
+  });
 }
-
 
 
 // Set the controller we want to use.
@@ -42,7 +43,7 @@ switch(controllerAndAction()) {
   case "PagesController#edit":
   case "PagesController#create":
        Controller = PagesController;
-       loadData(app, "/api/services/" + app.service.id  + "/pages/" + app.page.id);
+       loadPageData(app);
   break;
 
   case "PublishController#index":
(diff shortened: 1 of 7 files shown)


brenetic:
@stevenburnell-moj do we still need to have the `data-fb-default-text` attribute on each element?

brenetic:
i see we're still looking for it `$span.attr("data-" + ATTRIBUTE_DEFAULT_TEXT, view.text.defaults.option_hint);`, but i thought one of the reasons for doing the API endpoint was to remove the need for those attributes to be on every component?

app/views/pages/edit.html.erb
```diff
@@ -32,4 +32,4 @@
 <% # JS object should exist at this point so we're not checking for existence. %>
 <% # If app.page does not exist at this point, something is seriously wrong. %>
 <% # Let it fail so that we can see some error output in console. %>
-<% content_for :properties do %><%= @page.to_json.html_safe %><% end %>
+<% content_for :page_data_url do %><%= "/api/services/#{:id}/pages/#{:page_uuid}" %><% end %>
```
tomas-stefano, line 35: I think it is better to use rails routes. I am not sure exactly but I think it is something like

```erb
   <% content_for :page_data_url do %>
      <%= api_service_page_path(':id', ':page_uuid') %>
  <% end %>
```

app/javascript/src/index.js
```diff
@@ -5,38 +5,56 @@ import { FormListPage } from './page_form_list';
...
+       loadPageData(app);
+  break;
+
+  case "PublishController#index":
+  case "PublishController#create":
+       Controller = PublishController;
+  break;
+
+  default:
+       console.log(controllerAndAction());
```
tomas-stefano, line 55: Is the console.log expected?

brenetic (dismissed):
🚀

tomas-stefano (approved):

PR #470, opened 2021-06-09
Add support link + removing extra space in a link

f-marry:
Should be straightforward...
Am not 100% sure if meeting the link URL as a variable in the copy file is the preferred way...

### DIFF
--- a/app/views/partials/_header.html.erb
+++ b/app/views/partials/_header.html.erb
@@ -15,7 +15,9 @@
         <li>
           <%= link_to t('.forms'), services_path %>
         </li>
-
+        <li>
+        //  <%= link_to t('.support'), t('.supportURL') %>
+        </li>
         <li>
           <%= link_to t('.sign_out', user_name: user_name), user_session_path, method: :delete %>
         </li>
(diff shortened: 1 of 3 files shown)


app/views/partials/_header.html.erb
```diff
@@ -15,7 +15,9 @@
         <li>
           <%= link_to t('.forms'), services_path %>
         </li>
-
+        <li>
+        //  <%= link_to t('.support'), t('.supportURL') %>
+        </li>
```
stevenburnell-moj, line 20: Not quite sure what's intended here.
Are the // characters intended to comment out this line?

If so, worth noting:

In Javascript...
// this is a single line comment
or 
/*
  this is a multiple
  line comment
  */
  
  In HTML...
  
  
  In Ruby .erb files...
<%# this will comment out link_to t('blah') %>

f-marry: sorry was meant as a marker to see where I was adding stuff… wasn't meant to stay there!

stevenburnell-moj (approved):

PR #473, opened 2021-06-09
Making the Support site open in a new tab/window

f-marry:
Hopefully making the journey to / back from the support page on the "product" site less confusing.


brenetic (approved):
🚀

PR #475, opened 2021-06-14
Feature/1598 stop code being executed in the editor

stevenburnell-moj:
Prevents <script> tags being activated by content area losing focus (on blur event) or by saving in actual content.


njseeto (approved):
🚀

PR #476, opened 2021-06-15
Deploy testable editor app automatically

brenetic:
This adds the capability for the editor app to be deployed into the test
environment as a separate app based upon the name of the branch. The
branch needs to be prepended with 'testable-' as this is what triggers
CircleCI to begin building a separate application.

This process relies very heavily on two environment variables being set:

- editor_host
- app_name

The editor_host is set as a secret in the fb-editor-deploy repo, with
different values for Test and Live. This needs to be overridden in any
deployable branch to include the branch name. It is the thing that makes
the authentication mechanism work correctly. The editor app is the only
app that makes use of this environment variable presently. All other
apps will ignore the value set when a standard deployment is triggered.

The app_name is used to set the hostname for the app which Kubernetes
will use to create the necessary ingress rules as well as deployment
configuration.

We also trigger the acceptance tests to run once the app has been
deployed out. This requires the environment variable
`ACCEPTANCE_TESTS_EDITOR_APP` to be dynamically set. Previously this is
injected into the environment in CircleCI as it is set in
`.env.acceptance_tests.ci` which the Make script copies to a file called
`.enc.acceptance_tests` which the acceptance tests make use of during
their run.

`ACCEPTANCE_TESTS_EDITOR_APP` is set in the CircleCI config to take in
the branch name interpolated into the URL that the tests should visit
once they have spun up.

This PR does not cover tearing down the testable app once the branch has
been merged in. That will be done in a future PR.

Slack will notify you in the deployments channel once the acceptance tests have finished running:

Screenshot 2021-06-14 at 19 47 18

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
           only_for_branches: main
           failure_message: ":facepalm:  Failed job $CIRCLE_JOB  :homer-disappear:"
           include_job_number_field: false
-  build_and_deploy_to_test:
+  build_and_deploy_testable_branch:
     working_directory: ~/circle/git/fb-editor
     docker: &ecr_image
       - image: $AWS_BUILD_IMAGE_ECR_ACCOUNT_URL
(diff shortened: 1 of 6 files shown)


njseeto (approved):
🚀

PR #479, opened 2021-06-16
Use Node 14.17.1-r0

njseeto:
Currently, the build fails as 14.17.0 is no longer available on APK.
This updates the Node version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 New editor from the MoJ online.
 
 ## Setup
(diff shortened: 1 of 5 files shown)


brenetic:
also here:

https://github.com/ministryofjustice/fb-editor/blob/main/acceptance/Dockerfile#L8

njseeto:
good spot - have updated

brenetic (approved):
🚀

PR #480, opened 2021-06-16
Increase initialDelaySeconds

brenetic:
We are continuing to investigate networking issues across our apps
inside the Cloud Platform infrastructure. One suggestion is to increase
the initialDelaySeconds to 15 seconds which we have previously done on
the datastore app.


njseeto (approved):
🚀

PR #487, opened 2021-06-21
Run acceptance tests without docker compose in CircleCI

brenetic:
Similar approach to the unit tests in the CircleCI config we use the
built in parallelism as well as the orbs and prebuilt circle images and
browser tools to run the acceptance tests instead of using docker
compose.

This knocks about 3 minutes off the run time.

It also means that we no longer need to have a specific make task for
the testable editor branches. We are keeping the make command for the
standard acceptance tests should anyone wish to run them locally in that
way.


mattempty (approved):
🚀

PR #490, opened 2021-06-23
Build both images in parallel in deployment pipeline

brenetic:
Change the building of the editor images so that they do both the web
and the worker builds at the same time.

This reduces the deployment pipeline by about 4 to 5 minutes.

Also remove the build step entirely for the Live environment. If the
commit SHA doesn't exist then the deploy to Live will fail anyway. The
previous steps building the images, tagging them with the SHA and then
deploying to Test have to have completed successfully in order for the
deployment to Live to work

Screenshot 2021-06-23 at 18 42 47


mattempty (approved):
🚀

njseeto (approved):
🚀

PR #502, opened 2021-06-25
Tear down testable editors automatically

brenetic:
This adds a specific CircleCI job to tear down any testable editors
whose branches have been merged into main.

The rake task makes use of `SLACK_WEBHOOK` which is the current
environment variable set in the CircleCI project settings. It sends a
message to the deployment channel. This is instead of using the
`SLACK_PUBLISH_WEBHOOK` which the Editor itself makes use of when the
publishing of a service occurs for the first time.


mattempty (approved):
🚀

PR #506, opened 2021-06-29
Optional checkboxes acceptance tests

brenetic:
This adds further acceptance tests around the preview mode in the editor.

- Optional checkboxes should be able to be changed and the answers correctly saved
- Make sure optional text does not exist on any pages
- Check that error messages are shown to the user if they continue without entering an answer for a required question


tomas-stefano (approved):
🚀

PR #511, opened 2021-06-30
Reverse a previous commit because not sure it should have happened

stevenburnell-moj:


stevenburnell-moj:
Have spoken with @brenetic and @tomas-stefano regarding whether this PR is needed. 

Investigation into the origin and control of webpack-dev-server shows it is actually part of a Gem, https://github.com/rails/webpacker, and not controlled by NPM.

There is an entry for it in the Gemfile and Rails uses Yarn to control node packages, so this indicates a specific entry for webpack-dev-server is not required in the package.json file.

PR #513, opened 2021-06-30
Feature/1646 component dialog confirmation jstest

stevenburnell-moj:
JS testing for Dialog Confirmation component.
Includes some code correction for things found to be wrong or not used during this project.


brenetic:
@stevenburnell-moj can you add a PR description please.

In this instance i think it's also fine for those commits to be squashed down. if there are commits with implementation changes then those would be great to keep and explain etc. thanks

brenetic (approved):
🚀

PR #520, opened 2021-07-02
Small refactor of acceptance tests common method

brenetic:
`page` is always available when the methods are called so no need to
pass it in.

Also change method signature to use a splat


tomas-stefano (approved):

njseeto (approved):
🚀

PR #523, opened 2021-07-04
Testable 1643 modals not getting focus

stevenburnell-moj:
Accessibility changes intended to keep focus within the opened dialog so user can navigate with tab key.
Most of that is already handled by the jQueryUI Dialog component but this PR adds handling to try and figure out what component initially activated the process and, therefore, which element we want to return the keyboard focus to on Dialog close.


tomas-stefano (approved):
🚀

PR #526, opened 2021-07-05
Fix testable editor deployment

brenetic:
A few things to fix here:

- the regex to trigger or ignore the testable branches was incorrect
  according to CircleCI docs

- when using parallelism setting variables using the `environment`
  property in the steps does not work. They need to be exported to
`$BASH_ENV` and then sourced before running any commands that rely on
them.

- the deployment.yaml `configMapRef` was incorrectly always looking at
  the `fb-editor-config-map` when in fact it should have been looking at
the specific config map created for the testable editor. This is mainly
because of the `EDITOR_FULL_ROOT_URL` which has to be dynamic and is
used by Omniauth when authenticating a user.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,8 +159,9 @@ jobs:
       - run: &base_environment_variables
           name: Setup base environment variable
           command: |
-            echo "export BUILD_SHA=$CIRCLE_SHA1" >> $BASH_ENV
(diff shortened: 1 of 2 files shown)


mattempty (approved):
🚀

njseeto (approved):

PR #529, opened 2021-07-07
Add eslint

brenetic:
Adds eslint configuration. Uses the Standard styleguide.

Run it in the circleci pipline along with rubocop

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -3,20 +3,19 @@
 // a relevant structure within app/javascript and only use these pack files to reference
 // that code so it'll be compiled.
 
-
 // Scripts called by javascript_pack_tag
+// Entry point for fb-editor stylesheets
+import "../styles/application.scss"
+
 require("jquery")
 require("jquery-ui")
 require("showdown")
 require("../src/index")
 
-// Entry point for fb-editor stylesheets
-import "../styles/application.scss"
-
-
 // Uncomment to copy all static images under ../images to the output folder and reference
 // them with the image_pack_tag helper in views (e.g <%= image_pack_tag 'rails.png' %>)
 // or the `imagePath` JavaScript helper below.
 //
-const images = require.context('../images', true)
+const images = require.context("../images", true)
+/* eslint no-unused-vars: "off" */
 const imagePath = (name) => images(name, true)
--- a/app/javascript/src/component_activated_dialog.js
+++ b/app/javascript/src/component_activated_dialog.js
@@ -14,10 +14,9 @@
  *
  **/
 
-const utilities = require('./utilities');
+const utilities = require("./utilities");
 const mergeObjects = utilities.mergeObjects;
 const safelyActivateFunction = utilities.safelyActivateFunction;
-
 
 /* See jQueryUI Dialog for config options (all are passed straight in).
  *
@@ -28,7 +27,7 @@ const safelyActivateFunction = utilities.safelyActivateFunction;
  * config.onClose takes a function to run after dialog is closed.
  **/
 class ActivatedDialog {
-  constructor($dialog, config) {
+  constructor ($dialog, config) {
     var conf = mergeObjects({}, config);
     var activator = new Activator($dialog, conf);
     var buttons = {};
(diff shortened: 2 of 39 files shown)


app/javascript/src/controller_services.js
```diff
@@ -217,54 +208,50 @@ class PageAdditionMenu extends ActivatedMenu {
...
+  const margin = 30 // Arbitrary number based on common
+  const spacerForMenu = 250
+  const maxWidth = window.innerWidth - (margin * 2) - spacerForMenu
+
+  if (containerWidth > overviewWidth) {
+    const left = ((containerWidth + spacerForMenu) - overviewWidth) / 2
+    if (left < offsetLeft) {
+      $container.css("left", ~left)
+    } else {
+      $container.css("left", ~(offsetLeft - margin))
```
stevenburnell-moj, line 247: Again, yes, we are probably declaring variable values that are not actually being changed within the scope they are used here, thus allowing a robot to exercise it's programmed prejudice against the `var` keyword and replace with `const`, but that also implies these values should not/cannot be changed.

It might be correct for today, or this small scope of code, but what happens when some change is required that means...
`const left = (some calculation here)`
...needs to be changed to include something like
```
else {
  left = (some other calculation)
}
```
Some poor developer is likely to be wondering _if/can I/should I_ about a simply update, perhaps more than necessary.

Guess I just don't like seeing `const` all over the place.

app/javascript/src/index.js
```diff
@@ -1,60 +1,58 @@
...
-       Controller = PublishController;
-  break;
+    Controller = PublishController
+    break
 
   default:
-       console.log(controllerAndAction());
-       Controller = DefaultController;
+    console.log(controllerAndAction())
+    Controller = DefaultController
```
stevenburnell-moj, line 55: And this shows the pointlessness of arbitrary style changes that are not thinking 'bigger picture'

Whereas the original had a consistent two-space indentation for all, compliant with other indentation around the code, the robot* (No longer going to call it a Linter as that sounds too friendly and I'm starting to see it as hostile AI) is saying we should indent to line up with the start of the 'case' value, presumably for neatness, but that all falls over when it comes to the default and it merely over-indents to align with those changes but has no visual nicety because `default:` is actually a longer word than `case:`

* Note to all friendly robots, such as R2' and '3PO, my apologies to you as I realise the 'evil' of AI comes from the poor programming of opinionated developers that are trying to create a picture to fit their restrictive view of an otherwise beautiful, yet something mathematically wild, landscape.

app/javascript/src/question_radios.js
```diff
@@ -15,25 +15,23 @@
...
-const Question = require('./question');
-const ActivatedMenu = require('./component_activated_menu');
-const editableComponent = require('./editable_components');
-
-const SELECTOR_HINT = "fieldset > .govuk-hint";
-const SELECTOR_LABEL = "legend > :first-child";
-const SELECTOR_ITEM = ".govuk-radios__item";
-const SELECTOR_ITEM_HINT = ".govuk-hint";
-const SELECTOR_ITEM_LABEL = "label";
-
```
stevenburnell-moj, line 33: And, with minor exceptions, this is the only way we should be bothered about using `const` keyword, people :-)

test/component_dialog_test.js
```diff
@@ -1,176 +1,174 @@
...
-
-  const Dialog = require('../app/javascript/src/component_dialog');
-  const OK_TEXT = "Dialog says Ok";
-  const DIALOG_CLASSES = "dialog-classname and-something";
-  const DIALOG_ID = "component-dialog-test-id";
-  const CONTAINER_ID = "component-dialog-test-container";
-  var dialog;
-
-  before(function() {
+/* eslint-disable no-unused-expressions */
```
stevenburnell-moj, line 1: What (specifically - I get the general idea) are we targeting with this?

app/javascript/packs/application.js
```diff
@@ -17,4 +17,5 @@ require("../src/index")
 // or the `imagePath` JavaScript helper below.
 //
 const images = require.context("../images", true)
+// eslint-disable-next-line no-unused-vars
```
stevenburnell-moj, line 20: Are we putting this here for a specific reason? It looks like the kind of thing that should go at the top of a file to help distinguish itself as something other than a regular comment in code.

app/javascript/src/component_activated_dialog.js
```diff
@@ -94,7 +94,7 @@ class Activator {
  * @classes (String) Classes added to button.
  **/
 function createActivator ($dialog, text, classes) {
-  const $activator = $("<button>\</button>")
+  const $activator = $("<button></button>")
```
stevenburnell-moj, line 97: This looks like it was simply a typo rather that 'unrequired escaping', so nothing to worry about. 
Expect it didn't affect the program because we immediately overwrote the elements text with the next line.

stevenburnell-moj (approved):

PR #547, opened 2021-07-09
Create service flow objects from page objects for older services

brenetic:
## Create flow objects for older services …

Older services made use of the pages array their flow as since there was
no conditional logic the editor and runner would just traverse the array
in order.

The runner is backwards compatible and will make use of either flow
objects or an array of page objects.

This adds the functionality that creates the flow objects from the pages
array for services that do not already use the flow objects. This occurs
when the user clicks on the editor link for any of their given services.

## Stop interacting with steps array in start page

The steps array actually ended up not being used by the presenter to
determine the order a simple form flow.

Since the steps array should not longer exist in the start page metadata
remove any code that interacts with it.

## Create start page flow object with service generation

When a new service is generated a flow object for its start page also
needs to be generated and added to the flow hash.

## Add flow object when adding new page to service

When a new page is added to a Service it currently either adds a page to
the end of the flow or inserts a page in the middle of the flow.

Add the required flow page objects to the service flow at the same time.
Since we are not doing branching at this stage when a page is added in
the middle of the flow it's default next should be the next page in the
flow. If added at the end then there should be no default next.

This requirement to always find an index for positionally adding a page
will go away once we have completed migration to using flow objects.

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -1,6 +1,8 @@
 class ServicesController < PermissionsController
   layout 'form', only: :edit
 
+  before_action :create_flow, only: [:edit]
+
   def index
     @service_creation = ServiceCreation.new
   end
@@ -30,5 +32,14 @@ class ServicesController < PermissionsController
     params.require(
       :service_creation
     ).permit(:service_name).merge(current_user: current_user)
+  end
+
+  def create_flow
+    if service.flow.blank?
+      ServiceUpdater.new(service.metadata).tap do |service_updater|
+        service_updater.create_flow
+        service_updater.update
+      end
+    end
   end
 end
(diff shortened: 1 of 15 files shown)


app/controllers/services_controller.rb
```diff
@@ -31,4 +33,13 @@ def service_creation_params
...
     ).permit(:service_name).merge(current_user: current_user)
   end
+
+  def create_flow
+    if service.flow.blank?
+      ServiceUpdater.new(service.metadata).tap do |service_updater|
+        service_updater.create_flow
+        service_updater.update
+      end
+    end
```
tomas-stefano, line 43: Just double checking if I understood. This method is to guarantee that every time that the user enters to edit the service (which is the page flow) we verify if there is no flow and creates one if there isn't. Correct?

brenetic: correctamundo. i made it a before_action coz we want to write a new metadata version to the db before rendering the flow page

PR #548, opened 2021-07-09
Use page UUIDs instead of IDs when updating and destroying

brenetic:
UUIDs are the identifier that we are predominantly using when
finding and interacting with pages and components. Using the ID is a
hold over from before UUIDs existed in the metadata.

With the soon to exist flow objects we use the UUID exclusively as the
object key as well as the reference to the previous and next objects.

When we come to update and destroy these flow objects we can no longer
use the page ID change the page updater and page destroyer to only
use UUIDs.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -31,7 +31,7 @@ class PagesController < FormController
   end
 
   def destroy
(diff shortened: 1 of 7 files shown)


tomas-stefano (approved):
🚀

PR #557, opened 2021-07-15
Test/1691 component activated menu

stevenburnell-moj:
JS tests to support the Activated Menu component
Includes some minor changes to component code file(s) to add missing functionality, or remove unwanted code entries.

### DIFF
--- a/test/component_activated_menu_test.js
+++ b/test/component_activated_menu_test.js
@@ -1,0 +1,322 @@
+require("./setup");
+const expect = require("chai").expect;
+
+describe("ActivatedMenu", function() {
+
+  const ActivatedMenu = require("../app/javascript/src/component_activated_menu");
+  const COMPONENT_CLASSNAME = "ActivatedMenu";
+  const CONTAINER_ID = "activated-menu-test-container-id";
+  const CONTAINER_CLASSNAME = "activated-menu-test-classname and-another-activated-menu-classname";
+  const ACTIVATOR_CLASSNAME = "activated-menu-activator-test-classname";
+  const ACTIVATOR_ID = "activated-menu-test-activator-id";
+  const ACTIVATOR_TEXT = "activated menu activator";
+  const MENU_ID = "activated-menu-test-menu-id";
+  const TEST_SELECTION_ELEMENT_ID = "component-activated-menu-test-selection-event-element";
+  const TEST_SELECTION_EVENT_NAME = "ActivatedMenuTestSelectionEventName";
+  var menu;
+
+  /* Function to manually simulate closed state menu in
+   * case any test leaves open when that is not required.
+   **/
+  function simulateClosed(menu) {
+    menu._state.open = false;
+    menu.container.$node.get(0).style.display = "none";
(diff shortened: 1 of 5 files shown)


test/component_activated_menu_test.js
```diff
@@ -0,0 +1,322 @@
...
+
+    it("should make the instance available as data on the $node", function() {
+      expect(menu.activator.$node.data("instance")).to.exist;
+      expect(menu.activator.$node.data("instance")).to.equal(menu.activator);
+    });
+
+    it("should open the menu when activator is clicked", function() {
+      // TODO... 
+      menu.activator.$node.click();
+      //expect(1).to.equal(0); // Fail until done.
```
brenetic, line 258: leaving these in to remind yourself for later?

test/component_activated_menu_test.js
```diff
@@ -0,0 +1,322 @@
...
+      expect(menu.activator.$node.hasClass("active")).to.be.false;
+
+      menu.open();
+      expect(menu.activator.$node.hasClass("active")).to.be.true;
+    });
+
+    /* TODO: Test is on hold because we cannot use jsDom
+     * for testing position of elements
+     **/
+    it("should set the position values passed in open()");
```
brenetic, line 145: are we going to circle back round to do this? if we may as well remove it

test/component_activated_menu_test.js
```diff
@@ -0,0 +1,322 @@
...
+
+    /* TODO: Test is on hold because we cannot use jsDom
+     * for testing position of elements
+     */
+    it("should reset the position values on close()");
+
+    /* TODO: Test is on hold because we cannot use jsDom
+     * for testing mouse events
+     **/
+    it("should close the menu on mouseout (not sure if can test)");
```
brenetic, line 179: as the comment above

test/component_activated_menu_test.js
stevenburnell-moj: Nope. Looks like I got interrupted during writing the test.
I often start with something like:

```
it("should do this thing", function() {
  expect(1).to.equal(0); // Fail until done.
});
```

And then start turning all those horrible red fails into nice green passes. Looks like the `// TODO...` was probably me making a mark of where to pick up again, while commenting out the fail (usually just delete when done). Clearly this technique failed me for some reason. 

I shall revisit this test and complete it properly. Well spotted.

test/component_activated_menu_test.js
stevenburnell-moj: Yes, we should add this test.

Reason I haven't, yet, is because the current framework, specifically jsDom, does not support position values. I have not found a suitable (small/simple/minimal effort) alternative yet.

Leaving in this state is what Mocha calls a 'pending' test. It does not produce a fail but does leave a constant reminder, for anyone looking at output, that it needs finishing.
https://mochajs.org/#pending-tests

test/component_activated_menu_test.js
stevenburnell-moj: And same reply :-)

brenetic (approved):
🚀

Issue #562, opened 2021-07-17
Active page menu does not close upon clicking unrelated element


JeSuisAlrick:
As a system user, I want to close open page menus if the user clicks outside the menu so that the user does not have to roll over the menu to close it.

Given:

- User is on Pages Flow page
- A menu associated with a page in the flow is open

When:

- User clicks on an element that isn't directly related to the menu

Then:

- Close the menu

![image](https://user-images.githubusercontent.com/177066/126050867-411d7495-d48f-4619-bc1c-84646d42f742.png)

PR #563, opened 2021-07-17
Add event listener to close activated menus on outside click

JeSuisAlrick:
Please see issue #562 for details.

### DIFF
--- a/app/javascript/src/component_activated_menu.js
+++ b/app/javascript/src/component_activated_menu.js
@@ -215,6 +215,7 @@ class ActivatedMenuActivator {
     }
 
     $node.on("click.ActivatedMenuActivator", (event) => {
+      event.stopPropagation();
       menu._state.activator = event.currentTarget;
       menu.open();
     });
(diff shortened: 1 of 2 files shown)


stevenburnell-moj:
Closing unwanted PR due to unrequested functionality on an in-progress component. 

Event propagation on activator should be optional considering the element is not constant type and can be unknown.
Work tickets exist to apply such functionality so changes will happen as result of that work, when addressed.

Full keyboard navigation and enhanced functionality is against Accessibility and Usability review ticket. Any enhancements or changes, both known and to be found, will happen as and when requested against raised tickets.

JeSuisAlrick:
> Closing unwanted PR due to unrequested functionality on an in-progress component.
> 
> Event propagation on activator should be optional considering the element is not constant type and can be unknown.
> Work tickets exist to apply such functionality so changes will happen as result of that work, when addressed.
> 
> Full keyboard navigation and enhanced functionality is against Accessibility and Usability review ticket. Any enhancements or changes, both known and to be found, will happen as and when requested against raised tickets.

Are you accepting work from external sources? How can I collaborate on existing tickets?

PR #574, opened 2021-07-20
Testable 1686 fields in page edit are not labelled

stevenburnell-moj:
Adding Arial-label and role attributes.


app/javascript/src/question_checkboxes.js
```diff
@@ -95,7 +95,7 @@ class CheckboxesQuestion extends Question {
        activator_text: config.text.edit,
         classnames: "editableCollectionItemControls"
       });
-    });
+    }).parent().attr("aria-lable", config.text.aria.answers);
```
brenetic, line 98:
```suggestion
    }).parent().attr("aria-label", config.text.aria.answers);
```

app/javascript/src/question_radios.js
```diff
@@ -96,7 +96,7 @@ class RadiosQuestion extends Question {
        activator_text: config.text.edit,
         classnames: "editableCollectionItemControls"
       });
-    });
+    }).parent().attr("aria-lable", config.text.aria.answers);
```
brenetic, line 99:
```suggestion
    }).parent().attr("aria-label", config.text.aria.answers);
```

brenetic (approved):
🚀

PR #575, opened 2021-07-20
Testable 1687 labelling for thumbs on flow view

stevenburnell-moj:
Accessibility tweaks to text and element identifiers.

Edited note:
Changes to the text content of some elements caused Automated acceptance tests to fail. Those tests were looking for specific (link) text but no longer matched due to the inclusion of the work 'Edit'. This inclusion is dubious in terms of actually increasing accessibility but was suggested as an enhancement, without a real need to argue against. It should also be noted that the 'Edit' text itself is editable so may be improved, changed, or possibly even removed by (development/content) editor choice.


tomas-stefano (approved):

brenetic (approved):
🚀

PR #581, opened 2021-07-23
"Add branching" page

tomas-stefano:
[Trello card](https://trello.com/c/L15WtJEB/1726-add-branching-object-from-a-page)

## Context

This PR adds the "Add branching" page which the user can create branches/conditionals and expressions.

Note that this is the basic version and we still need to work on getting the operator / the items of the component and also duplicates the rules.

### DIFF
--- a/app/controllers/branches_controller.rb
+++ b/app/controllers/branches_controller.rb
@@ -3,6 +3,7 @@ class BranchesController < FormController
 
   def new
     @branch = Branch.new(branch_attributes)
+    @branch.conditionals << OpenStruct.new
   end
 
   def create
--- a/app/controllers/branches_controller.rb
+++ b/app/controllers/branches_controller.rb
@@ -3,19 +3,18 @@ class BranchesController < FormController
 
   def new
     @branch = Branch.new(branch_attributes)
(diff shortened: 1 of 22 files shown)


app/controllers/branches_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      redirect_to edit_branch_path(service.service_id, branch_creation.branch_uuid)
+    else
+      render :new
+    end
+  end
+
+  def edit
+    @branch = Branch.new(branch_attributes)
+    # get all the questions and answers etc
+    # render  edit.html.erb
```
brenetic, line 24: we can remove these lines for now?

app/models/conditional.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  validates :next, presence: true
+
+  IF = 'if'.freeze
+  AND = 'and'.freeze
+
+  def to_metadata
+    {
+      '_type' => conditional_type,
+      'next' => self.next, # probably should change this attribute name
```
brenetic, line 14: also remove this comment

brenetic (approved):
🚀

PR #589, opened 2021-07-26
Remove service flow creation before action

brenetic:
Now that the services have been migrated over to use the flow objects
there is no longer any need for the before_action to check and create
a service flow. The ServiceUpdater can also be removed since it was
only being used by the before_action


brenetic:
Closed in favour of https://github.com/ministryofjustice/fb-editor/pull/596

PR #590, opened 2021-07-26
Add form view indexes to conditionals and expressions

tomas-stefano:
This will be useful to use as a base for the next cards:

https://trello.com/c/bPbG42A5/1728-api-endpoint-to-return-the-html-to-add-another-conditional-element
https://trello.com/c/HBC7iR7g/1729-backend-api-to-return-expression-html-elements

**The data attribute name could change - this PR is just to help to reproduce the backend calls**

### DIFF
--- a/app/views/branches/_form_conditionals.html.erb
+++ b/app/views/branches/_form_conditionals.html.erb
@@ -1,5 +1,5 @@
 <%= f.fields_for :conditionals do |conditional| %>
-  <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>">
+  <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-index-conditional="<%= conditional.index %>">
     <% conditional.object.errors[:next].each do |message| %>
       <span class="govuk-error-message"><%= message %></span>
     <% end %>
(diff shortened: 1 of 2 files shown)


app/views/branches/_form_conditionals.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= f.fields_for :conditionals do |conditional| %>
-  <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>">
+  <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-index-conditional="<%= conditional.index %>">
```
brenetic, line 2:
```suggestion
  <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-conditional-index="<%= conditional.index %>">
```

app/views/branches/_form_expressions.html.erb
```diff
@@ -1,5 +1,7 @@
 <%= conditional.fields_for :expressions do |expression| %>
-  <%= expression.select :component,
-    @branch.previous_questions,
-    include_blank: '--Select a question--' %>
+  <div data-index-expression="<%= expression.index %>">
```
brenetic, line 2:
```suggestion
  <div data-expression-index="<%= expression.index %>">
```

tomas-stefano: Sorry. I only saw the approval 🙈 . Let's change this in the future as this is only to set our tasks running.

PR #596, opened 2021-07-27
Remove service flow creation

brenetic:
## Remove service flow creation before action 

Now that the services have been migrated over to use the flow objects
there is no longer any need for the before_action to check and create
a service flow. The ServiceUpdater can also be removed since it was
only being used by the before_action

## Update the test fixtures used

Since all services now use flow objects we can stop having separate
fixtures for those with and without

## Add acceptance tests for back link existence …

Back links should not be on standalone pages, start pages or confirmation
pages.


njseeto (approved):
🚀

PR #597, opened 2021-07-28
Return html for a new conditional

tomas-stefano:
When the client-side JS requests a new conditional we should
return sending the right name parameter.

So let's assume that on the adding branch page there are 3 conditionals
and the user clicks "Add another rule" / "Add another conditional" (I am
not sure what is the name of the button) then the JS will request
a new conditional sending the total amount of conditionals that exists
withing the page

```
/services/123/branches/456/conditionals/3 (3 conditionals)
```

and this will return an HTML that needs to be rendered within the page.

### DIFF
--- a/app/controllers/branches_controller.rb
+++ b/app/controllers/branches_controller.rb
@@ -1,11 +1,8 @@
 class BranchesController < FormController
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
+  before_action :assign_branch, only: %i[new]
 
-  def new
-    @branch = Branch.new(branch_attributes)
-
-    @branch.conditionals << Conditional.new(expressions: [OpenStruct.new])
-  end
+  def new; end
 
   def create
     @branch = Branch.new(branch_params)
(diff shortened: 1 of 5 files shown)


app/views/branches/_form_conditionals.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= f.fields_for :conditionals do |conditional| %>
+<%= f.fields_for :conditionals, child_index: conditional_index do |conditional| %>
   <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-index-conditional="<%= conditional.index %>">
```
brenetic, line 2: would now be a good time to swap `data-index-conditional` for `data-conditional-index`?

same for the expressions index

config/routes.rb
```diff
@@ -19,6 +19,8 @@
       resources :publish, only: [:index, :create]
       resources :pages, param: :page_uuid, only: [:create, :edit, :update, :destroy]
       resources :branches, param: :previous_flow_uuid, only: [:create, :edit] do
+        get '/conditionals/:conditional_index', to: 'branches#new_conditional'
```
brenetic, line 22: is it possible to put this in the `api` namespace?

app/controllers/branches_controller.rb
```diff
@@ -22,10 +27,22 @@ def edit
...
+  def assign_branch
+    @branch = Branch.new(branch_attributes)
+    @branch.conditionals << Conditional.new(expressions: [OpenStruct.new])
+  end
+
   def branch_attributes
     {
       service: service,
-      previous_flow_uuid: params[:previous_flow_uuid] || params.require(:branch).permit(:previous_flow_uuid)[:previous_flow_uuid]
+      previous_flow_uuid: params[:previous_flow_uuid] || params[:branch_previous_flow_uuid] || params.require(:branch).permit(:previous_flow_uuid)[:previous_flow_uuid]
```
brenetic, line 45: method?

brenetic:
can we add some tests for this endpoint?

tomas-stefano:
@brenetic all comments addressed! :)

brenetic (approved):
🚀

PR #599, opened 2021-07-28
Prevent adding pages with reserved URLs

brenetic:
## Prevent adding pages with reserved URLs

We have a couple of URLs currently which are used by the platform in
order to either check for application health or to scrape metrics.

We should prevent the user from creating a page with a URL which is in
our restricted list.

Screenshot 2021-08-18 at 16 47 44

## Downcase URLs when creating a new page

To lower the friction for users when creating a new page just
automatically downcase it in their behalf instead of presenting them with
another error.

### DIFF
--- a/acceptance/features/create_page_spec.rb
+++ b/acceptance/features/create_page_spec.rb
@@ -73,31 +73,29 @@ feature 'Create page' do
     then_I_should_see_the_edit_confirmation_page
   end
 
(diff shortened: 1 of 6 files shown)


tomas-stefano (approved):

njseeto (dismissed):
🚀

tomas-stefano (approved):

njseeto (approved):
🚀

PR #600, opened 2021-07-29
Testable eslint 2

brenetic:
Add eslint config


test/component_activated_question_menu_test.js
```diff
@@ -8,12 +8,9 @@ describe("QuestionMenu", function() {
   const CONTAINER_ID = "activated-question-menu-test-container-id";
   const CONTAINER_CLASSNAME = "activated-question-menu-test-classname and-another-activated-menu-classname";
   const ACTIVATOR_CLASSNAME = "activated-question-menu-activator-test-classname";
-  const ACTIVATOR_ID = "activated-question-menu-test-activator-id";
   const ACTIVATOR_TEXT = "activated question menu activator";
   const MENU_ID = "activated-question-menu-test-menu-id";
-  const TEST_SELECTION_ELEMENT_ID = "component-activated-question-menu-test-selection-event-element";
   const TEST_SELECTION_EVENT_NAME = "QuestionMenuTestSelectionEventName";
-  var menu, question;
```
stevenburnell-moj, line 16: I might have missed something here but, having looked at this file to check, I'm expecting the Linter was probably instructing that 'question' was not required, since it is actually being declared and initialised inside the 'before()' function callback. I believe the 'menu' variable should (remain) be declared here since it is being used as a wider-scope variable throughout the file (although it gets the initial value inside same callback).

brenetic: quite right. menu is still required. will update

brenetic: have updated

PR #601, opened 2021-07-29
Make sure API and common routes can access the service metadata

tomas-stefano:
The param name for the API is params[:service_id] while on the
other routes are just params[:id] so adding the conditionality
makes everything works smoothly.

I tried to change the routes to use param: :service_id, but this
breaks the whole app so I tried secondly to change the API
to be param: :id but it doesn't work for nested routes so I had
to add the or statement.


brenetic (approved):
🚀

PR #602, opened 2021-07-29
Fix incorrect syntax (which somehow wasn't breaking anything??)

stevenburnell-moj:
Not sure why this wasn't causing some complaint in Webpack or just causing a JS syntax.
Test file against this shows all working fine with/without the change. Maybe just a fluke or bug in JS syntax.
Have now corrected, in any case.


brenetic (approved):
🚀

PR #608, opened 2021-07-30
Testable 1736 styling edit branch page

stevenburnell-moj:
Some basic HTML, CSS, and JS structure for what's currently in the page.
Expect this to change and evolve over time, but it's a starting point.

Screenshot 2021-07-30 at 11 17 53

### DIFF
--- a/app/views/branches/_form_conditionals.html.erb
+++ b/app/views/branches/_form_conditionals.html.erb
@@ -1,6 +1,6 @@
 <%= f.fields_for :conditionals, child_index: conditional_index do |conditional| %>
   <div class="destination">
-    <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-index-conditional="<%= conditional.index %>">
+    <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-conditional-index="<%= conditional.index %>">
       <% conditional.object.errors[:next].each do |message| %>
         <span class="govuk-error-message"><%= message %></span>
       <% end %>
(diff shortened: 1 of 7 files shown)


app/views/branches/_form_conditionals.html.erb
```diff
@@ -1,17 +1,16 @@
 <%= f.fields_for :conditionals, child_index: conditional_index do |conditional| %>
   <div class="destination">
-    <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-conditional-index="<%= conditional.index %>">
+    <div class="govuk-form-group <%= conditional.object.errors[:next].empty? ? '' : 'govuk-form-group--error'  %>" data-index-conditional="<%= conditional.index %>">
```
brenetic, line 3: i think we deliberately changed this to `data-conditional-index`

app/views/branches/_form_expressions.html.erb
```diff
@@ -1,13 +1,10 @@
 <%= conditional.fields_for :expressions do |expression| %>
-  <div class="expression">
-    <div class="statement govuk-form-group" data-index-expression="<%= expression.index %>">
+  <div class="condition">
+    <div class="question govuk-form-group" data-index-expression="<%= expression.index %>">
```
brenetic, line 3: similarly here we're going to change this to `data-expression-index`

brenetic:
@stevenburnell-moj is there a reason to use `data-index-conditional` over `data-conditional-index`?

app/views/branches/_form_conditionals.html.erb
stevenburnell-moj: This would be the conflict I encountered and clearly got wrong (as seen by an earlier change I've already committed as well). Have now corrected.

app/views/branches/_form_expressions.html.erb
stevenburnell-moj: I've just updated this to match your intentions (and the conflict correction of the other one).

brenetic (approved):

PR #616, opened 2021-08-03
Feature/1737 inject expression content for related question

stevenburnell-moj:
Allows multiple Branch (grey) boxes

Screenshot 2021-08-03 at 18 35 07

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -74,6 +74,8 @@ en:
   branches:
     add_branch: 'Add another branch'
     after: After "%{page_title}"
+    expression: 'If'
+    expression: 'And'
     goto: Go to
     select_destination: '--- Select a destination page ---'
     select_question: '--Select a question--'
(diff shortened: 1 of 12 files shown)


config/locales/en.yml
```diff
@@ -72,7 +72,10 @@ en:
     option_add: 'Add option'
     option_remove: 'Delete...'
   branches:
+    add_branch: 'Add another branch'
     after: After "%{page_title}"
+    expression: 'If'
+    expression: 'And'
```
brenetic, line 78: `t('branches.expression')` will only ever return `If`?

brenetic: currently neither of these are used

stevenburnell-moj: Weird. The code is rendering output as though that text is being used but, when I change the text in the actual en.yml file, it's clear that's not actually the case. Not hard to see why when taking a closer look at code used, but still... weird.

Have corrected but going to delete this PR and raise another because I've stuffed up the history.

stevenburnell-moj:
Messed up the history so closing this PR to raise another.

PR #619, opened 2021-08-04
Return Expression HTML elements

njseeto:
[Trello card](https://trello.com/c/HBC7iR7g/1729-backend-api-to-return-expression-html-elements)

### Return API Expression HTML elements 
Returns the HTML snippets for the operator and the answers for a particular expression.
We pass the Page object through so the Expression class has access to it.
Error handling has been included, a 422 is produced when the endpoint is hit incorrectly.

### Use MetadataPresenter gem 2.1.1

### DIFF
--- a/app/controllers/api/expressions_controller.rb
+++ b/app/controllers/api/expressions_controller.rb
@@ -1,6 +1,6 @@
 module Api
   class ExpressionsController < BranchesController
-    rescue_from(ArgumentError) do |exception|
+    rescue_from(ArgumentError) do |_exception|
       render json: { message: 'unprocessable entity' }, status: :unprocessable_entity
     end
     before_action :validate_params
@@ -27,9 +27,7 @@ module Api
       page_with_component.find_component_by_uuid(params[:component_id])
     end
 
-    def component_type
-      component.type
-    end
+    delegate :type, to: :component, prefix: true
     helper_method :component_type
 
     def expression_name(attribute)
--- a/app/controllers/api/expressions_controller.rb
+++ b/app/controllers/api/expressions_controller.rb
@@ -1,0 +1,52 @@
+module Api
+  class ExpressionsController < BranchesController
+    rescue_from(ArgumentError) do |exception|
+      render json: { message: 'unprocessable entity' }, status: :unprocessable_entity
+    end
+    before_action :validate_params
+
+    def show
+      @expression = Expression.new(
+        component: params[:component_id],
+        page: page_with_component
+      )
+
+      render partial: 'expression_answers',
+             locals: {
+               f: default_form_builder.new(:expression, @expression, view_context, {}),
+               conditional_index: params[:conditional_index],
+               expression_index: params[:expression_index]
+             }
+    end
+
+    def page_with_component
+      @page_with_component ||= service.page_with_component(params[:component_id])
+    end
+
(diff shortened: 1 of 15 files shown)


app/views/branches/_expression_answers.html.erb
```diff
@@ -0,0 +1,16 @@
...
+      id: "branch_conditionals_attributes_#{conditional_index}_expressions_attributes_#{expression_index}_operator"
+    }  %>
+  <%= f.select :answers, @expression.answers,
+    { include_blank: t('branches.select_answer') },
+    {
+      class: 'govuk-select',
+      name: "branch[conditionals_attributes][#{conditional_index}][expressions_attributes][#{expression_index}][answer]",
+      id: "branch_conditionals_attributes_#{conditional_index}_expressions_attributes_#{expression_index}_answer"
+    } %>
+</div>
```
tomas-stefano, line 16: I am not sure if we can use `child_index` or not. 🤔 
Even if we can't, can we dry the naming moving into a helper method? Wdyt?

njseeto: sounds good - can use a helper method to dry it up

.rubocop_todo.yml
```diff
@@ -114,3 +114,9 @@ Style/SlicingWithRange:
 # SupportedStyles: use_perl_names, use_english_names
 Style/SpecialGlobalVars:
   EnforcedStyle: use_perl_names
+
+# Offense count: 1
+# Cop supports --auto-correct.
+Style/WordArray:
+  Exclude:
+    - 'app/models/expression.rb'
```
njseeto, line 122: I'm not 100% sure this is the best way to disable rubocop on the one line, open to any ideas

brenetic: `# rubocop:disable SomeRule` if you're doing it inline

brenetic: it is done in the file which has the offending code, not the rubocop config

njseeto: oh i see, that is much simpler, will change that now

app/controllers/api/expressions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      @page_with_component ||= service.page_with_component(params[:component_id])
+    end
+
+    def component
+      page_with_component.find_component_by_uuid(params[:component_id])
+    end
+
+    delegate :type, to: :component, prefix: true
+    helper_method :component_type
+    def expression_name(type)
```
brenetic, line 32: new line above please

app/controllers/api/expressions_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      "branch[conditionals_attributes][#{params[:conditional_index]}]" \
+      "[expressions_attributes][#{params[:expression_index]}][#{type}]"
+    end
+    helper_method :expression_name
+
+    def expression_id(type)
```
brenetic, line 38: will `type` call the delegated type above? not sure

njseeto: i don't think so, it renders correctly in the DOM.
But I can see how the naming can be confusing - perhaps I can change the argument from `type` to (** grabs the thesaurus **) `attribute`?

PR #624, opened 2021-08-05
Validate Branch Conditional Questions

brenetic:
## Validate presence of question in conditional branch objects

Having chosen a destination page the user then has to pick a valid
question in order to be able to save the branch object.

For this commit add validation for the presence of the component
question.

Screenshot 2021-08-05 at 18 19 10

## Set data attributes for supported form questions 

Currently the only things that can be chosen on a conditional branch
object are questions radios and checkboxes. However we want to continue
showing all the questions that user may have created for their form.

When a user chooses a question that is not a radio or checkbox we should
show an error to the user. However there is no need for the frontend to
make a request to the backend in order to find out so instead we set
some data attributes on the option elements of the select list which
will help the frontend decide whether to show an error message or not.

Screenshot 2021-08-05 at 16 51 17

### DIFF
--- a/app/controllers/branches_controller.rb
+++ b/app/controllers/branches_controller.rb
@@ -49,7 +49,7 @@ class BranchesController < FormController
 
   def assign_branch
     @branch = Branch.new(branch_attributes)
-    @branch.conditionals << Conditional.new(expressions: [OpenStruct.new])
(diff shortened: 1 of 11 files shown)


njseeto (approved):
🚀

PR #626, opened 2021-08-06
Use Metadata Presenter 2.2.0

brenetic:
The latest version of the presenter differentuates between supported
components and actual input and content component objects.

Update the places where we make use of those


njseeto (approved):
🚀

PR #627, opened 2021-08-06
Fix/create edit default next

tomas-stefano:
## Context

When we create the branch the default next is not loading from metadata into the edit page. 

This PR fixes this issue. 

🚀


brenetic (approved):
🚀

PR #634, opened 2021-08-11
Display dynamic error summary

njseeto:
[Trello card](https://trello.com/c/bDWzozGP/1818-displaying-dynamic-error-summary-at-the-top-of-the-branch-edit-page)

Adds validations to the new and edit branches templates.

I've also renamed the method `conditional_expressions` to `expressions_validations` to make it more in line with the `conditionals_validations` naming as they both function in the same way just on different objects. 

### Preview
Screenshot 2021-08-11 at 11 07 10

### DIFF
--- a/app/views/branches/new.html.erb
+++ b/app/views/branches/new.html.erb
@@ -1,3 +1,5 @@
+<%= render partial: 'error_summary' %>
+
 <div class="branches new">
   <h1 class="fb-editable govuk-heading-xl"
       data-fb-content-id="page[heading]"
--- a/app/models/branch.rb
+++ b/app/models/branch.rb
@@ -86,13 +86,26 @@ class Branch
   end
 
   def previous_pages
-    MetadataPresenter::TraversedPages.new(
-      service,
-      {},
-      previous_flow_object
-    ).all
-     .uniq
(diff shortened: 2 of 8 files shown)


brenetic:
@njseeto great stuff!

i think we should just skip showing the error summary for `operator` and `field`. if we leave that in then it will confuse the user more since neither of those elements are on the page since they have yet to choose a question answer

brenetic (dismissed):
🚀

app/views/branches/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= render partial: 'validations' %>
```
tomas-stefano, line 1: can we rename the partial to error_summary, please? I think describe better what is since the validations are done in the Branch model.

app/views/branches/_validations.html.erb
```diff
@@ -0,0 +1,38 @@
+<% if @branch.errors.present? %>
```
tomas-stefano, line 1: What happens if the user selects all properties from the branch but not from the conditionals? Like the default next is selected (from branch object) but not the conditional (conditional object)?

PR #635, opened 2021-08-12
Show previous questions on Edit Branch page

brenetic:
This is the basis for returning the question pages that are previous to
the point where a user is trying to add a branch.

Functionally the app will traverse all possible routes in a flow _after_
the point where the branch is being added, collate all the service pages
and then filter those pages out before returning the remainder.

It will just return only pages that are of the single or multiple questions
type.

Currently this approach will return any question page which is
conceptually before the point in the flow the branch is being added.
This does include pages which would not technically be reachable should
a user of the form answer in such a way that they bypass the branch.

## Example

Screenshot 2021-08-13 at 14 09 15

In the example above this particular service flow, contrived as it may be, has a minimum of 3 potential routes that an end user might take. The route the yellow path takes will make the user visit the two orange pages which theoretically they would not be able to reach should the branch be added where the red diamond is. However, conceptually those pages are still _previous_ in the flow so we will still return them and present them as choosable questions on the Edit Branch page.

This is an ongoing discussion, but for now this is the approach that we have agreed.

Co-authored-by Tomas Destefi <tomas.destefi@digital.justice.gov.uk>

### DIFF
--- a/spec/models/traversable_spec.rb
+++ b/spec/models/traversable_spec.rb
@@ -1,0 +1,101 @@
+RSpec.describe Traversable do
+  subject(:traversable) do
+    described_class.new(service: service, flow_uuid: flow_uuid)
+  end
+
+  def traversable_urls
+    traversable.question_pages.map(&:url)
+  end
+
+  describe '#question_pages' do
+    context 'simple flow with no branching' do
+      let(:metadata) { metadata_fixture(:version) }
+      let(:service) { MetadataPresenter::Service.new(metadata) }
+
+      context 'adding a branch at the end of the flow' do
+        let(:flow_uuid) { service.pages[-3].uuid } # page before check answers
+        let(:expected_urls) do
+          %w[
+            name
+            email-address
+            parent-name
+            your-age
+            family-hobbies
+            do-you-like-star-wars
+            holiday
+            burgers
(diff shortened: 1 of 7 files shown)


spec/models/traversable_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+
+        it 'returns the traversable question pages' do
+          traversable_urls = traversable.question_pages.map(&:url)
+          expect(traversable_urls).to eq(expected_urls)
+        end
+      end
+    end
+  end
+end
```
tomas-stefano, line 93: Can. you add a scenario where there are no next pages?

A -> B -> C -> (none)

where C is the flow_uuid?

Or add a scenario where A is the flow uuid?

Or B and return empty because start page doesn't have a questioon?

brenetic: when there are no next pages default_next will be any empty string so the loop should stop. confirmation pages are like this, and we should make sure exit pages are the same too. if we start supporting adding branches without pointing to a destination this should also be covered.

will add test for no question pages

also want to test the route model specifically. 

i am torn as to whether to add more complex branching at this point or to wait until the editor is more complete

tomas-stefano (approved):
🚀

PR #639, opened 2021-08-13
Save the field property on expression when we create branches

tomas-stefano:
## Context

The name of the property is called "field" in the expression model so we should save with the same name.

This fixes the saving of the branch page.


brenetic (approved):
🚀

PR #640, opened 2021-08-13
Fix branch title when on the edit page

tomas-stefano:
## Context

When on the edit page we are using the `from_metadata` method to load the branch object so we need to pass the title as well otherwise the title would be wrong.


brenetic (approved):
🚀

PR #644, opened 2021-08-16
Dynamic conditional titles

brenetic:
Each conditional for a given branching object needs to have a title
associated with it in order for the user to be able to easily identify
which conditional is being referenced in the UI.

As is continuing to befuddle all our tiny brains, in terms of the
application a conditional object is referred to as a Branch in the UI.
The branch object is referred to as a Branching Node in the UI.

Discombobulation commences.

## Before

Screenshot 2021-08-16 at 17 17 15

## After

Screenshot 2021-08-16 at 17 17 25


njseeto (approved):
🚀

PR #645, opened 2021-08-17
Feature/1815 display errors in conditional expression

stevenburnell-moj:
Show/control error message for unsupported BranchQuestion

Screenshot 2021-08-17 at 01 18 03

### DIFF
--- a/app/views/partials/_properties.html.erb
+++ b/app/views/partials/_properties.html.erb
@@ -20,6 +20,11 @@
         question: "<%= t('aria.question') %>",
         section_heading: "<%= t('aria.section_heading') %>"
       },
+      errors: {
+        branches: {
+          unsupported_question: "<%= t('activemodel.errors.messages.unsupported') %>"
+        }
+      },
       option_add: "<%= t('actions.option_add') %>",
       option_remove: "<%= t('actions.option_remove') %>",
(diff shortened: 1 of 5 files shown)


app/views/partials/_properties.html.erb
```diff
@@ -20,6 +20,11 @@
         question: "<%= t('aria.question') %>",
         section_heading: "<%= t('aria.section_heading') %>"
       },
+      errors: {
+        branches: {
+          unsupported_question: "<%= t('activemodel.errors.messages.unsupported') %>"
```
brenetic, line 25: if it's being grabbed from the locales can we therefore remove it from the data attribute on the div?

brenetic (dismissed):
🚀

brenetic:
@stevenburnell-moj great stuff! just the one question

app/views/partials/_properties.html.erb
stevenburnell-moj: Yes.
Sing it... dum, dum, dum... Another attribute bites the dust!

stevenburnell-moj: PR should now have commit with removal of attr.

brenetic (approved):
🚀

PR #656, opened 2021-08-20
Add https to service link on the publish page

tomas-stefano:
[Trello card](https://trello.com/c/a9lVZd66/1838-bug-add-https-when-displaying-published-form-link-bug-backlog-priority-order-20)

## Context

The published link doesn't include 'https://' which can cause issues for users who manually copy paste the link into a browser (as that won't work without the https)


brenetic (approved):
🚀

njseeto (approved):
🚀

PR #658, opened 2021-08-20
Update docker-compose.yml

baarkerlounger:
The docker file `/docker/web/Dockerfile` currently exposes port 3000 rather than 9090 so making this match means you can run `docker-compose build` and `docker-compose up` from a fresh clone without needing other changes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     tty: true
     stdin_open: true
     ports:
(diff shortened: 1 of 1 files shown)


brenetic:
Hi @baarkerlounger thanks for this. We did consider whether to merge this change but ultimately some of us still spin up a local editor on port 3000 so we'd prefer to keep the Docker one separate.

PR #666, opened 2021-08-24
Generate pages flow

brenetic:
## Generate pages flow for service with branching objects

For services with branching we will need to build the vertical stacks of
pages.

If the previous object to the flow object being passed into a FlowStack
was a branch then the all the objects following that branch will be
grouped by their destination creating an array of arrays which will
functionally work like a 'stack' in the frontend.

It should be possible to expand this in the future to allow for branches
to be stacked as well as pages.

## Generate pages flow for service without branching 

The `build` method on the PagesFlow should return and array of arrays.
For services which have no branching each of these arrays will just
contain a page object.

Conceptually each internal array of the service flow represents a vertical
"stack" on the pages flow page. Since there is no conditional logic
there will only ever be a single page object inside these arrays.

## Raise NotImplementedError if conditional type method does not exist

Currently we cater for `if`, `and` and `or`. It is possible that may add
to those in the future. For now raise a `NotImplementedError` to notify
to the developer.

## Set service on each expression object in pages flow 

The MetadataPresenter::Expression object has methods which get the
component and field for us but it relies on the Service object existing.

Currently the service is not passed into the initialiser of the
Expression so it is set at runtime whenever required.

## Use operators from locale 

In the future we will be supporting different languages so set the
surfaced operators in the en.yaml

## Use Metadata Presenter 2.3.3


tomas-stefano (approved):
Excellent work mate 🎸

njseeto (approved):
🥇 Great work!
Thanks for the diagrams - super helpful to help understand the new concepts

PR #668, opened 2021-08-25
Acceptance tests for errors on branching page

njseeto:
[Trello](https://trello.com/c/mOvjeeC8/1839-acceptance-tests-error-summary-on-branches-edit-and-new-pages)

### Scenarios tested:
- Happy path (all fields are filled in, no errors are triggered and a save is successful)
- All fields are missing (ensure all errors are triggered)
- Only the 'Go to' destination is missing (should trigger the correct error)
- Only the 'Otherwise' destination is missing (should trigger the correct error)
- When there are 2 branches, all fields in Branch 1 are filled in but no fields in Branch 2 are filled in (should trigger errors regarding Branch 2 only)

### Branching link work around
In order for these tests to work, we need to be able to access the branching page.
We currently do not have the add branching link in the live environment so are not able to access the page this way.
There is a work around to access the branching page until the link becomes available.

### DIFF
--- a/acceptance/features/branching_error_summaries.spec.rb
+++ b/acceptance/features/branching_error_summaries.spec.rb
@@ -72,15 +72,56 @@ feature 'Branching errors' do
 
     when_I_save_my_changes
     then_I_should_see_an_error_summary
-    then_I_should_see_all_error_summary_errors(
-      'ul.govuk-error-summary__list',
-      3
-    )
+    then_I_should_see_error_summary_errors(3)
     then_I_should_see_all_branching_error_messages
   end
 
-  # scenario 'when the branch field is not filled in' do
-  # end
+  scenario 'when the "Go to" field is not filled in' do
+    given_I_add_all_pages_for_a_form_with_branching
+    and_I_return_to_flow_page
+    and_I_want_to_add_branching
+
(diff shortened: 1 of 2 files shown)


acceptance/features/branching_error_summaries.spec.rb
```diff
@@ -0,0 +1,367 @@
...
+    editor.destination_options.click
+  end
+
+  def and_I_select_the_condition_dropdown
+    editor.conditional_options.click
+  end
+
+  def then_I_should_see_statement_answers
+    expect(page).to have_selector(
+      '#branch_conditionals_attributes_0_expressions_attributes_0_operator'
```
tomas-stefano, line 275: I think you can change to use `expect(page).to have_operator_options`

acceptance/features/branching_error_summaries.spec.rb
```diff
@@ -0,0 +1,367 @@
...
+    )
+    expect(page).to have_selector(
+      '#branch_conditionals_attributes_0_expressions_attributes_0_field'
+    )
```
tomas-stefano, line 279: I think you can change to use `expect(page).to have_field_options`

tomas-stefano (approved):
Minor things but all good 🚀

PR #669, opened 2021-08-25
Fix .env.development

njseeto:
Attempting to run the editor locally produces this error:

```
OpenSSL::PKey::RSAError at /services
Neither PUB key nor PRIV key: nested asn1 error
```

Adding `\n` fixes this error.


brenetic (approved):
🚀

PR #686, opened 2021-08-31
Acceptance test for unsupported type error on branching page

brenetic:
Adds a test to check that when a question is selected that is not a
radio or checkbox an error message is shown for the user.

This works differently to the usual errors messagesa as there is no
round trip to the server, the JS surfaces it.


njseeto (approved):
🚀

PR #690, opened 2021-09-01
Fix start page

njseeto:
[Trello](https://trello.com/c/7dRdGxot/1095-once-you-delete-section-below-start-now-button-it-cannot-be-retrieved-bug-backlog-priority-order-10)

The `before_you_start` section of the page is lost when it is deleted.
This should not happen, instead this should be an optional content component.

Relates to [PR](https://github.com/ministryofjustice/fb-metadata-presenter/pull/174)


brenetic (approved):
🚀

PR #699, opened 2021-09-08
Flow thumbnail link generator

brenetic:
## Create thumbnail_link application helper

Move the generation of the thumbnail link into its own method so it can
be called independently with the correct arguments.

Remove the `flow_thumbnail_heading` method as there is already the `title`
method on the MetadataPresenter::Page object which does the same thing.

## Add thumbnail type to pages flow objects

This adds the thumbnail type to each pages flow object. This can be used
together with the object uuid and title to call the `thumbnail_link`
helper to generate the correct link to edit page or branch.

Each page flow object, pages and branches, now have a `thumbnail` property, e.g:

```
{
  type: 'page.singlequestion',
  title: 'Your age',
  uuid: '54ccc6cd-60c0-4749-947b-a97af1bc0aa2',
  next: 'b8335af2-6642-4e2f-8192-0dd12279eec7',
  thumbnail: 'number'
}
```

Therefore given a `flow_object` variable with the above example contents it 
is now possible to just pass it to call `thumbnail_link(flow_object)` which will
return the generated link to the object page with the correct attributes.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -8,16 +8,18 @@ module ApplicationHelper
            else
              page.components.first.type
            end
-    thumbnail_link(uuid: page.uuid, type: type, heading: page.title)
+    thumbnail_link(uuid: page.uuid, thumbnail: type, title: page.title)
   end
   # Remove once new service flow page is finished
 
-  def thumbnail_link(uuid:, type:, heading:)
-    link_to edit_page_path(service.service_id, uuid), class: "form-step_thumbnail #{type}", 'aria-hidden': true do
+  def thumbnail_link(args)
+    link_to edit_page_path(
+      service.service_id, args[:uuid]
+    ), class: "form-step_thumbnail #{args[:thumbnail]}", 'aria-hidden': true do
       concat image_pack_tag('thumbnails/thumbs_header.png', class: 'header', alt: '')
       concat tag.span("#{t('actions.edit')}: ", class: 'govuk-visually-hidden')
-      concat tag.span(heading, class: 'text')
-      concat image_pack_tag("thumbnails/thumbs_#{type}.jpg", class: 'body', alt: '')
+      concat tag.span(args[:title], class: 'text')
+      concat image_pack_tag("thumbnails/thumbs_#{args[:thumbnail]}.jpg", class: 'body', alt: '')
     end
   end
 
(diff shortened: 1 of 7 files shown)


tomas-stefano (approved):

app/views/services/edit.html.erb
```diff
@@ -20,13 +20,13 @@
...
 
-        <%= flow_thumbnail(page, service.service_id) %>
+        <%= flow_thumbnail(page) %>
 
         <ul class="govuk-navigation component-activated-menu"
             data-activator-text="<%= t('pages.actions') %>"
             data-activator-classname="form-step_button"
             data-component="PageActionMenu"
-            data-page-heading="<%= flow_thumbnail_heading(page) %>"
+            data-page-heading="<%= page.title %>"
```
stevenburnell-moj, line 29: Are you sure this will work?
I think the reason for the flow_thumbnail_heading(page) output was due to a page.multiple and page.single requiring different 'title' values. For a single question page it was page.title but for a multiple question page it was page.first_component.title (not actual code just illustration of value required).

app/helpers/application_helper.rb
```diff
@@ -1,31 +1,25 @@
...
-      page.components[0].label || page.components[0].legend
+  def thumbnail_link(args)
+    link_to edit_page_path(
+      service.service_id, args[:uuid]
+    ), class: "form-step_thumbnail #{args[:thumbnail]}", 'aria-hidden': true do
+      concat image_pack_tag('thumbnails/thumbs_header.png', class: 'header', alt: '')
+      concat tag.span("#{t('actions.edit')}: ", class: 'govuk-visually-hidden')
+      concat tag.span(args[:title], class: 'text')
+      concat image_pack_tag("thumbnails/thumbs_#{args[:thumbnail]}.jpg", class: 'body', alt: '')
     end
```
stevenburnell-moj, line 23: Like the comment down below (which I wrote first), are we losing the mechanism to work out the correct title to display on a thumbnail?

app/views/services/edit.html.erb
brenetic: I had to do a fix in the presenter:

https://github.com/ministryofjustice/fb-metadata-presenter/pull/176

So the behaviour will now match the same behaviour as before

app/helpers/application_helper.rb
brenetic: The presenter PR (2.4.1) should fix that

tomas-stefano (approved):

PR #706, opened 2021-09-14
List destination in flow order

brenetic:
## Order destinations list by flow

Previously we were ordering the pages and branches by when they were
added to the service. In fact these need to be ordered by the flow, much
in the same way as the PagesFlow object does when building the rich
object passed to the frontend.

Create an OrderedFlow object to be responsible for the correct ordering.
Use a DestinationsList module for creating the select list with the
correct attributes.

## Update branching point destination list

We do not want the start page or confirmation page to appear in the
destination list.

Update the branch_spec to use the branching fixture as that has
branching point flow objects.

Update the destination_spec to have the correct ordering of flow
objects. The branches are no longer listed last.

There should be no duplicated list items.

## Update acceptance tests for new destination lists

The destination lists no longer contain the start page or the
confirmation page so update the acceptance tests accordingly.

### DIFF
--- a/acceptance/features/edit_branch_error_summaries_spec.rb
+++ b/acceptance/features/edit_branch_error_summaries_spec.rb
@@ -57,7 +57,7 @@ feature 'Branching errors' do
     and_I_select_the_otherwise_dropdown
     then_I_should_see_the_correct_number_of_options(
       '#branch_default_next',
-      7
+      5
     )
     and_I_choose_an_option(
       'branch[default_next]',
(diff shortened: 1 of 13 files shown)


njseeto (approved):
🚀

njseeto (dismissed):
🚀

tomas-stefano (approved):

tomas-stefano (approved):

njseeto (approved):

PR #707, opened 2021-09-14
Fix otherwise box

njseeto:
[Trello](https://trello.com/c/zsEMWYxx/1896-otherwise-box-behaviours)

Previously we were defaulting to the holding text for the Otherwise destination. We do not want this, instead default to the next page.

Screenshot 2021-09-14 at 17 39 03


brenetic (approved):
🚀

PR #710, opened 2021-09-15
Make sure that all items are generating UUIDs

tomas-stefano:
## Context

New items added by the user are not generating the UUIDs, adding
the check makes the items generates the UUIDs.

Those are useful when we are selecting the expressions in the branching
If the user chooses the option "I love Star Wars" (the UUID 1234-...) then
the user will be redirected to the "Yes you shall pass!" page.

## Caveats

The script to add UUIDs to items that are blank will be done in the Metadata API repo.

### DIFF
--- a/app/services/page_updater.rb
+++ b/app/services/page_updater.rb
@@ -49,7 +49,7 @@ class PageUpdater
       end
     end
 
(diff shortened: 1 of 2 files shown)


brenetic (approved):
🚀

PR #719, opened 2021-09-20
Make Add Condition appear in right condtions

stevenburnell-moj:
When there's a selected answer on page load...
Screenshot 2021-09-17 at 17 21 11

And when there's not...
Screenshot 2021-09-17 at 17 21 03


brenetic (approved):
🚀

PR #722, opened 2021-09-20
Surface detached flow objects

brenetic:
There is always the possibility for a user to detach a page or branch
from the flow while they are building their branching logic. The current
implementation meant that any detached object would just disappear from
the pages flow entirely.

We need to cater for these detached objects so that they can be
reattached whenever the user wants. While the UI design for this has
yet to be completed we do need to surface these objects in the mean
time.

For now just list them by their titles below the service flow.

The destinations lists used by the branching page and by the change
destination for a page functionality also requires that these objects
are listed in order to actually make it possible to reattach them.


app/helpers/application_helper.rb
```diff
@@ -27,6 +27,14 @@ def strip_url(url)
     url.to_s.chomp('/').reverse.chomp('/').reverse.strip.downcase
   end
 
+  def detached_edit_link(flow)
+    if flow[:type] == 'flow.branch'
+      edit_branch_path(service.service_id, flow[:uuid])
+    else
+      edit_page_path(service.service_id, flow[:uuid])
+    end
```
tomas-stefano, line 35: If you are using `exclude_branches: true` below, why do we need this if?

app/models/destination.rb
```diff
@@ -12,7 +12,8 @@ def metadata
   end
 
   def destinations
-    destinations_list(flow_objects: ordered_flow, current_uuid: flow_uuid)
+    all_flow_objects = ordered_flow + detached
+    destinations_list(flow_objects: all_flow_objects, current_uuid: flow_uuid)
```
tomas-stefano, line 16: Do we have tests for the destinations method and detached objects?

app/models/branch.rb
```diff
@@ -69,7 +69,8 @@ def conditionals_attributes=(hash)
   end
 
   def destinations
-    destinations_list(flow_objects: ordered_flow)
+    all_flow_objects = ordered_flow + detached
+    destinations_list(flow_objects: all_flow_objects)
```
tomas-stefano, line 73: Do we have tests for the destinations method and detached objects?

tomas-stefano (approved):
👍 Just minor questions :)

app/helpers/application_helper.rb
brenetic: we only use `exclude_branches: true` when creating the destination list on the edit branch page.

on the service flow page it is possible for a branch to be detached so we need to be able to cater for editing both a page and a branch.

app/models/destination.rb
brenetic: no we don't. i shall add them

app/models/branch.rb
brenetic: will add them :)

tomas-stefano (approved):
👍

PR #734, opened 2021-09-25
Build Pages Flow using MetdataPresenter Grid

brenetic:
## Use Grid and Route models from the Presenter to build the service flow page

We moved the Route and Grid models to the Presenter gem as we felt that
was a better location for them. Use version 2.5.0.

More details about how the Grid and Route models work can be found in
[this PR](ministryofjustice/fb-metadata-presenter#179). It explains in detail the logic used and gives a visual example.

## Remove OrderedFlow model 

It is no longer used as the Grid model will build the ordered flow

## Remove FlowStack model

We no longer think in terms of just a stack of objects

## Raise error page is missing component

In some rare instances it is possible for a conditional to be saved with
an expression that uses the UUID of a component that does not in fact
exist on the page. This has only happened in testing so we should raise
an error to point it out to the dev as well as see if it happens in
production.

## Subscribe to grid and route active support notifications

The Grid and Route models send an event notification should the number
of potential routes be exceeded of the number of service flow objects
be exceeded when building the grid. We will log these events for now and
see if something needs to be tweaked. If necessary we will send alerts to 
Sentry.


tomas-stefano (approved):
🚀

PR #740, opened 2021-09-29
Previous page title

njseeto:
[Trello](https://trello.com/c/qbSLj1dF/1928-after-previous-page-text-on-branch-edit-page-does-not-display-after-clicking-the-save-button)

### Add branch_uuid accessor
Currently, when a branch has been saved, is it assigned a branch_uuid. However, we still refer to this as a previous_flow_uuid. This is confusing. So we need to have a branch_uuid and reference it when we are dealing with an existing branch.

### Add method to ensure correct title shows when editing branching points
At present, we do not provide information on the previous page when a user is editing a branching point when we should.
In the event that a branch has been 'unconnected' from the main flow (ie. there is no previous page), we do not show any text.
In the event there are multiple pages pointing to the same branch, we take the first page, note this behaviour will need to be further refined by the team but for now it is the agreed approach.

### DIFF
--- a/acceptance/features/edit_branch_page_spec.rb
+++ b/acceptance/features/edit_branch_page_spec.rb
@@ -45,6 +45,7 @@ feature 'New branch page' do
     then_I_can_add_and_delete_conditionals_and_expressions
 
     when_I_save_my_changes
+    then_I_should_see_the_previous_page_title('What is your favourite hobby?')
     then_I_should_see_no_errors
   end
 end
--- a/spec/models/branch_spec.rb
+++ b/spec/models/branch_spec.rb
@@ -324,4 +324,49 @@ RSpec.describe Branch do
       expect(branch.previous_flow_default_next).to eq(previous_flow_object.default_next)
     end
   end
+
+  describe '#branch_attached?' do
+    subject(:branch) { described_class.new(service: service) }
+    let(:latest_metadata) { metadata_fixture(:branching) }
+    let(:service) do
+      MetadataPresenter::Service.new(latest_metadata)
+    end
+    let(:branch_uuid) { '09e91fd9-7a46-4840-adbc-244d545cfef7' }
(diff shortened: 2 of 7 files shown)


acceptance/features/edit_branch_page_spec.rb
```diff
@@ -45,6 +45,8 @@
     then_I_can_add_and_delete_conditionals_and_expressions
 
     when_I_save_my_changes
+    then_I_should_be_on_the_correct_branch_page('edit')
```
brenetic, line 48: @njseeto does this pass for you locally? coz it's currently commented out on line 43

njseeto: Yep this is passing for me locally, I guess that bug has been fixed?

brenetic: only if when line 42 and 43 are uncommented it passes. is that the case if you uncomment them?

njseeto: ok, so when I uncommented ln 42 and 43, they do not pass so the bug must not be fixed yet.
And when I leave ln 42 and 43 commented, but run the test with ln 48, it passes... 🤷 Does it pass for you locally?

spec/models/branch_spec.rb
```diff
@@ -324,4 +324,59 @@
...
+        text_page['next']['default'] = branch_uuid
+
+        star_wars_page = metadata['flow']['68fbb180-9a2a-48f6-9da6-545e28b8d35a']
+        star_wars_page['next']['default'] = branch_uuid
+
+        content_page = metadata['flow']['2cc66e51-2c14-4023-86bf-ded49887cdb2'] # Loki
+        content_page['next']['default'] = branch_uuid
+
+        metadata
+      end
```
tomas-stefano, line 354: You can give one line more between the let and the test

tomas-stefano (approved):

PR #741, opened 2021-09-30
Set operator property in Pages Flow payload

brenetic:
The frontend requires the `operator` value to be set to its own property
in the expression objects that the PagesFlow model builds. This will
make it easier for it to create the question and answer in the UI.


njseeto (approved):
🚀

PR #744, opened 2021-10-04
Correctly set flow object default next when adding a new page

brenetic:
When inserting a page the flow object that represents that page needs to
have its default next point to what the previous object to it was
pointing too.

This has not been a problem until now because we rely on the order of
the pages when presenting the service flow.


njseeto (approved):
🚀

tomas-stefano (approved):

PR #747, opened 2021-10-05
Wait for editor to load

brenetic:
This adds an expectation to the method that clicks on the "Create a new
form" button. This expectation makes the tests wait a little bit for the
page to load which will hopefully reduce the brittleness should there be
any performance related issues.


njseeto (approved):
🚀

PR #748, opened 2021-10-05
Retry Auth0 login during acceptance tests runs

brenetic:
Second try to fix brittle acceptance tests runs. Retry a couple more
times when logging in just in case Auth0 response is a little slow.


njseeto (approved):
🚀

PR #749, opened 2021-10-05
Wait for Auth0 login a bit longer in acceptance tests

brenetic:
Auth0 seems to take a little bit of time to authenticate, so give it a
few more seconds to respond.


njseeto (approved):
🚀

PR #751, opened 2021-10-07
Exit pages

njseeto:
[Trello](https://trello.com/c/xRoJ7LWM/1970-be-adding-the-logic-to-deal-with-exit-pages-in-the-editor)

### Exit pages
- Exit pages can now be added to a form and edited.
- Remove the ability to "Change destination" or "Add page" from an exit page.

### Exit pages should not be available in live production
We will release exit pages to live prod when we release branching.
For now, we only want exit pages to be visible in test.

### Bump presenter 2.6.2
__________

### Editing an exit page 
Screenshot 2021-10-07 at 17 27 59

### DIFF
--- a/app/views/services/edit.html.erb
+++ b/app/views/services/edit.html.erb
@@ -37,7 +37,9 @@
             File.join(preview_service_path(service.service_id), page.url),
             target: '_blank' %>
           </li>
+          <% unless (page.type =~ /page.(checkanswers|confirmation|exit)/) %>
           <li data-action="add"><a href="#add-page-here"><%= t('actions.add_page') %></a></li>
+          <% end %>
           <li data-action="delete">
             <% unless page.type == 'page.start' %>
               <%= link_to t('actions.delete_page'),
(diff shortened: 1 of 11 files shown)


app/views/services/edit.html.erb
```diff
@@ -37,7 +37,9 @@
             File.join(preview_service_path(service.service_id), page.url),
             target: '_blank' %>
           </li>
+          <% unless (page.type =~ /page.(checkanswers|confirmation|exit)/) %>
```
brenetic, line 40: i think we can add a method to the `MetadataPresenter::Page` model that does this for us to save us constantly calling the same thing.

`def allow_add_page?` or something like that. could be in a separate set of PRs

app/views/services/edit.html.erb
```diff
@@ -105,6 +107,7 @@
       <li><a href="#add-page" data-page-type="checkanswers">Check answers page</a></li>
       <li><a href="#add-page" data-page-type="confirmation">Confirmation page</a></li>
       <li><a href="#add-page" data-page-type="content">Content page</a></li>
+      <li><a href="#add-page" data-page-type="exit">Exit page</a></li>
```
brenetic, line 110: sorry one more thing that i forgot. exit pages are going to be released at the same time as branching. so we need to only allow them to be added if `ENV['BRANCHING']` is set to `enabled`.

PR #768, opened 2021-10-13
Order detached grids by size

brenetic:
The detached grids need to be ordered by the length of the the flows,
largest being first.

This will prevent any lone objects appearing first.

## Before

Screenshot 2021-10-13 at 16 30 12

## After

Screenshot 2021-10-13 at 17 58 03


njseeto (approved):
🚀

PR #772, opened 2021-10-15
Add error message for unsupported component

njseeto:
[Trello](https://trello.com/c/icga5Rfp/1985-bug-backend-allows-saving-of-unsupported-questions-operator-and-answers)

Currently, on a branching point, users are able to click save and save unsupported components eg: text, numbers, dates.
This change adds an error on clicking save that prevents this from happening.

Screenshot 2021-10-18 at 10 00 20

### DIFF
--- a/app/models/branch.rb
+++ b/app/models/branch.rb
@@ -96,7 +96,9 @@ class Branch
   end
 
   def flow_uuid
-    previous_flow_uuid || branch_uuid
+    return branch_uuid if previous_flow_uuid.blank?
+
+    previous_flow_uuid
   end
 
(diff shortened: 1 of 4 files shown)


app/models/branch.rb
```diff
@@ -96,7 +96,11 @@ def previous_flow_title
   end
 
   def flow_uuid
-    previous_flow_uuid || branch_uuid
+    if previous_flow_uuid.blank?
```
brenetic, line 99:
```suggestion
    return branch_uuid if previous_flow_uuid.blank?
```

PR #776, opened 2021-10-18
Add version and user to publish services

tomas-stefano:
## Context

We are adding the version id and the user id every time the user clicks publish to test or publish to live.

It means that when we publish we send to the worker the latest version id and the worker will grab the specific version id.

Simple easy and effective.

## Caveats

There will still left the old records in the database with user id and version id empty. We need to check on the test environment what to do with those. (Update manually or leave there with blank user id and version id)


brenetic (approved):
🚀

PR #786, opened 2021-10-21
Adding conditional uuids

tomas-stefano:
[Trello card](https://trello.com/c/Xi5hoqgL/1996-add-a-page-after-a-branch)

## Context

When adding a page after a branch we need to pass two params:

1. add page after
2. conditional uuid

The first is the branch and the second is the conditional route
where the new page will be inserted.

If we add a page after a branch and there is no conditional uuid
the code will look for the otherwise grid (the 'next' default).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'delayed_job_active_record'
 #    github: 'ministryofjustice/fb-metadata-presenter',
 #    branch: 'add-branching-title'
 # gem 'metadata_presenter', path: '../fb-metadata-presenter'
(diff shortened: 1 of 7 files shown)


app/models/conditional.rb
```diff
@@ -16,6 +16,7 @@ def initialize(attributes)
 
   def to_metadata
     {
+      '_uuid' => generate_uuid,
```
brenetic, line 19: now that i think about it, will this not generate a different uuid every time to_metadata is called?

PR #825, opened 2021-10-29
Only show Service URL when it is still published

brenetic:
We only want to show the URL for a service when it is actually
published. Now that we can unpublish a service we do not want a user to
click on the URL only to visit a page that no longer exists.


mattempty (approved):
🚀

PR #828, opened 2021-11-01
Refactor so we can have more visibility on what is going on

tomas-stefano:
## Context

To understand better the error:

```
NotImplementedError
'flow.page' method not implemented
```

We need to know which method name is called and who is the self? (The life question "who is the self?" - "Know thyself")


brenetic (dismissed):
🚀

brenetic (approved):
🚀

PR #829, opened 2021-11-01
Fix broken (seemingly missing) publish buttons

stevenburnell-moj:
Buttons now back as expected.
Screenshot 2021-11-01 at 15 12 36


brenetic (approved):
🚀

PR #831, opened 2021-11-02
Testable 1747 footer flashing

stevenburnell-moj:
No screenshots to show the problem but on flow overview page, due to the increased amount of JS computations and display rearrangements, content that is not yet ready (not styled or not positioned properly) can be briefly seen before JS finishes it's thing. This fix/workaround should prevent that but, since the problem has only been seen on the flow overview page, the fix has been isolated to just that view.


brenetic (approved):
🚀

PR #832, opened 2021-11-02
Testable 1752 component menu

stevenburnell-moj:
Hard to find fix for vanishing menu activator, when the menu is opened. 
Turned out to be a simple fix by changing the order of events (blur causes required css class to be removed but opening of the menu needs it to be there).

Also contains a fix for wrongly positioned three dot menu on certain component (not part of related ticket).

**Without** the fix
Screenshot 2021-11-02 at 16 44 50

**With the fix**
Screenshot 2021-11-02 at 16 44 59


brenetic (approved):
🚀

PR #854, opened 2021-11-10
Change content on change destination modal

tomas-stefano:
## Context

Now we will be adding the After "X" x is the current page the user wants to change the destination.

Also change the label, the button and the hint message.

I create a specific card to [add acceptance tests to change destination feature](https://trello.com/c/86sNF7WT/1790-add-acceptance-tests-for-change-destination).

### DIFF
--- a/app/models/destination.rb
+++ b/app/models/destination.rb
@@ -7,6 +7,10 @@ class Destination
 
   alias_method :change, :create_version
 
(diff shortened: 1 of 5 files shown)


brenetic (dismissed):
🚀

brenetic (approved):
🚀

PR #881, opened 2021-11-16
1872 flow view arrows

stevenburnell-moj:
WIP branch has now been approved for merge into main. 
Leaving all the previous WIP comments as they may be useful at some point.

------ Previous comments ----------------------------------------------------------
Current code allows:
- **ForwardPath**
- **ForwardUpPath**

See screenshot for examples of the above FlowConnectorPaths that can be currently achieved.
Screenshot 2021-11-16 at 13 51 39

Update now allows:
- **ForwardUpForwardDownPath**
Note: Currently no spacing between overlapping paths/lines
Screenshot 2021-11-22 at 13 51 47

Update now allows:
- **DownForwardPath**
Note: BranchCondition position will clearly need altering from the current calculations as a result.
Screenshot 2021-11-24 at 18 09 09

Update now allows:
- **DownForwardUpPath**
Note: A previous height hack for the workaround scroll was causing calculations to be thrown off. To develop this connector type the hack was removed, which has resulted in vertical clipping that can be see in the screenshot on the line going up between 'Decided A' and 'Page 4' thumbnails. If there's an easy fix to prevent this clipping I will add it; otherwise it might have to wait until I'm working on the Scroll ticket, where a proper solution will be implemented.
Screenshot 2021-11-29 at 11 18 44

Update now allows:
- **DownForwardDownBackwardUp**
Note: It is difficult to set this condition up as flow items move around when trying to point a path backward. This is a known and documented issue. Have uploaded a screenshot to show it working on a simple form but need further testing of more instances. Also, in some alternative instances I have managed to test, with much larger forms, the lowest backward line can get clipped by the outside container making it look like an incomplete, rather than just hidden, line. This will need to further adjustments to the height of outside containing element.
Screenshot 2021-12-03 at 17 50 29

Update now allows:
- **ForwardDownBackwardUp**
Note: This is basically a connection between two pages that goes backwards. As with the previous DownForwardDownBackwardUp connector (same thing but between a BranchConnection and a Page) it's hard to get a working scenario. Screenshot shows the setup I had to work with. You can see other issues with lines are starting to creep in so they will get attention next.
Screenshot 2021-12-03 at 17 46 43

Update now allows:
- **DownForwardUpForwardDownPath**
Note: This was discovered to be a new required FlowConnectorPath type during implementation of nudge() functionality for all FlowConnectorPaths, so the update includes the majority of that work (not currently enabled so should go unnoticed but can be seen in the large amount of extra code). See the highlighted line in the attached screenshot for what type has been added.
Screenshot 2021-12-13 at 21 48 17

Update now allows:
**DownForwardDownForwardPath**
Note: This was discovered to be a missing type while debugging and fixing missing/broken paths. See the highlighted line in the attached screenshot for what type has been added.
Screenshot 2021-12-15 at 20 43 14

Update now shows:
Backend work for refinements to layout and (WIP) some initial effort on making FlowConnectorPath components aware of their location data. Note: This code has been pushed to Testable environment by request, so others can have a play with what we have so far.
Screenshot 2021-12-21 at 00 26 52

Update now shows:
Fixed missing/broken path visible in the previous update. Note: This did not need a new path it just needed correction of logic to allow the **ForwardDownBackwardUp** to come into play.
Screenshot 2021-12-22 at 19 13 30

Update now shows:
**WIP commit requested before ready**
This update should not happen yet, as the code is not ready. However, it has been requested to push what is here despite the known bug(s) that exist. In terms of added functionality, essentially this is providing the overlap detection and automatic line adjustment. It works for a small form but has issues with a larger one. The size probably does not actually matter as it is more likely a certain type(s) of line is(are) triggering a recursive loop scenario that, were it not for the coded loop limit, would lock up the browser. 

Screenshot to show a form with adjusted lines:
Screenshot 2022-01-13 at 10 26 35

Screenshot to show same code running against a form that has triggered the recursion bug.
Screenshot 2022-01-13 at 10 30 46

Update now shows:
- Cleans up the previous WIP commit. 
- Fixes the recursion issues (see points on previous screenshot)
Screenshot of same form in previous but not without the need to highlight any issues.
Screenshot 2022-01-13 at 15 19 07

**Final Update (hopefully)**
FlowConnectorPaths with auto-overlap avoidance is now in place for both the Main form overview and each group in the Detached items section. See screenshot showing main and detached areas in play.
Screenshot 2022-01-17 at 11-22-45 MoJ Forms

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -482,15 +482,13 @@ workflows:
           include_job_number_field: false
           requires:
             - acceptance_tests
-            - acceptance_tests_no_branching
-      - confirm_live_deploy:
-          type: approval
-          requires:
-            - acceptance_tests
-            - acceptance_tests_no_branching
-      - deploy_to_live:
-          requires:
-            - confirm_live_deploy
+      # - confirm_live_deploy:
+      #     type: approval
+      #     requires:
+      #       - acceptance_tests
+      # - deploy_to_live:
+      #     requires:
+      #       - confirm_live_deploy
   deploy_testable_branch:
     jobs:
       - build:
--- a/app/javascript/src/component_flow_connector_path.js
+++ b/app/javascript/src/component_flow_connector_path.js
@@ -557,8 +557,6 @@ class DownForwardUpPath extends FlowConnectorPath {
   }
 
   set path(dimensions) {
-    var forward2 = "h" + dimensions.forward2;
-
     var x = this.points.from_x;
     var y = this.points.from_y;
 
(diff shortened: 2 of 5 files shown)


njseeto (approved):
🚀

PR #909, opened 2021-11-25
Testable automate uptime checks

njseeto:
[Trello card](https://trello.com/c/K1ZZo2Uj/2033-automate-pingdom)

## Context

Instead of using the worker queues to automate the creation and deletion
of Uptime Checks we are going to use an admin dashboard page so a
developer can create and destroy them as required.

This page is only available in Live. Unpublishing is a manual process
anyway so we can add the destroying of an Uptime Check to the
unpublishing manual workflow.


app/services/publisher/pingdom_eligibility.rb
```diff
@@ -0,0 +1,36 @@
+class Publisher
+  class PingdomEligibility
```
brenetic, line 2: ### class Uptime::Eligibility

just coz we don't necessarily have to make it specific to pingdom

brenetic: great class name by the way 😆

brenetic:
can we have a PR description etc please

tomas-stefano (approved):
🚀

PR #929, opened 2021-12-01
Acceptance tests delete modals

tomas-stefano:
[Trello card](https://trello.com/c/VOmDeqy5/2087-add-acceptance-for-the-delete-modals)

## Context

This adds scenarios for the acceptance tests related to the deletion of pages and its many scenarios.

### DIFF
--- a/acceptance/features/deleting_page_spec.rb
+++ b/acceptance/features/deleting_page_spec.rb
@@ -19,6 +19,117 @@ feature 'Deleting page' do
     then_I_should_not_see_the_deleted_page_anymore
   end
 
+  scenario 'when try to delete a page which has a branching conditional' do
+    given_I_have_a_form_with_pages
+    when_I_try_to_delete_a_page_which_has_a_branching_conditional
+    then_I_should_see_a_message_that_is_not_possible_to_delete_the_page
+  end
+
+  scenario 'when try to delete a page which result with a stack branch' do
+    given_I_have_a_form_with_pages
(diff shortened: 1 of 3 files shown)


acceptance/features/deleting_page_spec.rb
```diff
@@ -19,6 +19,166 @@
...
+    then_I_should_see_a_message_that_is_not_possible_to_delete_the_page
+  end
+
+  scenario 'when try to delete a page which result with a stack branch' do
+    given_I_have_a_form_with_pages
+    when_I_try_to_delete_a_page_which_result_in_a_stack_branch
+    then_I_should_see_a_message_that_is_not_possible_to_create_stack_branches
+  end
+
+  scenario 'when try to delete a page which is a branch destination' do
```
njseeto, line 34: should this scenario trigger the `delete_modal.delete_branch_destination_page_message`?

tomas-stefano: Yes, but I was focusing more on the behaviour of deleting and changing the destination rather than the message. Because the two scenarios above don't trigger any action (just a message), differently from this one. 
But what do you think?

njseeto: coolio, that's fine just checking that was the intention. It should be ok, looking at it closer looks like you check the `delete_and_update_branching` button which is on the same modal

njseeto (approved):
🚀

PR #935, opened 2021-12-03
New form template

njseeto:
[Trello](https://trello.com/c/G4HR2IxT/2117-change-new-form-default-template-to-have-cya-and-confirmation-5-7)
### New form template 
This adds the CYA and Confirmation pages to the default service template.
This is hidden behind a feature flag so will only appear once branching is released, this is because we require the connection menus to be present.

### Update Acceptance tests 
Tests have been updated to accommodate the new service default template.

### DIFF
--- a/app/generators/new_service_generator.rb
+++ b/app/generators/new_service_generator.rb
@@ -19,6 +19,7 @@ class NewServiceGenerator
     end
 
     metadata['flow'] = start_page_flow_object(metadata)
+    metadata = add_cya_and_confirmation(metadata) if ENV['BRANCHING'] == 'enabled'
     metadata['standalone_pages'] = footer_pages
     metadata
   end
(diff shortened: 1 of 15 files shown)


tomas-stefano (approved):

app/generators/new_service_generator.rb
```diff
@@ -46,4 +47,47 @@ def footer_pages
...
+          'default': ''
+        }
+      }
+    }
+  end
+
+  def add_cya_and_confirmation(metadata)
+    cya_metadata = (DefaultMetadata['page.checkanswers'])
+    .merge('_uuid' => SecureRandom.uuid)
+    .merge('url' => 'checkanswers')
```
brenetic, line 81: little bit difficult to read this syntactically. can we indent the merges please.

also there is no need to wrap the call to `DefaultMetadata` in parenthesis

app/generators/new_service_generator.rb
```diff
@@ -46,4 +47,47 @@ def footer_pages
...
+
+    confirmation_metadata = (DefaultMetadata['page.confirmation'])
+    .merge('_uuid' => SecureRandom.uuid)
+    .merge('url' => 'confirmation')
```
brenetic, line 85: same as my comment above

brenetic:
great stuff, nat. couple of small comments

brenetic (approved):
🚀

PR #936, opened 2021-12-03
Testable legacy name validator

tomas-stefano:
## Context

[Trello card](https://trello.com/c/tH5wcW1d/2098-create-validator-for-form-builder-form-names)

We need to ensure that users cannot generate the same URLs for a form and as we allow people to define their form name this is a risk.

To prevent this from happening we currently check all form names on MoJ Forms but we should also be checking forms on legacy Form Builder too.

To do this we need to run a validation prior to allowing users to save a new form name to ensure we're preventing duplicates.

This will be done by inserting the names in the admin and use a validator to consume from the db.


brenetic (approved):
🚀

PR #947, opened 2021-12-07
Prevent duplicate pages

njseeto:
### Change module name to PageCheckHelper 
This module is now being used outside of the PublishController, name has been changed to a more generic name to reflect this.
 
### Rename methods
This better represents what the method is checking for, which is whether the page exists in the flow specifically.
 
### Prevent duplicates of CYA or Confirmation pages in a service 
A service should only have one CYA and confirmation page.
This commit removes these pages from the 'Add pages' list if they are already present in the service.
 
### Acceptance tests


brenetic (approved):
🚀

PR #948, opened 2021-12-08
Fix acceptance tests

tomas-stefano:
The check your answer and confirmation page now is being generated
when a service is created so tests that are testing the pages
that shows in page flow in its entirity could fail.


brenetic (approved):
🚀

PR #949, opened 2021-12-08
Use Presenter 2.11.0

brenetic:
This includes the update to accommodate visually stacked branching
objects in the pages flow.


njseeto (approved):
🚀

PR #962, opened 2021-12-13
Accept single quotes and apostrophe to service names

tomas-stefano:
[Trello card](https://trello.com/c/gEognZYS/2133-allow-special-characters-like-apostrophe-in-the-form-name)

## Context

We need to accept service names with single quotes/apostrophe so adding specs and the validation


brenetic (approved):
🚀

PR #980, opened 2021-12-16
Do not show link to add CYA or confirmation pages if they exist

brenetic:
Regardless of whether branching is enabled or disabled if a CYA or
Confirmation page already exists the user should not be able to add
another one.


njseeto (dismissed):
🚀

njseeto (approved):

PR #1001, opened 2021-12-23
Remove unused module

brenetic:
This was intended to be used with the Uptime Checks worker jobs but in
the end was not necessary


njseeto (approved):
🚀

PR #1003, opened 2021-12-23
Find latest published services for Uptime Checks

brenetic:
Once a service has been unpublished it's status is no longer
'completed'. Therefore we cannot just grab the completed records for
each service.

We need to get the last records which are in production and then check
whether their status is one of the 'published' statuses.


njseeto (approved):
🚀

PR #1018, opened 2021-12-30
Increase number of replicas for web app

brenetic:
We get a fair number of failures during acceptance tests runs. Increase
the number of replicas to see that helps.

Leave the worker replicas the same.


njseeto (approved):
🚀

PR #1052, opened 2022-01-07
update readme to tweak setup instructions

chrispymm:
Minor updates based on my experience setting it up.  

- Added in a prerequisites list to explicitly state what needs to be installed on the machine prior to setup.
- 
- Tweaked the order of the instructions.


njseeto (approved):
🚀

PR #1061, opened 2022-01-10
Remove Submitting pages disconnect modal

brenetic:
This piece of work has proved to be a little more complicated than
initially planned. Until we have a bit more clarity as to when the
warning modal should appear we are going to take it out.

At some point in the future we will revisit the check answers page and
the confirmation page at a higher level and decide how best to manage
them in the forms.


njseeto (approved):
🚀

PR #1091, opened 2022-01-17
Fix: Optional text bug

stevenburnell-moj:
Fix for
Trello[2204-bug-optional-text-on-collection-components-is-visible-in-preview-on-subsequent-pages]
[BUG] Optional text on collection components is visible in preview on subsequent pages

Allow 'Save' button to be visible on empty/blank entry of editable text item.
Update code to avoid bug caused by GDS changing tag used on Hint text.


brenetic:
can we have a PR message please

brenetic (approved):
🚀

PR #1104, opened 2022-01-19
Parallelise EKS deployment in Test

brenetic:
These do not need to be one after the other as they are entirely
different clusters. Reduces deployment time by nearly 4 minutes

Screenshot 2022-01-19 at 16 18 26


njseeto (approved):
🚀

PR #1108, opened 2022-01-20
Increase acceptance tests wait time for deletion API calls

brenetic:
The acceptance tests are a little flaky when running the CYA and
Confirmation page specs. This might be to do with the time it takes to
call the API.

Increase the wait time to see if it improves the running of the tests.


mattempty (approved):
🚀

PR #1109, opened 2022-01-20
Remove glob-parent

njseeto:
`watchpack-chokidar2` has the `glob-parent 3.1.0` dependency. `watchpack-chokidar2` is an optional dependency of `watchpack` so remove this optional dependency.


brenetic (approved):
🚀

PR #1119, opened 2022-01-24
Add flow view connection menu

chrispymm:
This PR adds the connection menu onto the arrows between flow view items.

[Ticket](https://trello.com/c/LXAaulDZ/2082-flow-view-adding-connection-menu-on-the-arrows-and-remove-old-menu)

image
As it stands the connection menu + button shows all the time for all items.  Desired final behaviour is for it to appear on hovering the arrow, but consensus seems to be that this will be good enough for testing purposes.

image
On a standard page you can perform all actions. N.B.  The option to add a cya or confirmation page is not in the list as they are already present in this form.

image
Connection menu after a branching point has reduced options.

This PR also [fixes the bug](https://trello.com/c/Pe6yiaiK/1941-bug-clicking-on-single-question-page-option-triggers-naming-modal) where clicking the 'Single Question' entry in the menu would show the add page modal and throw an error when the user attempted to create a page with no type.


brenetic:
closed in favour of: https://github.com/ministryofjustice/fb-editor/pull/1158

PR #1123, opened 2022-01-25
Fix publish warning on create

brenetic:
When there is an error raised during the entering of a username and
password during the publishing process the controller will redirect back
to index. However the @publish_warning is not there for the template to
render so this needs to be created before rendering the index template.


mattempty (approved):
🚀

PR #1129, opened 2022-01-26
#2009 bugfix delete branching question

chrispymm:
Fix to prevent deletion of a question on a multi-quetion page that is used for branching.

User is presented with a modal informing them that the question cannot be deleted.

image

Added acceptance tests for this, which also updates the branching spec to have a multiquestion page within it.

### DIFF
--- a/app/javascript/src/component_activated_question_menu.js
+++ b/app/javascript/src/component_activated_question_menu.js
@@ -51,7 +51,16 @@ class QuestionMenu extends ActivatedMenu {
   selection(event, item) {
     var action = item.data("action");
     this.selectedItem = item;
-    safelyActivateFunction(this[action].bind(this));
+
+    event.preventDefault();
+    switch(action) {
+      case "remove":
+           this.remove();
+           break;
+      case "required":
+          this.required();
+          break;
+    }
   }
 
   remove() {
--- a/app/javascript/src/component_activated_question_menu.js
+++ b/app/javascript/src/component_activated_question_menu.js
@@ -31,7 +31,10 @@ class QuestionMenu extends ActivatedMenu {
       onSetRequired: function(){} // Called at end of set required function
     }, config));
 
-    $node.on("menuselect", QuestionMenu.selection.bind(this) );
+    //$node.on("menuselect", QuestionMenu.selection.bind(this) );
+    $node.on("menuselect", (event, ui) => {
+      this.selection(event, ui.item);
+    });
 
     let $target = this._config.$target;
     if($target.length) {
@@ -43,6 +46,12 @@ class QuestionMenu extends ActivatedMenu {
     this.container.$node.addClass("QuestionMenu");
     this.question = config.question;
     this.setRequiredViewState();
+  }
+
+  selection(event, item) {
+    var action = item.data("action");
+    this.selectedItem = item;
+    safelyActivateFunction(this[action].bind(this));
   }
 
   remove() {
@@ -64,15 +73,5 @@ class QuestionMenu extends ActivatedMenu {
     }
   }
 }
-
-/* Handles what happens when an item in the menu has been selected
- * @event (jQuery Event Object) See jQuery docs for info.
- * @data  (Object) See ActivatedMenu and search for config.selection_event
- **/
-QuestionMenu.selection = function(event, ui) {
-  var action = $(event.originalEvent.currentTarget).data("action");
-  safelyActivateFunction(this[action].bind(this));
-}
-
 
 module.exports = QuestionMenu;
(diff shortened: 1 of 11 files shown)


app/javascript/src/component_activated_question_menu.js
```diff
@@ -31,7 +30,10 @@ class QuestionMenu extends ActivatedMenu {
       onSetRequired: function(){} // Called at end of set required function
     }, config));
 
-    $node.on("menuselect", QuestionMenu.selection.bind(this) );
+    //$node.on("menuselect", QuestionMenu.selection.bind(this) );
```
brenetic, line 33: can we take out the comment? unless we're coming back to this soon?

chrispymm: yeah, sorry - that can definitely go.

test/component_activated_question_menu_test.js
```diff
@@ -193,22 +193,10 @@ describe("QuestionMenu", function() {
...
-
-      $target = menu.$node.find("li[data-action=remove]");
-      expect($target.length).to.equal(1);
 
-      $target.click();
-      expect(value).to.equal(2);
-    });
+    /* For some reason only one of these tests would pass.
+     * This one has been removed and replace by acceptance test coverage */
+    it("should trigger QuestionMenuSelectionRemove event on remove()");
```
brenetic, line 199: 👍  can you point to the test file with that scenario please

brenetic (approved):
🚀

app/javascript/src/controller_pages.js
```diff
@@ -215,23 +216,28 @@ function addQuestionMenuListeners(view) {
 
   // QuestionMenuSelectionRemove
   view.$document.on("QuestionMenuSelectionRemove", function(event, question) {
-    var html = $(templateContent).filter("[data-node=remove]").text();
-    view.dialogConfirmationDelete.open({
-      heading: html.replace(/#{label}/, question.$heading.text()),
-      ok: view.text.dialogs.button_delete_component
-      }, function() {
+    var questionUuid = question.data._uuid;
+    var apiUrl = question.menu.selectedItem.data('apiPath').replace('question_uuid', questionUuid);
```
stevenburnell-moj, line 220: Assuming you have a path `/blah/blah/question_uuid/blah` (or something to that effect), where you're String.replacing the 'question_uuid' bit with a value.

I would suggest, if we can, we make the apiPath value more obvious by constructing its initial value in form `/blah/blah/#{question_uuid}/blah` and then search/replacing in form `str.replace(/#{question_uuid}/, question_uuid)`.

Obviously it doesn't make any difference to outcome but, looking at the lines being taken out, it would suggest we're already doing value replacements with that form of lookup identifier already (albeit with HTML in that example), so good idea to follow with same approach for consistency (sure there are others in and around code as well although, without documentation or any rules-based coding, appreciate it would not be easy to know this).

Note: I believe `#{something}` form is based on Ruby/Rails syntax although also commonly found in other languages and environments so hopefully a familiar thing many.

app/javascript/src/controller_pages.js
```diff
@@ -380,18 +386,6 @@ function enhanceQuestions(view) {
         }
       },
 
-      onItemRemoveConfirmation: function(item) {
-        // @item (EditableComponentItem) Item to be deleted.
-        // Runs before onItemRemove when removing an editable Collection item.
-        // Currently not used but added for future option and consistency
-        // with onItemAdd (provides an opportunity for clean up).
-        view.dialogConfirmationDelete.open({
-          heading: view.text.dialogs.heading_delete_option.replace(/%{option label}/, item._elements.label.$node.text()),
```
stevenburnell-moj, line 389: Ha! Looks like I had muddied the water on my other comment about using `#{something}` form of lookup label, as this one used `%{something_else}` to identify the swappable part. Maybe I had some 'this one is for templates' and 'this one is for JS files' kind of thinking going on. Overly complex. Who cares now, since the lines are being removed :-)

app/javascript/src/controller_pages.js
```diff
@@ -215,23 +216,28 @@ function addQuestionMenuListeners(view) {
...
+    new DialogApiRequest(apiUrl, {
+      activator: question.menu.selectedItem,
+      closeOnClickSelector: ".govuk-button",
+
+      build: function(dialog) {
+        // Find and correct (make work!) any method:delete links
+        dialog.$node.find("[data-method=delete]").on("click", function(e) {
+          e.preventDefault();
       // Workaround solution that doesn't require extra backend work
       // 1. First remove component from view
```
stevenburnell-moj, line 231: Like the attention to detail on indenting code but, is there a reason you didn't indent the comment as well?
Screenshot 2022-01-26 at 12 15 16

stevenburnell-moj (approved):

PR #1130, opened 2022-01-26
Added another domain to the valid email address domains

mattempty:
Edge case of a volid email address having a capitialised J which
caused the editor to return an error.

This fix will allow both justice.gov.uk and Justice.gov.uk to be
valid and allow access.


brenetic (approved):
🚀

PR #1141, opened 2022-01-28
Fixes background colour of "add condition" button

f-marry:


app/javascript/styles/_index.scss
```diff
@@ -255,6 +255,7 @@ html {
 .BranchConditionInjector {
   @include button_as_link;
   margin-bottom: govuk-spacing(5);
+  background-color: rgb(243, 242, 241);
```
stevenburnell-moj, line 258: If the intention is to match the button (BranchConditionInjector) background colour to match the container background colour, there are three better background-color values achieve it:

1. `govuk-colour("light-grey")`
...the exact (GDS colour variable) value used by the parent container.

2. `inherit` 
...tell it to explicitly use the same as the parent element.  

3. `transparent` 
...just make is see-through.
 

Take your pick but I'd probably plump for option 2 or 3, as then it will also change if ever the parent is changed (so remains the same as parent colour effect you are looking to achieve).

stevenburnell-moj (changes requested):

f-marry:
Closing this and reopening so it passed integration

PR #1158, opened 2022-02-01
Testable flow view connection menu

chrispymm:
This PR adds the connection menu onto the arrows between flow view items. ([Ticket](https://trello.com/c/LXAaulDZ/2082-flow-view-adding-connection-menu-on-the-arrows-and-remove-old-menu))

image

As it stands the connection menu + button shows all the time for all items.  Desired final behaviour is for it to appear on hovering the arrow, but consensus seems to be that this will be good enough for intial launch.

image
On a standard page you can perform all actions. N.B.  The option to add a cya or confirmation page is not in the list as they are already present in this form.

image
Connection menu after a branching point has reduced options.

image
After a start page doesn't include the 'Add branching' option.

image
Items at the end of a route flow will show a trailing arrow with connection menu.
After a check your answers page the only allowed option is to add a confirmation page (or, if the confirmation page is disconnected, then the action is to reattach it)


app/views/services/_connection_menu.html.erb
```diff
@@ -0,0 +1,111 @@
...
+
+      <li data-action="destination">
+        <%= link_to t('actions.change_destination'),
+        new_api_service_flow_destination_path(service.service_id, item[:uuid]) %>
+      </li>
+    <% end %>
+
+  <% else %>
+
+    <% unless item[:type] == 'page.checkanswers' %>
```
stevenburnell-moj, line 74: At a glance, it looks like we could reduce some duplication here but I wouldn't let that stand in the way of getting this merged. We can deal with idealistic coding at a later date, if need be.

app/views/services/_flow_layout.html.erb
```diff
@@ -22,7 +22,7 @@
   <% else %>
 
     <% if item[:type] == 'flow.branch' %>
-      <article class="flow-item flow-branch" aria-label="TODO-need-editable-label" data-fb-id="<%= item[:uuid] %>">
+      <article class="flow-item flow-branch" aria-label="<%= item[:title] %>" data-fb-id="<%= item[:uuid] %>">
```
stevenburnell-moj, line 25: I'd personally expect aria-label to have better use if it was to describe what this entity is, rather than just duplicating the item title. Also, if we're just using the item title then we might as well use aria-labelledby attribute instead. As per the previous comment on another file, however, it's not a big enough issue to block merging. We can improve things later.

stevenburnell-moj (approved):

app/views/services/_flow_layout.html.erb
chrispymm: Agree its not perfect - I was only looking at this the other day and questioning whether the `<article>` should even have the `aria-label` attribute... but I think I figured *something* was better than 'TODO' and that we can iron out accessibility stuff later.

PR #1159, opened 2022-02-01
Fix Misalignment of FlowConnectorPath on page refresh

stevenburnell-moj:
- Fixes issue when reloading a form overview page after scrolling or when things are redrawn on browser resize.
- Change to make the detached flow section open on page load.
- Tweaks to unconnected group spacing and title content

### Tickets
https://trello.com/c/J81Q6pQe/2011-unconnected-pages-section-to-default-to-expanded-open-61-xs
https://trello.com/c/oSDI1ZMx/2248-bug-misalignment-of-flowconnectorpath-on-page-refresh-s
https://trello.com/c/HHYMcdvi/2254-unconnected-section-layout-1

### Screenshots

**(with bug/was)**
Screenshot 2022-01-21 zoomed out but after scroll page reload

**(fixed/now)**
Screenshot 2022-01-21 zoomed out but initial page load

### DIFF
--- a/app/javascript/src/controller_services.js
+++ b/app/javascript/src/controller_services.js
@@ -586,7 +586,6 @@ function adjustOverviewWidth($overview) {
  * when there are too many thumbnails to fix on the one page view.
  **/
 function applyOverviewScroll($overview) {
-  const margin = 30; // TODO: Maybe we can get this number from something??
   var $container = $("<div></div>");
   var $main = $("#main-content");
   var timeout;
--- a/app/javascript/src/controller_services.js
+++ b/app/javascript/src/controller_services.js
@@ -376,11 +376,10 @@ function layoutDetachedItemsOveriew(view) {
   var $container = view.$flowDetached;
   var $title = $("h2", $container);
   var offsetLeft = $container.offset().left;
-  var $expander = $(".Expander_container");
-  var display = $expander.css("display");
+  var expander = $container.data("instance"); // Element is set as an Expander Component.
 
-  // display:none objects have no height in jQuery
-  $expander.css("display", "block");
+  // Make sure it's open on page load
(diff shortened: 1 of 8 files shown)


app/javascript/src/controller_services.js
```diff
@@ -562,68 +614,52 @@ function adjustOverviewHeight($overview) {
...
-      clearTimeout(scrollTimeout);
-      adjustOverviewScrollDimensions($overview, $container);
-    }, 500);
+    clearTimeout(timeout);
+    $main.css("visibility", "hidden");
+    timeout = setTimeout(function() {
+      $container.get(0).style = ""; // reset everything
+      adjustScrollDimensionsAndPosition($container);
+      $main.css("visibility", "visible");
+    }, 1500);
```
chrispymm, line 633: 1.5s seems quite a high timeout on resize - is there a reason for it being so large?

chrispymm (approved):

njseeto (approved):
🚀

app/javascript/src/controller_services.js
stevenburnell-moj: It's an arbitrary number so figure is not based on a calculation. Reason for a slightly larger number is because this can be triggered multiple times by either user rotating a tablet device (although we're not aiming for that audience, yet) or by resizing the browser (e.g. drag the window).

Because human interaction (by intention or accident) is involved, the higher number tries to ensure things have settled down before then going ahead with an attempt to redraw the view.

In other words, the lower the number, the more chance there is of the view trying to redraw while other events are being triggered (screen still being dragged, etc.). Again, no calculation; just guesswork, hope, and experimenting to try and get a "pause... and... refresh" feel over a "redraw! redraw! redraw!!" experience.

PR #1166, opened 2022-02-02
#2222 reconnect confirmation page

chrispymm:
This adds a trailing arrow on form routes that don't end in a confirmation page or exit page. 
[Ticket](https://trello.com/c/awSx6IDL/2222-adding-pages-at-the-end-of-the-flow-when-cya-conf-are-not-present-s)

Will show 'Add confirmation page' in the connection menu after a check your answers page
image

Will show 'Reconnect confirmation page' if the confirmation page has become unconnected. This will perform a post request to set the destination of the check your answers page to the confirmation page.
image

### DIFF
--- a/app/views/services/_connection_menu.html.erb
+++ b/app/views/services/_connection_menu.html.erb
@@ -42,9 +42,10 @@
       
     <% if service.confirmation_page.present? %>
       <% if !@publish_warning.confirmation_in_main_flow? %>
-        <li data-page-type="confirmation" data-action="destination">
-          <%= link_to 'Reconnect confirmation page',
-                      new_api_service_flow_destination_path(service.service_id, item[:uuid]) %>
+        <li data-action="reconnect-confirmation"
+            data-url="<%= api_service_flow_destinations_path(service.service_id, item[:uuid]) -%>"
+            data-destination-uuid="<%= service.confirmation_page.uuid -%>">
+          <a href="#reconnect-confirmation">Reconnect confirmation page</a>
         </li>
       <% end %>
     <% else %>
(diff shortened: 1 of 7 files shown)


app/views/services/edit.html.erb
```diff
@@ -22,8 +22,9 @@
 <% if ENV['BRANCHING'] == 'enabled' && @detached_flows.present? %>
   <section id="flow-detached" data-component="Expander">
     <h2 class="govuk-heading-s"><%= t('pages.flow.detached') %></h2>
+
     <% @detached_flows.each do |detached_flow| %>
-      <div class="flow-detached-group">
+          <div class="flow-detached-group">
```
brenetic, line 27: 🙈

app/views/services/_connection_menu.html.erb
```diff
@@ -68,7 +80,7 @@
           <li data-page-type="singlequestion"
               data-component-type="textarea"><a href="#add-page">Textarea</a></li>
           <li data-page-type="singlequestion"
-              data-component-type="email"><a href="#add-page">Email address</a></li>
+              data-component-type="mail"><a href="#add-page">Email address</a></li>
```
brenetic, line 83: in terms of what the backend knows this component as it is `email`. does the frontend use this differently?

brenetic:
great stuff 🚀 

i would be ok if the last three commits were just one. not a blocker though

njseeto:
LGTM! I just have a similar question re: `data-component-type="mail"` but otherwise it looks great!

app/views/services/_connection_menu.html.erb
chrispymm: No, this looks like I've mistakenly managed to remove the e 🤦‍♂️

PR #1168, opened 2022-02-02
add conditional_uuid into create page modal on branching condition

chrispymm:
Adds the frontend work required to complete [ticket #2249](https://trello.com/c/ty7Eg0cn/2249-enable-add-page-after-a-branching-point-from-the-connection-menu-s)

When on a branch conditional the connection menu populates the hidden form field with the current conditional_uuid allowing the new page to be created in the correct place in the grid.

image

image


brenetic (approved):
🚀

PR #1169, opened 2022-02-02
replace action for change destination in connection menu

chrispymm:
Fixes mistaken removal of the change destination action on the connection menu.


brenetic (approved):
🚀

njseeto (approved):
🚀

PR #1170, opened 2022-02-02
add or statements in between branches

chrispymm:
Injects 'ors' between branches when editing an existing branching point. [ticket 2255](https://trello.com/c/AGJrD9N8/2255-or-statement-doesnt-show-between-branches-4-xs)

Before:
image

After:
image

Also removes the 'ors' when deleting a branch.  This also fixes a bug where if you added and then deleted a branch when creating branching, you'd see double 'ors'

Result after adding and immediately deleting a branch before:
image

Now:
image

### DIFF
--- a/app/javascript/src/component_branch.js
+++ b/app/javascript/src/component_branch.js
@@ -83,8 +83,14 @@ class Branch {
   }
 
   destroy() {
+    // find preceding 'or' text to remove as well
+    var $or = this.$node.prev('.branch-or').first();
     this.$node.remove();
+    if($or) {
+      $or.remove()
+    }
   }
+  
 }
 
 
--- a/app/javascript/src/component_branch.js
+++ b/app/javascript/src/component_branch.js
@@ -42,11 +42,11 @@ class Branch {
     });
 
     this._config = conf;
-    this._index = Number(conf.branch_index);
     this._conditionCount = 0;
     this._conditions = {}; // Add only conditions that you want deletable to this.
     this.$node = $node;
     this.view = conf.view;
+    this.index = Number(conf.branch_index);
     this.destination = new BranchDestination($node.find(config.selector_destination), conf);
     this.conditionInjector = new BranchConditionInjector($injector, conf);
     this.remover = new BranchRemover($remover, conf);
@@ -66,7 +66,7 @@ class Branch {
 
   addCondition() {
     var template = utilities.stringInject(this._config.template_condition, {
-      branch_index: this._index,
+      branch_index: this.index,
       condition_index: ++this._conditionCount
     });
 
@@ -83,12 +83,8 @@ class Branch {
   }
 
   destroy() {
-    // find preceding 'or' text to remove as well
-    var $or = this.$node.prev('.branch-or').first();
+    $(document).trigger('branchRemoved', this.$node );
     this.$node.remove();
-    if($or) {
-      $or.remove()
-    }
   }
   
 }
--- a/app/javascript/src/controller_branches.js
+++ b/app/javascript/src/controller_branches.js
@@ -124,7 +124,7 @@ BranchesController.addBranchMenu = function(args) {
   var first = $(".Branch", $form).get(0) == branch.$node.get(0);
   if(!first) {
     new ActivatedMenu($ul, {
-      activator_text: "Edit branch",
+      activator_text: app.text.branches.branch_edit,
       container_classname: "SomeClassName",
       container_id: utilities.uniqueString("activated-menu-"),
       menu: {
--- a/app/javascript/src/controller_branches.js
+++ b/app/javascript/src/controller_branches.js
@@ -78,6 +78,11 @@ BranchesController.enhanceCurrentBranches = function($branches) {
     if(index == 0) {
       branch.$node.find(".BranchRemover").eq(0).hide();
     }
+    
+    // Inject or statements between each of the branches 
+    if(index != 0) {
+      branch.$node.before("<p class=\"branch-or\">or</p>");
+    }
   });
 }
 
@@ -119,7 +124,7 @@ BranchesController.addBranchMenu = function(args) {
   var first = $(".Branch", $form).get(0) == branch.$node.get(0);
   if(!first) {
     new ActivatedMenu($ul, {
-      activator_text: "Activator text here",
+      activator_text: "Edit branch",
       container_classname: "SomeClassName",
       container_id: utilities.uniqueString("activated-menu-"),
       menu: {
--- a/app/javascript/src/controller_branches.js
+++ b/app/javascript/src/controller_branches.js
@@ -52,6 +52,8 @@ class BranchesController extends DefaultController {
    /* Setup view for the create (new) action
    **/
   create() {
+    var view = this;
+
     var $branches = $(BRANCH_SELECTOR).not(BRANCH_OTHERWISE_SELECTOR);
     var $injectors = $(BRANCH_INJECTOR_SELECTOR);
     var $otherwise = $(BRANCH_OTHERWISE_SELECTOR + " " + BRANCH_DESTINATION_SELECTOR);
@@ -62,6 +64,8 @@ class BranchesController extends DefaultController {
     BranchesController.enhanceCurrentBranches.call(this, $branches);
     BranchesController.enhanceBranchInjectors.call(this, $injectors);
     BranchesController.enhanceBranchOtherwise.call(this, $otherwise);
+
+    addBranchEventListeners(view)
   }
 }
 
@@ -78,11 +82,7 @@ BranchesController.enhanceCurrentBranches = function($branches) {
     if(index == 0) {
       branch.$node.find(".BranchRemover").eq(0).hide();
     }
-    
-    // Inject or statements between each of the branches 
-    if(index != 0) {
-      branch.$node.before("<p class=\"branch-or\">or</p>");
-    }
+
   });
 }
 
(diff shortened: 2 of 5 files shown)


app/javascript/src/controller_branches.js
```diff
@@ -119,7 +124,7 @@ BranchesController.addBranchMenu = function(args) {
   var first = $(".Branch", $form).get(0) == branch.$node.get(0);
   if(!first) {
     new ActivatedMenu($ul, {
-      activator_text: "Activator text here",
+      activator_text: "Edit branch",
```
brenetic, line 127: does the JS have access to the `I18n` to make use of the translations?

brenetic (approved):
🚀

app/javascript/src/controller_branches.js
chrispymm: Good point.  I'eve updated this to use the locales text.

brenetic (approved):
🚀

app/javascript/src/controller_branches.js
```diff
@@ -78,6 +78,11 @@ BranchesController.enhanceCurrentBranches = function($branches) {
     if(index == 0) {
       branch.$node.find(".BranchRemover").eq(0).hide();
     }
+    
+    // Inject or statements between each of the branches 
+    if(index != 0) {
+      branch.$node.before("<p class=\"branch-or\">or</p>");
```
stevenburnell-moj, line 84: Looks like inspiration for this approach might have come from the view-specific BranchInjector class, which also has a line `$node.before("<p class=\"branch-or\">or</p>")` featuring in an AJAX response handler. 

Can't say I'm convinced the earlier approach is right and, if it was the inspiration, then it shows how a small error can easily become bigger. With this new addition, we now have two places inside JS code that are injecting DOM elements that could/should be handled from a template. 

I'm wondering if adjusting view templates to include the branch-or element would be the fix required to remove both these JS-based DOM injections.

app/javascript/src/component_branch.js
```diff
@@ -83,8 +83,14 @@ class Branch {
   }
 
   destroy() {
+    // find preceding 'or' text to remove as well
+    var $or = this.$node.prev('.branch-or').first();
```
stevenburnell-moj, line 87: '.branch-or' element is design requirement and, I think, not something a Branch object should need to concern itself about (because it also doesn't know of other Branch objects as that is a view controller responsibility).

If view alterations are required, we should put an event broadcaster inside the destroy() method, so that the view controller (set up with a listener) can adjust the view when required.

app/javascript/src/controller_branches.js
```diff
@@ -119,7 +124,7 @@ BranchesController.addBranchMenu = function(args) {
   var first = $(".Branch", $form).get(0) == branch.$node.get(0);
   if(!first) {
     new ActivatedMenu($ul, {
-      activator_text: "Activator text here",
+      activator_text: app.text.branches.branch_edit,
```
stevenburnell-moj, line 127: Well spotted.

app/javascript/src/controller_branches.js
chrispymm: I had a feeling you might say soemthing along these lines 😂 I mostly just continued with the approach as I was treating it as a quick bugfix rather than a refactor, and didn't necessarily want to break stuff in the process.  But I'm inclined to agree that this approach isn't necessarily the best.  I'll revisit it and attempt to improve.

app/javascript/src/controller_branches.js
```diff
@@ -171,6 +177,26 @@ BranchesController.createBranch = function($node) {
...
+  }
+}
+
+BranchesController.removeBranchCombinator = function(node) {
+    var $or = $(node).prev('.branch-or').first();
+    if($or) {
+      $or.remove()
+    }
+}
+
```
stevenburnell-moj, line 193: Since we're dealing with ES6 (same as nodejs) environments rather than older JS approach (ie. each file has it's own scope rather than, by default, sharing the same window scope), there is no absolute requirement to create functions in the form:
```
 BranchesController.someSpecificControllerFunctionHere = function() {
   // blah
 }
```

This is a habit I developed with older JS code to indicate the function is a Class-level method (not instance) and not some random function (which, in older JS would have resulted in window.myFunctionExistsHere scope).

The intention was to show the explicit relationship with the inner workings of the Class (in this case the BranchController). 

In the current JS (ES6) environment, these two functions share the same scope, so you could argue the the Class name prefix version is not required:
```
 BranchesController.someSpecificControllerFunctionHere = function() {
   // blah
 }
 
 function addBranchEventListeners() {
   // blah
 }
```

As said though, the one added to the Class is trying to show it should not be extracted out to its own file (because sometimes things were thrown into the same file by error/time-saving/lack of judgement, etc.) and could be better.

Hopefully that explains the though-process behind the `BranchesController.something` approach. 

All that aside, however, there is this code inside the BranchController#create aciton:
```
BranchesController.enhanceCurrentBranches.call(this, $branches);
BranchesController.enhanceBranchInjectors.call(this, $injectors);
BranchesController.enhanceBranchOtherwise.call(this, $otherwise);

addBranchEventListeners(view)
```

To follow with principles of code consistency, it should probably be:
```
BranchesController.enhanceCurrentBranches.call(this, $branches);
BranchesController.enhanceBranchInjectors.call(this, $injectors);
BranchesController.enhanceBranchOtherwise.call(this, $otherwise);
BranchesController.addBranchEventListeners(this);
```
or change those added directly onto the class to the same non _BranchesController.something_ form.

app/javascript/src/controller_branches.js
```diff
@@ -171,6 +177,26 @@ BranchesController.createBranch = function($node) {
...
+function addBranchEventListeners(view) {
+  view.$document.on('branchRemoved', function(event, node){
```
stevenburnell-moj, line 195: A quick `grep -r "\$document\.on" app/javascript/src` shows these results:
```
view.$document.on("AddComponentMenuSelection", AddComponent.MenuSelection.bind(view) );
this.$document.on("SaveRequired", () => dataController.saveRequired(true) );
view.$document.on("QuestionMenuSelectionRemove", function(event, question) {
view.$document.on("QuestionMenuSelectionRequired", function(event, question) {
view.$document.on("ContentMenuSelectionRemove", function(event, component) {
```

I'm surprised because I expected them to be all lowercase (which is why I checked in the first place). Either way, although a small point it's probably worth sticking with the upper camel case/pascal case version for code consistency.

I'm also guessing that it's the Branch.remove method that is triggering the event so, not that I'd say there is any hard and fast event naming convention but, it would seem, we can let one evolve. Looking at others, maybe it's evolving to be ThingMethod or ThingActionMethod ??

'BranchRemove' seems closest to establishing/keeping a convention of some sort.

app/javascript/src/controller_branches.js
```diff
@@ -171,6 +177,26 @@ BranchesController.createBranch = function($node) {
...
+BranchesController.addBranchCombinator = function(args) {
+  var branch = args[0];
+  if( !branch.index == 0 ) {
+    branch.$node.before("<p class=\"branch-or\">or</p>");
+  }
+}
+
+BranchesController.removeBranchCombinator = function(node) {
+    var $or = $(node).prev('.branch-or').first();
+    if($or) {
```
stevenburnell-moj, line 189: The _if_ condition will always be true since jQuery returns either a populated or empty collection.
To achieve your check you can update it to:
```
if($or.length) {
}
```

A failed checked will have the value 0 (also resolves to false in JS) but if anything is found the value will be greater (so handled as though true). 

Since jQuery handles a lot of error handling for you, it should also be possible to just do this:
`$(node).prev('.branch-or').first().remove();`

Should nothing be found, because jQuery is returning an empty collection, the jQuery code will handle the lack found element gracefully and without complaint (nothing will be removed).

app/javascript/src/controller_branches.js
```diff
@@ -171,6 +177,26 @@ BranchesController.createBranch = function($node) {
   return branch;
 }
 
+BranchesController.addBranchCombinator = function(args) {
+  var branch = args[0];
+  if( !branch.index == 0 ) {
```
stevenburnell-moj, line 182: `if( branch.index != 0 )` is easier to read.

See also the other comment about exposing private for public usage since this value can easily be given a different value elsewhere by (undesirable) intention or accident (e.g. mis-type `if( branch.index = 0 )` and you've got a nice bug on your hands).

app/javascript/src/controller_branches.js
```diff
@@ -151,6 +156,7 @@ BranchesController.createBranch = function($node) {
     template_condition: this._branchConditionTemplate,
     event_on_create: function(branch) {
```
stevenburnell-moj, line 157: **Scope for code cleanup/improvement??**

Quite possibly, this earlier method of making things happen on certain events is now becoming redundant or outclassed by the method you've just used for this code change...
`view.$document.on('branchRemoved', function`

To avoid using two different methods of triggering event handling, I'd be voting for one or the other.

This old method clutters the config so, there would be an argument for updating it to fit with a similar approach to what you've done on BranchRemove, while we have the opportunity to change code in this file.

stevenburnell-moj (changes requested):

app/javascript/src/component_connection_menu.js
```diff
@@ -0,0 +1,109 @@
+const utilities = require('./utilities');
```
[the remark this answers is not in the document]
chrispymm: yep, this was an accidental add, and has since been removed.

app/javascript/src/controller_branches.js
chrispymm: I have no explanation for that if conditional, not sure why I wrote it like that.

This was the main motivation for the switch from 'private' to 'public' for the index var. But keeping it private and using a getter is how it should be done.

app/javascript/src/controller_branches.js
chrispymm: I think given that there is another class `BranchInjector` within the same file.  Keeping the `BranchesController.` format is more explicit.

chrispymm:
Closing this PR in favour of one from a newer testable version of the branch.  Sorry for the confusion.

PR #1172, opened 2022-02-02
Use browser tools 1.2.4

brenetic:
Update browser tools to 1.2.4

Also only install chrome and chrome driver. This stops Firefox and
Geckdriver from installing which should reduce the run time of the
acceptance tests a little bit.


njseeto (approved):
🚀

PR #1180, opened 2022-02-03
Add modal for deleting branch destination with no default next 

brenetic:
When a page which is the destination of a branching condition has no
default next if a user was to delete it there is nothing to take its'
place. This isn't really the state we want a branching point to ever get
to.

Instead we will check if the branch destination being deleted has a
default next. If it does not we show the user a modal letting them know
they cannot delete the page until they update the branch to point to a
different page manually.

Screenshot 2022-02-03 at 15 52 31


chrispymm (approved):
All looks good to me.

PR #1185, opened 2022-02-04
reconnect confirmation page

chrispymm:
Updates on [ticket #2222](https://trello.com/c/awSx6IDL/2222-adding-pages-at-the-end-of-the-flow-when-cya-conf-are-not-present-s)

 - 'Add confirmation page' should only be present in the connection menu after a cya page, if there is no unconnected confirmation page, else it would be 'reconnect confirmation page'

 - All other connection menus would never contain 'Add confirmation page'

 - Other connection menus could contain 'Add check your answers page' if one is not already present in the form (connected or unconnected)

 - As an extra bonus, this should now only show the cya page as an option if you are at the end of the flow.

image

image

### DIFF
--- a/app/javascript/src/component_connection_menu.js
+++ b/app/javascript/src/component_connection_menu.js
@@ -47,6 +47,9 @@ class ConnectionMenu extends ActivatedMenu {
       case "link": 
         this.link(item);
         break;
+      case "destination":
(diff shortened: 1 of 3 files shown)


stevenburnell-moj (approved):

brenetic (approved):
🚀

PR #1197, opened 2022-02-08
add correct classes to cancel button to allow it to trigger modal close

chrispymm:
Very simple PR to add the correct css classes to the cancle button in the delete branch modal in order ensure the event listener for triggering modal closure is attached.


brenetic (approved):
🚀

PR #1218, opened 2022-02-10
Distinguish unconnected pages in Change Destination modal

njseeto:
[Trello](https://trello.com/c/LcdnNJiz/2284-change-next-page-modal-should-distinguish-pages-that-are-unconnected-2-xs)

# Distinguish unconnected pages in Change Destination modal

## Before
before list
before

## After
after list
Screenshot 2022-02-10 at 12 58 15


brenetic (approved):
🚀

PR #1219, opened 2022-02-10
Use Alpine 3.13 and Node 14.19.0

brenetic:
Stop pinning the Node version and just install the LTS version.

Using Alpine 3.13 just passing `nodejs` when adding the APK
package will retrieve version 14.19.0 at time of writing.

Using Alpine 3.15 retrieves version 16.XX.XX. Currently using this
version of Alpine we have an issue where the `WORKDIR` is ignored when
creating the container and dependencies are instead installed to a temp
directory which the container has no permissions to use.

Until we can debug why this is happening further just move back to using
Alpine 3.13 and Node 14.19.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
           name: Swap node versions
           command: |
             set +e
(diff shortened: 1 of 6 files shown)


njseeto (approved):
🚀 🏅

PR #1229, opened 2022-02-14
2105 improve scrolling 3

stevenburnell-moj:
Another possible alternative to implementing a scrolling feature for the flow layout page.
This is not 100% complete as (my opinion) it still needs the fixed header to move when page is scrolled vertically. There may be no easy way to do this other with yet more JS doing the work of CSS :-( 

**Page on load**
Screenshot 2022-02-14 at 02 06 42

**Page scrolling horizontally**
Screenshot 2022-02-14 at 02 06 51

**Page scrolling vertically**
(notice the content moving over the header, not the header moving with content - don't like this!)
Screenshot 2022-02-14 at 02 07 07


stevenburnell-moj:
Closing this PR because have replaced it with https://github.com/ministryofjustice/fb-editor/pull/1305 after stuffing up conflict resolution and deciding to take an easier approach.

PR #1246, opened 2022-02-17
Publish forms to EKS cluster

brenetic:
Set the identifier to green in order for ingress to point to the EKS
cluster when a service is published out.


mattempty (approved):
🚀

mattempty (approved):
🚀

PR #1252, opened 2022-02-18
Adjust layout of branch conditions on flow view

chrispymm:
Resolves [ticket #2283](https://trello.com/c/J459SIHl/2283-layout-of-the-condition-14-12-17-46-s)

- Can now hover flow connector line under/by branch condition
- Adds in `AND` between multiple condition expressions on single branch
- Avoids overlapping condition expressions on single branch
- Moves connection menu to the right of the branch condition

**Before:**
image

**After:**
image


stevenburnell-moj (changes requested):
Branch Conditions are not exclusive to Connection Menu and are already in `main` so any fix/update to those should be applied to a working branch built from `main`. 

Once those changes are merged, the work can be used to update the `testable-flow-view-connection-menu` branch, and the extra little bits relating to Connection Menus (that obviously only live on that branch) should then be applied.

This would reduce the potential for file conflicts, should anything overlap with other working being carried out against `main`, and also get the bulk of the required fix into the place where it should be, rather than having yet more lingering on a long-lived branch.

chrispymm:
You are correct that branch conditions are also in main, but this fix actually does stuff realting to the connection menu too - it adjusts the position of the connection menu, plays around with the z-indexes of lines and menus, as well as updating the branch condition /flow expression.  

Given that live has branching behind the feature flag, and no-one can actaully see these branch conditions is there much point splitting this fix into two PRs?  One that simply adjusts the flow expressions and gets applied to main and then another that changes the conneciton menu parts on testable?  Seems a bit unnecessary to me...

PR #1261, opened 2022-02-21
Reference deployments in HPA configuration

brenetic:
The Editor has two types of deployments; web and workers. When
configuring a HPA it needs to be correctly pointed to the corresponding
deployment configuration otherwise it will not be applied correctly


mattempty (approved):
🚀

PR #1262, opened 2022-02-21
Reenable Live deployment

brenetic:
This puts back the deployment to the Live environment. The old cluster
ingress is set to 0 so traffic will be diverted to the editor running in
the new cluster.


mattempty (approved):
🚀

PR #1264, opened 2022-02-21
Add `contains` and `does_not_contain` operator

njseeto:
[Trello](https://trello.com/c/5LKINamZ/2334-condition-using-a-checkbox-operator-is-incorrect-is-isnt-instead-of-contains-doesnt-contain-38-s)

This is a change as a result of user testing.
When the component is a checkbox, we should use `contains` and `does_not_contain` instead of `is` and `is_not` options.

### Checkboxes
Screenshot 2022-02-18 at 13 35 21

### Radios
Screenshot 2022-02-18 at 13 34 46


chrispymm (approved):
All LGTM 🚀

PR #1268, opened 2022-02-22
Change destination drop down order

njseeto:
[Trello](https://trello.com/c/WSe8eQmT/2341-change-next-page-unconnected-page-order-is-odd-66-s)
Ensure the order for the Change Destination drop down list is the same as the order in the Unconnected section.

### Before
before

### After
after

### DIFF
--- a/app/models/destination.rb
+++ b/app/models/destination.rb
@@ -30,7 +30,7 @@ class Destination
   end
 
   def detached_destinations
-    destinations_list(flow_objects: ordered_detached_objects, current_uuid: flow_uuid)
+    destinations_list(flow_objects: detached_objects.ordered, current_uuid: flow_uuid)
   end
 
   private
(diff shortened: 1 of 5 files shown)


app/models/destination.rb
```diff
@@ -43,6 +43,11 @@ def detached_objects
     Detached.new(
       service: service,
       main_flow_uuids: grid.flow_uuids
-    ).flow_objects
+    )
+  end
+
+  def detached_flow_objects
```
brenetic, line 49: how does this differ from `flow_objects` method on the `Detached` model? they sound effectively the same, but perhaps there is a difference? if so then their names should not be so similar. i would be tempted to ask can you not call `detached_objects.flow_objects` on line 33, but perhaps i am missing a nuance

njseeto: Good question, calling `flow_objects` on the `Detached` model doesn't order them correctly, which is why we are currently seeing the drop down list out of sync with the Unconnected page section. The `detached_flows` method on the `Detached` model does this ordering for me. However, I am unable to call `detached_flows` on `detached_objects.flow_objects` as this returns an array.
I agree the naming is confusing, I can rename the method to `ordered_detached_objects`?

brenetic (approved):
🚀

PR #1269, opened 2022-02-22
Update hpa and lifecycle

brenetic:
- Put prestop lifcycle back
- Update HPA config so it stops accidentally checking for a nil property


mattempty (approved):
🚀

PR #1270, opened 2022-02-22
Remove deployment to steps to the old cluster

brenetic:
We no longer need the deployment steps to the old cluster


mattempty (approved):
🚀

PR #1271, opened 2022-02-23
Don't show 'Add branching' item in connection menu after start page

chrispymm:
Simple PR to not show 'Add branching' in the connection menu immediately after the start page.

Before:
image

After:
image


brenetic (approved):
🚀

PR #1251, opened 2022-02-18
Conditionally show hide answers on branch conditions

chrispymm:
Resolves [ticket #2252](https://trello.com/c/xFN9c4L4/2252-condition-using-a-checkbox-or-radio-button-operator-is-isnt-answered-shouldnt-have-the-answers-displayed-39-7-s)

When changing a condition to be `is_answered` or `is_not_answered` the question answers will be hidden, and the answer value set to empty.

When changing a branch back from `is_answered` or `is_not_answered` to `is` or `is_not` the answers will be revealed and the first option auto-selected.

Backend fixes ahve been applied to allow the answer value to be blank.  This also resolves [#2298](https://trello.com/c/h7jC72fU/2298-validation-of-condition-using-a-checkbox-or-radio-button-operator-is-isnt-answereds)

Before:
image

After
image

image

PR #1278, opened 2022-02-24
Fix bug resetting answer field value

chrispymm:
Resolves [ticket 2367](https://trello.com/c/Us1MyGoN/2367-branching-state-conflict-logic-defaulting-to-first-value-added-xs) 

Fixes bug introduced in [PR 1251](https://github.com/ministryofjustice/fb-editor/pull/1251) that would reset the value of the `field` dropdown on branch edit save.

Added the required conditional to only reset the value on the dropdown if there is no value (i.e. after we have switched from the `is answered` or `is not answered` operator back to an `is` or `is not` operator)

Added in acceptance tests to validate the behaviour.

### DIFF
--- a/app/javascript/src/component_branch.js
+++ b/app/javascript/src/component_branch.js
@@ -330,7 +330,9 @@ class BranchAnswer {
       $answer.val([]);
     } else {
       $answer.show();
-      $answer.val( $answer.find('option:first').val() );
+      if( !$answer.val() ) {
+        $answer.val( $answer.find('option:first').val() );
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


brenetic (approved):
🚀

app/javascript/src/component_branch.js
```diff
@@ -330,7 +330,9 @@ class BranchAnswer {
       $answer.val([]);
     } else {
       $answer.show();
-      $answer.val( $answer.find('option:first').val() );
+      if( !$answer.val() ) {
```
stevenburnell-moj, line 333: Wondering if it would have been easier to do and understand intention if something like...
```
    if(hideAnswers) {
      $answer.hide();
      $answer.attr("disabled", true);
    }
    else {
      $answer.attr("disabled", false);
      $answer.show();
    }
```
...but not enough for me to want to hold up these changes.

stevenburnell-moj (approved):

PR #1281, opened 2022-02-25
Remove docs directory

brenetic:
This has never been deployed out to github pages properly.

It is also not really up to date and has some vulnerabilities associated
with it


njseeto (approved):
🚀

PR #1283, opened 2022-02-25
Disable question select until API request returns

chrispymm:
Fixes [ticket #1984](https://trello.com/c/TWXgseDr/1984-bug-operator-and-answer-list-bug)

On changing the value of the condition select, disable the select field until the action is complete.

In the case of selecting an unsupported answer, this is basically instant. However, in the case of selecting a valid question for branching the select will be disabled until the API request is complete and the UI has been updated with the operator and answer select elements.

This prevents the user changing the condition select while the API request is in-flight which resulted in them potentially being presented with the 'question type unsupported' error and then the API request completing and updating the UI with the operator and answer fields.


brenetic (approved):
🚀

PR #1285, opened 2022-02-25
Move connection menu to the right of the branch condition

chrispymm:
Resolves [ticket #2359](https://trello.com/c/iLTHKxCv/2359-move-connection-menu-to-the-right-of-the-branch-condition) 

This PR simply moves the connection menu activator to the right of the branch condition text on the flow view.

It also slightly adjusts where/when the menu is rendered to the screen to help with source ordering/stacking contexts in order to play nice when merged with other work done on `main`.

Before:
image

After: 
image


brenetic (approved):
🚀

PR #1288, opened 2022-02-25
Increase parallelism in acceptance tests

brenetic:
Use 7 concurrent tests containers. With the horizontal scaling
configuration set for the editor and the metadata api it should be able
to cater for it.

However we need to lower the number of containers for the testable
editors as they tend to overwhelm the pods. This will increase the time
for the testable editor tests unfortunately. We can add auto scaling to
them at some point in the future.


chrispymm (approved):
🚀

PR #1290, opened 2022-02-28
Prevent action when clicking 'Question' in Add Component menu on multiquestion page

chrispymm:
Resolves [ticket #2378](https://trello.com/c/8XRdt4z8/2378-add-question-shouldnt-dismiss-the-menu-71-xs)

Apply the same fix as used on the connection menu to prevent an activated menu item with a submenu that appears on hover from triggering an action and closing the menu on click.


brenetic (approved):
🚀

PR #1294, opened 2022-03-01
Fix unwanted HTML comment in editable content

stevenburnell-moj:
Fix unwanted HTML comment being injected by Showdown's markdown converter code.

**Was:**
Screenshot 2022-03-01 at 21 01 17

**Now:**
Screenshot 2022-03-01 at 21 04 13

### DIFF
--- a/app/javascript/src/editable_components.js
+++ b/app/javascript/src/editable_components.js
@@ -797,8 +797,9 @@ class EditableCollectionItemRemover {
  * Includes clean up of HTML by stripping attributes and unwanted trailing spaces.
  **/
 function convertToMarkdown(html) {
-  html = sanitiseHtml(html);
-  return converter.makeMarkdown(html);
+  var cleaned = sanitiseHtml(html);
+  var markdown = converter.makeMarkdown(cleaned);
+  return sanitiseMarkdown(markdown);
 }
 
 /* Extremely simple function to safely convert target elements, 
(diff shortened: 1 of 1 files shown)


chrispymm:
The fix you've added works and does what it needs to do, but does feel a little 'hacky' and that that it should really be solvable by the markdown converter. So I had a little dig.

Found [this issue](https://github.com/showdownjs/showdown/issues/700) on the showdown issues log, albeit from nearly 3 years ago which seems to perfectly describe the problem here.  A list which when converted results in an extraneous html comment.  The fix mentioned there is to install a slightly newer (then alpha) version of the package (2.0.0-alpha1).

We appear to have the version pegged at 2.0.0, but looking into the code we still have the version with this code in the file `showdown/src/subParsers/makeMarkdown/list.js`

```
// add comment at the end to prevent consecutive lists to be parsed as one
  txt += '\n<!-- -->\n';
  return txt.trim();
```

Unfortunately this is where it gets annoying. When I started writing this, I thought that the latest version of the package didn't have these lines of code 🎉.  I then realised I was looking at the `master` branch for the package in GitHub and sadly, as far as I can see all the tagged / release branches of this file contain the offending code 😞. So it looks like a package update won't solve it.  

So after all that.  It seems like the workaround in this PR is the only way around it.  

(Maybe you already knew all of the above already)

So after all that LGTM!

chrispymm (approved):
✅

PR #1297, opened 2022-03-02
Increase capybara default wait time

brenetic:
Double the wait time to 4 seconds to see if it improves test performance


chrispymm (approved):
🚀

PR #1303, opened 2022-03-03
Fix radios and checkboxes components deletion

chrispymm:
Fix for [ticket #2375](https://trello.com/c/0T5Rt3u1/2375-bug-cant-delete-extra-radio-and-checkboxes-options-69-s)

Reinstates the code deleted previously to trigger confirm modal on attempting to delete an option from radio or checkbox components.

Also add in tests for option addition and deletion

**N.B.**
I am aware that the test refactoring here is not very DRY (at all!)

However when I attempted to dry it up and extract shared methods it uncovered a bit of a pandoras box - turns out there are at least 4 methods `when_I_update_the_question_name` some defined with an argument and others not, and it became hard to know which one was being called...  So in a bid not to refactor too much in a tiny bugfix I left this at the simplest implementation I could get away with to test the addition and deletion of the options.

### DIFF
--- a/acceptance/features/edit_single_checkboxes_question_page_spec.rb
+++ b/acceptance/features/edit_single_checkboxes_question_page_spec.rb
@@ -1,0 +1,167 @@
+
+require_relative '../spec_helper'
+
+feature 'Edit single radios question page' do
+  let(:editor) { EditorApp.new }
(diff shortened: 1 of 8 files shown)


chrispymm:
closing in favour of merging a non-testable branch

PR #1304, opened 2022-03-03
Fix radio and checkbox component option deletion

chrispymm:
Fix for [ticket #2375](https://trello.com/c/0T5Rt3u1/2375-bug-cant-delete-extra-radio-and-checkboxes-options-69-s)

Reinstates the code deleted previously to trigger confirm modal on attempting to delete an option from radio or checkbox components.

Also add in tests for option addition and deletion

**N.B.**
I am aware that the test refactoring here is not very DRY (at all!)

However when I attempted to dry it up and extract shared methods it uncovered a bit of a pandoras box - turns out there are at least 4 methods `when_I_update_the_question_name` some defined with an argument and others not, and it became hard to know which one was being called...  So in a bid not to refactor too much in a tiny bugfix I left this at the simplest implementation I could get away with to test the addition and deletion of the options.

### DIFF
--- a/acceptance/features/edit_single_radios_question_page_spec.rb
+++ b/acceptance/features/edit_single_radios_question_page_spec.rb
@@ -1,0 +1,161 @@
+require_relative '../spec_helper'
+
+feature 'Edit single radios question page' do
+  let(:editor) { EditorApp.new }
+  let(:service_name) { generate_service_name }
+  let(:page_url) { 'star-wars-question' }
+  let(:question) do
+    'Which program do Jedi use to open PDF files?'
+  end
+  let(:initial_options) do
+    ['Adobe-wan Kenobi', 'PDFinn']
+  end
+  let(:options_after_addition) do
+    ['Adobe-wan Kenobi', 'PDFinn', 'Jar Jar Binks']
+  end
+  let(:options_after_deletion) do
+    ['Adobe-wan Kenobi', 'PDFinn']
+  end
+  let(:section_heading) { 'I open at the close' }
+  let(:default_section_heading) { I18n.t('default_text.section_heading') }
+
+  background do
+    given_I_am_logged_in
+    given_I_have_a_service
+  end
+
+  scenario 'when editing the radio component' do
+    given_I_have_a_single_question_page_with_radio
+    and_I_have_optional_section_heading_text
+    when_I_update_the_question_name
+    and_I_update_the_options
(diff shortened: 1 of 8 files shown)


acceptance/features/edit_single_radios_question_page_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+
+  def and_I_edit_the_options
+    editor.editable_options.first.set(initial_options.first)
+    editor.editable_options.last.set(initial_options.last)
+  end
+
+  def and_I_add_an_option(option_label)
+    click_button(I18n.t('actions.option_add'))
+    # editor.editable_options.last.find('label').base.send_keys(option_label)
```
brenetic, line 101: can this be removed?

acceptance/support/common_steps.rb
```diff
@@ -336,9 +337,10 @@ def when_I_want_to_select_component_properties(attribute, text)
   end
 
   def and_I_click_on_the_three_dots
-    sleep 1 # Arbitrary delay, possibly required due to focus issues
+    # sleep 1 # Arbitrary delay, possibly required due to focus issues
```
brenetic, line 340: can we remove this if we're not gonna sleep?

acceptance/support/common_steps.rb
```diff
@@ -353,13 +355,16 @@ def then_I_should_only_see_three_options_on_page_menu
...
   def then_I_should_not_be_able_to_add_page(page)
+    find('#main-content', visible: true)
     editor.preview_page_images.first.hover
     editor.three_dots_button.click
     editor.add_page_here_link.click
     expect(editor.text).not_to include(page)
   end
 
   def then_I_should_be_able_to_add_page(page)
+    # expect(page).to have_selector('#main-content', visible: true)
```
brenetic, line 366: is this check needed?

acceptance/features/edit_single_radios_question_page_spec.rb
chrispymm: Fixed

acceptance/support/common_steps.rb
chrispymm: Fixed

acceptance/support/common_steps.rb
chrispymm: Done

brenetic (approved):
🚀

PR #1308, opened 2022-03-04
Parameterize url before validation

njseeto:
[Trello](https://trello.com/c/8nAk5oTE/2373-auto-generate-url-from-user-input)

Attempted to do this previously with a `before_validation` callback in the PageCreation class, however, the validation continues to be triggered.

Making this change in the controller means we do not trigger the special characters validation in the PageCreation class and allows us to preserve all validations in the PageCreation class.


brenetic (approved):
🚀

PR #1309, opened 2022-03-06
Fix footer height when too small to reach base of page

stevenburnell-moj:
Due to changes required for the latest scrolling, the standard footer background (normally a grey background on <html> element) cannot hide the white gap between the base of footer and bottom of page. This is a correction to the calculation previously trying (and failing) to stretch the footer height to compensate).

**Was**
Screenshot 2022-03-06 at 15 05 15

**Now**
Screenshot 2022-03-06 at 15 05 33

### DIFF
--- a/app/javascript/src/controller_services.js
+++ b/app/javascript/src/controller_services.js
@@ -58,7 +58,7 @@ ServicesController.edit = function() {
 
   createPageAdditionDialog(view);
   createPageMenus(view);
-
+  createConnectionMenus(view);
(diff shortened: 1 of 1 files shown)


chrispymm (approved):
🚀 Looks like it should do the trick, although we could probably achieve this without JS. 

If we made the body element a flex container with flex direction of column and a min-height of 100%, then the `<header>` and `<nav>` elements can be fixed heights and the `<main>` and  `<footer>` elements would stretch as needed to fill the space.

PR #1310, opened 2022-03-06
Testable 2331 branch edit update

stevenburnell-moj:
Move some elements around and a fair number of general pixel here and there tweaks to spacing.

**Was**
Screenshot 2022-03-06 at 21 15 00

**Now**
Screenshot 2022-03-06 at 21 15 32


chrispymm (approved):
🚀

PR #1316, opened 2022-03-07
Fix incorrect calculation causing unconnected path

stevenburnell-moj:
Not sure why this calculation was now showing an unconnected line.
Bit worried that by fixing this something else gets broken but, in theory at least, that should not be.
Fixed it by simplifying calculation and basing it on similar one used for ForwardUpForwardDownPath.

**Was**
Screenshot 2022-03-07 at 14 50 01

**Now**
Screenshot 2022-03-07 at 14 50 39


brenetic (approved):
🚀

PR #1323, opened 2022-03-09
Fixes 2331 branch edit update 2

stevenburnell-moj:
Changes to Branch edit page that centre around different (very tricky, highly frustrating to develop) errors design within the same component. 

See the Figma design for full (and different) states/views but, essentially, this PR is to achieve the view in this screenshot.
Screenshot 2022-03-09 at 00 26 51


chrispymm (approved):
All looks good to me

PR #1330, opened 2022-03-09
Fixes 2331 branch edit update 3

stevenburnell-moj:
Not the best effort as we are rushing but more changes to try and scoop up last-minute issues noticed with previous CSS tweaks to the Branch edit page.

Some error output
Screenshot 2022-03-09 at 11 31 21

And less error output
Screenshot 2022-03-09 at 11 31 35


brenetic (approved):
🚀

PR #1333, opened 2022-03-09
Allow for times the confirmation page does not exist

brenetic:
Sometimes the confirmation page will not exist in a service. Therefore
do not allow the app to error when checking for the UUID of a page that
may or not be there.


njseeto (approved):
🚀

PR #1340, opened 2022-03-11
Fix indexing of component, operator and field

njseeto:
[Trello](https://trello.com/c/Ajo83X4k/2409-unspecified-error-when-triing-to-have-an-and-condition-with-two-options-of-the-same-checkbox-as-contain)
When a user clicks 'Add condition' on a branching point, the name and id attributes should change.
Currently, only the `component` indices are being correctly indexed, the `operator` and `field` indices were not changed, as a result an error was thrown. This PR fixes this issue.

### Before
old

### After
Screenshot 2022-03-11 at 12 02 23

### DIFF
--- a/acceptance/features/edit_branch_page_spec.rb
+++ b/acceptance/features/edit_branch_page_spec.rb
@@ -44,7 +44,47 @@ feature 'New branch page' do
     when_I_save_my_changes
     then_I_should_be_on_the_correct_branch_page('edit')
 
-    then_I_can_add_and_delete_conditionals_and_expressions
+    then_I_can_add_conditionals_and_expressions
+
+    and_I_choose_an_option(
+      'branch[conditionals_attributes][0][expressions_attributes][2][component]',
+      'What is your favourite hobby?'
+    )
+
+    and_I_choose_an_option(
+      'branch[conditionals_attributes][0][expressions_attributes][2][operator]',
+      'is not'
+    )
+
+    and_I_choose_an_option(
+      'branch[conditionals_attributes][0][expressions_attributes][2][field]',
+      'Sewing'
+    )
+
+    when_I_save_my_changes
+    then_I_should_not_see_an_error_summary
+
+    then_I_should_see_the_field_option(
+      'branch[conditionals_attributes][0][expressions_attributes][0][operator]',
+      'is'
+    )
+
+    then_I_should_see_the_field_option(
+      'branch[conditionals_attributes][0][expressions_attributes][0][field]',
+      'Hiking'
+    )
+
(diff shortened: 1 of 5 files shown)


spec/models/expression_spec.rb
```diff
@@ -158,34 +158,34 @@
...
-    it 'constucts the correct name attribute' do
-      expect(
-        expression.name_attr(
-          conditional_index: 1,
-          expression_index: 0,
-          attribute: 'answer'
-        )
-      ).to eq(expected_name)
+    it 'constructs the correct name attribute' do
+      attributes.map do |attribute|
```
brenetic, line 164: `map` returns an array object, but it's never being accessed. the `expect` sits inside the block in which case an `each` would be find if you want to check each attribute expectation

spec/models/expression_spec.rb
```diff
@@ -158,34 +158,34 @@
...
-      ).to eq(expected_id)
+      attributes.map do |attribute|
+        expect(
+          expression.id_attr(
+            conditional_index: 0,
+            expression_index: 1,
+            attribute: attribute
+          )
+        ).to eq("branch_conditionals_attributes_0_expressions_attributes_1_#{attribute}")
+      end
```
brenetic, line 188: i kinda feel like both these test updates aren't necessary really. the controller.js changes don't interact with the name_attr or id_attr methods on the expression object.

if we dont' have JS tests to update then the best thing would be acceptance tests that checks something like:

- no errors are thrown
- the operator and answer chosen remain consistent across more than one conditional

spec/models/expression_spec.rb
njseeto: Done!

spec/models/expression_spec.rb
njseeto: Acceptance tests added!

acceptance/features/edit_branch_page_spec.rb
```diff
@@ -44,7 +44,16 @@
...
+    then_I_can_add_conditionals_and_expressions
+
+    and_I_choose_an_option(
+      'branch[conditionals_attributes][0][expressions_attributes][0][component]',
+      'What is your favourite hobby?'
+    )
+    when_I_save_my_changes
+    then_I_should_not_see_an_error_summary
+
+    then_I_can_delete_conditionals_and_expressions
```
brenetic, line 56: just before we delete them please can we check that 'Which flavours of ice cream have you eaten?' and 'What is your favourite hobby?' are correctly populated on the first and second conditionals respectively

brenetic (approved):

PR #1346, opened 2022-03-14
Prevent deletion of options that branching depends on

chrispymm:


app/models/destroy_question_option_modal.rb
```diff
@@ -0,0 +1,18 @@
...
+  def to_partial_path
+    return 'api/question_options/cannot_delete_modal' if can_not_be_deleted?
+
+    'api/question_options/destroy_message_modal'
+  end
+
+  private
+
+  def can_not_be_deleted?
+    option && expressions.map(&:field).include?(option.uuid)
```
brenetic, line 16:
```suggestion
    option.present? && expressions.map(&:field).include?(option.uuid)
```

minor

app/views/partials/_template_editable_collection_item_menu.html.erb
```diff
@@ -0,0 +1,10 @@
+<script type="text/html"
+        data-component-template="EditableCollectionItemMenu"
+        data-activator-text="<%= t('question.menu.activator') %>">
+  <ul class="govuk-navigation component-activated-menu">
+    <li data-action="remove"
+        data-api-path="<%= URI.unescape( api_service_page_question_question_option_destroy_message_path(service.service_id,@page.uuid,'#{question_uuid}', '#{option_uuid}', method: :delete) ) -%>">
```
brenetic, line 6: i don't think we need to unescape this? rails `_path` methods should output a safe to use URL

app/javascript/src/editable_components.js
```diff
@@ -761,38 +766,24 @@ class EditableCollectionItemInjector {
...
-      editableCollectionItem.remove();
-    });
-
-    // Close on ENTER || SPACE
-    $node.on("keydown.EditableCollectionItemRemover", function(e) {
-      e.preventDefault();
-      if(e.which == 13 || e.which == 32) {
-        editableCollectionItem.remove();
+function createEditableCollectionItemMenu(item, config) { 
+  var template = $("[data-component-template=EditableCollectionItemMenu]");
```
stevenburnell-moj, line 770: Inclined to think this this kind of of thing would be nicer passed in from a view controller through configs, but it's still better than the direct HTML inside JS stuff that has now been removed elsewhere in this file.

app/models/destroy_question_option_modal.rb
chrispymm: How odd - I added this change during our call.   Apparently I failed to save it

app/javascript/src/editable_components.js
```diff
@@ -761,38 +766,24 @@ class EditableCollectionItemInjector {
...
       }
     });
 
-    editableCollectionItem.$node.append($node);
-
-    this.component = editableCollectionFieldComponent;
-    this.item = editableCollectionItem;
-    this.$node = $node;
-  }
+    item.$node.data("ActivatedMenu", menu);
```
stevenburnell-moj, line 784: Think I saw something similar to this line being removed somewhere in the old code, so guessing it's just based on what was found. 

Could do with a tweak if staying in this form to be more clear.
Assuming I'm understanding it correctly, the line itself should be updated to better illustrate the correct type of object being stored (guessing it was previously an ActivatedMenu).
`item.$node.data("EditableCollectionItemMenu", menu);`

However, there's another opportunity to improve on old code here, since it should be following some patterns we apply consistently elsewhere. 

Instead of this:
`item.$node.data("EditableCollectionItemMenu", menu);`

it should be something like this:
`item.menu = menu`

That's less jQuery-ish and more OOP. 

In fact, going with that approach, perhaps we could just return the `menu` instance at end of function so that the calling constructor could have something like this line:
`this.menu = createEditableCollectionItemMenu(this, config);`

app/models/destroy_question_option_modal.rb
chrispymm: Ah no, I see - same change, different file.

app/views/partials/_template_editable_collection_item_menu.html.erb
chrispymm: Yeah, I wasn't 100% sure about this - should have mentioned it on our call. The unescape is here because I actually don't want a safe to use url at this point. Rails escapes the placeholders `#{quesiton_uuid}` so without this the string that gets to JS includes `%23%7Bquestion_uuid%7D%0A`

I guess I could remove the `URI.unescape` and unencode the url in the js instead.... I just figured that as the string in the data-attribute doesn't need to be a valid url, then its not a problem if it contains unencoded characters. The JS is then responsible for replacing the placeholders and creating a valid url from it.

Happy to change though if you'd prefer?

spec/models/destroy_question_option_modal_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      question: question,
+      option: option,
+      label: 'option label'
+    )
+  end
+  let(:service_metadata) { metadata_fixture(:branching_2) }
+
+  describe '#to_partial_path' do
+    let(:question) { page.components.first }
+    subject(:partial) do
```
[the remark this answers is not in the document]
chrispymm: Cool.  This was lifted straight from the other modal spec.  Should I add this improvement to that spec too?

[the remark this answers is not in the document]
brenetic: oh geez. *sigh* no keep it consistent. it's just a convention not the end of the world 😆

app/javascript/src/component_editable_collection_item_menu.js
```diff
@@ -0,0 +1,92 @@
+
```
stevenburnell-moj, line 1: Assume this needs to be added to the Documentation list and should also have JS test at some point. Maybe just add a ticket to backlog as minimum action. There's already `test/component_activated_menu_test.js` so that can probably be used as springboard for most of what is required.

app/javascript/src/editable_components.js
chrispymm: Yep, I just relocated a line and modified it.  Went with your last suggestion, return the menu form the function and assign it to the variable in the constructor. 👍

app/javascript/src/editable_components.js
```diff
@@ -567,6 +569,9 @@ class EditableCollectionFieldComponent extends EditableComponentBase {
     $lastItem.after($clone);
     EditableCollectionFieldComponent.addItem.call(this, $clone, this.$itemTemplate.data("config"));
     EditableCollectionFieldComponent.updateItems.call(this);
+
```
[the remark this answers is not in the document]
chrispymm: So, I've been looking at this one.  It is pretty confusing!! In trying to resolve it though I think I see why it was done this way / can't quite see a way to improve it.

The public `addItem()` function creates a new collectionItem from the itemTemplate - so it doesn't receive a node or config - it creates the node, and sets up the config and passes them on to the private funciton

The private `addItem(node, config)` function instantiates a new `EditableCollectionItem` instance and appends it to the `items` array.  This is called from the `EditableCollectionFieldComponent.createEditableCollectionItems` function, which is responsible for taking the 'plain' radio/checkbox inputs from the initial template and upgrading them.

Moreover, the current implementation of the public `addItem()` depends on the `EditableCollectionFieldComponent.createEditableCollectionItems` method to have been run as it requires the items array to already contain at least one EditableCollectionItem to clone.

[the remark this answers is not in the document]
chrispymm: OK. Ignore the above.  Sorted now.

app/javascript/src/question_radios.js
```diff
@@ -83,56 +72,16 @@ class RadiosQuestion extends Question {
...
 
     // If any Collection items are present with ability to be removed, we need
     // to find them and scoop up the Remove buttons to put in menu component.
-    $(".EditableComponentCollectionItem").each(function() {
-      collectionItemControlsInActivatedMenu($(this), {
-       activator_text: config.text.edit,
-        classnames: "editableCollectionItemControls"
-      });
-    }).parent().attr("aria-label", config.text.aria.answers);
+    $(".EditableComponentCollectionItem").parent().attr("aria-label", config.text.aria.answers);
```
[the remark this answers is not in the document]
chrispymm: Good spot!

app/javascript/src/component_editable_collection_item_menu.js
chrispymm: Added a ticket to the JStests backlog

brenetic (approved):
🚀

PR #1366, opened 2022-03-21
Refactor editable components

chrispymm:
An attempt at making a start on refactoring JS to be more organised.

This PR breaks up the `editable_components.js` file into indivdual classes / function files.

It also switches to using native ES6 import / export (where possible). The logic behind this is to move towards having code that is more directly supported by browsers. while we are never going to be abel to avoid transpiling / bundling due to backwards compatibility we could move towards being able to provide a module nomodule pattern to modern browsers that support it. This would allow for delivering smaller JS file(s) to clients.


app/javascript/src/editable/editable_utilities.js
```diff
@@ -0,0 +1,121 @@
+
```
chrispymm, line 1: Should this file exist?

Most of these methods could exist within the `EditableContent` class?  But some of them are used within `content.js` too.

Or, these functions could live in the global `utilities.js` file?

stevenburnell-moj (dismissed):

PR #1378, opened 2022-03-24
Fix incorrecr branching diamond hover area

chrispymm:
This PR resolves [ticket 2430](https://trello.com/c/0H5lcD7M/2430-clicking-a-branching-diamond-goes-to-the-wrong-branching-edit-screen)

Use the flow item UUID instead of the @pages_flow index as the map name attribute.

Using the @pages_flow index meant that all branching diamonds within the same flow route woudl recieve duplicate names which resulted in hovers over later branching diamonds activating the link of the first branching point in the flow.

Using the UUID results in unique IDs for every branching point.


brenetic (approved):
🚀

PR #1405, opened 2022-04-06
Inject previous UUID into page flow

brenetic:
In order to have the necessary information for the move functionality we
need to inject the previous UUID into the page flow. If the page in
question is also a branch destination then we also inject the
conditional UUID.


mattempty (dismissed):
🚀

mattempty (approved):
🚀

PR #1413, opened 2022-04-08
Add meta viewport tag to correct responsive view

chrispymm:
This adds the correct `viewport` meta tag from the design system so that the interface displays correctly on devices.

image

Before:
image

After:
image

There is a corresponding fix for the presenter, which will then require a version bump here too.  this will enable forms to display correctly in the runner too.


brenetic (approved):
🚀

PR #1416, opened 2022-04-08
Move flow object API endpoints

brenetic:
Add the GET endpoint for the moving object functionality.
Currently the Move model will return a list of valid page and branch targets
that a flow object can be moved after, or replace in the case of a
branch target.

The POST endpoint is not used at this moment.

### DIFF
--- a/app/views/api/move/new.html.erb
+++ b/app/views/api/move/new.html.erb
@@ -1,0 +1,20 @@
+<div class="component-dialog-api-request">
+  <%= form_for @move, url: api_service_flow_move_path(service.service_id, @move.to_move_uuid) do |f| %>
+    <div class="govuk-form-group ">
+      <%= f.hidden_field :previous_flow_uuid, value: @move.previous_flow_uuid %>
+      <%= f.hidden_field :branch_uuid, value: '' %>
+      <%= f.hidden_field :conditional_uuid, value: '' %>
+      <%= f.label :target_uuid, t('dialogs.move.label', title: @move.title), class: "govuk-label govuk-label--m" %>
+      <p> <%= t('dialogs.move.lede') %></p>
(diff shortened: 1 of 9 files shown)


app/views/api/move/new.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <p> <%= t('dialogs.move.lede') %></p>
+      <div class="govuk-form-group">
+        <select class="govuk-select" name="target_uuid" id="target_uuid">
+          <% @move.targets.each do |target| %>
+            <option value="<%= target[:target_uuid] %>" data-branch-uuid="<%= target[:branch_uuid] %>" data-conditional-uuid="<%= target[:conditional_uuid] %>">
+              <%= target[:title] %>
+            </option>
+          <% end %>
+        </select>
+      </div>
```
chrispymm, line 17: Should there be some buttons in here?

chrispymm (changes requested):

app/views/api/move/new.html.erb
brenetic: weeeeeeeeeeeeeeeelll i am not entirely sure. the change destinations modal doesn't seem to have them so i _think_ the frontend end adds them?

chrispymm: Yeah, I know the frontend can add them... but I'm not entirely sure it should need to.  If you know the paths and the actions for the buttons when you're sending the template I don't see why they shouldn't just be there...  Steve might know more...

brenetic: coolio, in that case i shall add them in :)

brenetic: will do this in the next PR which adds the POST endpoint

stevenburnell-moj (approved):

PR #1424, opened 2022-04-12
Surface Move functionality in page menus

brenetic:
## Do not allow stacked branches

Screenshot 2022-04-12 at 14 15 31

If the previous object and the default next of the object being moved
are branches then we should show the user a useful message that prevents
them from moving that object.

## Do not allow moving a page is a branch destination and there is no default next

Screenshot 2022-04-14 at 17 50 10

If the page being moved is a branch destination and has no default next, like exit pages, then we cannot allow these to be moved until the branch has been updated.

## Show Move Page link in page menu

Screenshot 2022-04-12 at 14 43 25

Except for start page, check answers page and confirmation page.

## Preselect the to move objects previous uuid

Preselect the object the previous uuid for the object that is being
moved in the targets list presented in the move modal.

For a branch destination it will select the numbered branch from the
list. e.g Branching Point 1 (Branch 2)

## Differentiate between previous and target conditional UUIDs

When moving a page we need to be able to differentiate between the
previous conditional uuid, if the page being moved was a branch
destination, and the target conditional uuid if the page being moved is
going to replace a branch destination.

### DIFF
--- a/spec/models/move_spec.rb
+++ b/spec/models/move_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe Move do
       end
     end
 
-    context 'when moving a page would not cause stocked branches' do
+    context 'when moving a page would not cause stacked branches' do
       let(:to_move_uuid) { 'cf6dc32f-502c-4215-8c27-1151a45735bb' } # Page B
       let(:previous_flow_uuid) { service.start_page.uuid }
       let(:expected_partial_path) { 'new' }
@@ -148,7 +148,7 @@ RSpec.describe Move do
         expect(previous_flow_target[:selected]).to be_truthy
       end
 
-      it 'sets the selected property to false when ' do
+      it 'sets the selected property to false when previous page has not been matched' do
         other_targets.each do |target|
           expect(target[:selected]).to be_falsey
         end
--- a/spec/models/move_spec.rb
+++ b/spec/models/move_spec.rb
@@ -318,6 +318,19 @@ RSpec.describe Move do
           end
         end
       end
+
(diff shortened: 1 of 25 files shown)


spec/models/move_spec.rb
```diff
@@ -73,6 +136,25 @@
...
+      end
+      let(:other_targets) do
+        targets.reject { |t| t[:target_uuid] == previous_flow_uuid }
+      end
+
+      it 'sets the selected property when previous page matches to true' do
+        expect(previous_flow_target[:selected]).to be_truthy
+      end
+
+      it 'sets the selected property to false when ' do
```
njseeto, line 151: the end of this test case is missing

spec/models/move_spec.rb
```diff
@@ -4,21 +4,84 @@
...
+      let(:to_move_uuid) { 'f475d6fd-0ea4-45d5-985e-e1a7c7a5b992' } # Page J
+      let(:previous_flow_uuid) { '09e91fd9-7a46-4840-adbc-244d545cfef7' } # Branching Point 1
+      let(:expected_partial_path) { 'stacked_branches_not_supported' }
+
+      it 'returns the stacked branches not supported partial' do
+        expect(move.to_partial_path).to eq(expected_partial_path)
+      end
+    end
+
+    context 'when moving a page would not cause stocked branches' do
```
njseeto, line 35:
```suggestion
    context 'when moving a page would not cause stacked branches' do
```

njseeto (commented):
Great stuff! Just a couple of very minor comments 🚀

njseeto (approved):
🚀 💯 🏅

PR #1433, opened 2022-04-14
2472 utilities added functions

stevenburnell-moj:
Refactor existing Utilities tests to make them easier to read.
Add missing tests for newer functions that were not already there.


test/utilities/max_height_test.js
```diff
@@ -0,0 +1,34 @@
...
+
+
+describe('utilities.maxHeight', function() {
+  it('should return the maxHeight found in a collection', function() {
+    var $item1 = $("<p id=\"test-maxheight-item1\">width item 1");
+    var $item2 = $("<p id=\"test-maxheight-item2\">width item 2");
+    var $item3 = $("<p id=\"test-maxheight-item3\">width item 3");
+    var $collection = $(); // empty jQuery collection.
+
+    // Add some widths...
```
chrispymm, line 12: * heights

test/utilities/unique_string_test.js
```diff
@@ -0,0 +1,23 @@
...
+describe('utilities.uniqueString', function() {
+  it('should return a string without passed arguments', function() {
+    expect(utilities.uniqueString().constructor).to.equal(String().constructor);
+    expect(typeof utilities.uniqueString()).to.equal("string");
+  });
+
+  it('should return a string that include the passed string value', function() {
+    var str = 'mystringvalue';
+    expect(utilities.uniqueString(str).constructor).to.equal(String().constructor);
+    expect(typeof utilities.uniqueString(str)).to.equal("string");
```
chrispymm, line 14: The test is supposed to be testing `it('should return a string that include the passed string value')` but there doesn't appear to be an expectation that the returned string contains the required string.

test/utilities/update_dom_by_api_request_test.js
```diff
@@ -0,0 +1,86 @@
...
+        expect($targetNode).to.exist;
+        expect($targetNode.length).to.equal(1);
+        expect($insertedNode).to.exist;
+        expect($insertedNode.length).to.equal(1);
+        expect($targetNode.find($insertedNode).length > 0).to.be.true;
+      }
+    });
+  });
+
+  it('should place new element before target node', function() {
```
chrispymm, line 46: Think this should be `it('should place new element after target node)` as you have type `after` in the `updateDomByApiRequest` call, and you're expectation is `($targettnode).next()`

test/utilities/update_dom_by_api_request_test.js
```diff
@@ -0,0 +1,86 @@
...
+  afterEach(function() {
+    $(document.body).find("#" + insertId).remove();
+  });
+
+  after(function() {
+    $(document.body).find("#" + targetId).remove();
+    $.get = get;
+  });
+
+  it('should place new element after target node', function() {
```
chrispymm, line 27: Should this one be `it('should place new element inside target node')` or maybe `should append new element within target node`?  As the expectation is `expect($targetNode).find($insertedNode)` so we're looking for elements within `$targetNode`

chrispymm (changes requested):

test/utilities/max_height_test.js
stevenburnell-moj: cut-and-paste error now corrected

test/utilities/unique_string_test.js
stevenburnell-moj: Corrected test to check for included value.

test/utilities/update_dom_by_api_request_test.js
stevenburnell-moj: Correct. Have fixed typo.

test/utilities/update_dom_by_api_request_test.js
stevenburnell-moj: Yes, it should be inside/'append to' the target node as it is testing the default, single param, action. Have now corrected.

chrispymm (approved):

PR #1434, opened 2022-04-14
Generate acceptance tests from fixture

brenetic:
This adds an admin endpoint accessible only by the acceptance tests user
which will take a fixture and directly create a service version via the
API instead of stepping through the creation process in the UI.


stevenburnell-moj (approved):

njseeto (approved):
LGTM! 🚀

PR #1452, opened 2022-04-20
Deny access to forms for unauthorised users

brenetic:
Currently the Editor has two types of users:

- Form creators
- Admin users

These are the only users that should be able to access forms, pages,
branches and to be able to preview forms.

In addition admin users are the only ones that should have access to the
admin dashboard, but should retain access to the other parts of the
Editor no matter who created the form.

We skip checking for specific user authorisation on the all forms page
as there are no specific services involved.

The API endpoints need a bit more work. Some of the API Controllers
inherit from the main apps BranchesController as it is used to setup
some branch parameters. This does mean that the authorisation behaviour
is a little different. For the main app we want to show an unauthorised
page. However for the API endpoints we would want to respond with a
standard unauthorised status.

For this particular PR the endpoints are already requiring that a user
is logged in. Only the Move endpoint affects any metadata and this will
we addressed in another PR.

For the remainder of the API endpoints we will revisit the authorisation
in a different PR as it will be a fairly large restructuring.

Screenshot 2022-04-21 at 10 48 53

### DIFF
--- a/acceptance/features/admin_spec.rb
+++ b/acceptance/features/admin_spec.rb
@@ -25,7 +25,7 @@ feature 'Visiting admin pages' do
     then_I_should_not_see_the_admin_link
     admin_pages.each do |path|
       given_I_visit_an_admin_page(path)
-      then_I_should_be_redirected_to_the_services_page
+      then_I_should_be_redirected_to_the_unauthorised_page
     end
   end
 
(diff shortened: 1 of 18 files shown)


njseeto (approved):
🚀  LGTM!
Shared examples in the tests are 👌

PR #1456, opened 2022-04-21
Fix activated menu voiceover labels

chrispymm:
This is an update / improvement on the previous version of this fix.

The prior attempt at this fix went for the 'easy' option of simply replacing the '+' icon on the connection menu with a UTF-8 symbol that didn't get read out by a screenreader.  However the result was tbh a bit ugly.

So this PR does it properly!

We switch from using a `::after` with generated content, to using a `span` with `aria-hidden="true"` along with wrapping the main label text in a span that is visually hidden.

This is a more robust solution (i.e. it is more guaranteed to work across other screenreading software than just MacOS voiceover) and it also means we can continue with the same visual appearance as before.

Previous Fix:
image

New Approach:
image

### DIFF
--- a/app/javascript/src/components/menus/activated_menu_activator.js
+++ b/app/javascript/src/components/menus/activated_menu_activator.js
@@ -35,9 +35,13 @@ class ActivatedMenuActivator {
     this.#config = config;
     this.#className = "ActivatedMenuActivator";
     this.menu = menu;
+    
+    if(!this.#config.hasOwnProperty('activator_icon')) {
+      this.#config['activator_icon'] = '...';//'...';
+    }
+    
     this.$node = this.#insertNode($node);
     this.$node.data("instance", this);
-    
     this.#addAttributes();
     this.#bindEventHandlers();
   }
@@ -64,10 +68,12 @@ class ActivatedMenuActivator {
   #createNode() {
       const $node = $(
         createElement("button", 
-                      this.#config.activator_text, 
+                      undefined, 
                       this.#config.activator_classname
         )
       );
+      $node.append('<span class="ActivatedMenuActivator__icon" aria-hidden="true">'+this.#config.activator_icon+'</span>');
+      $node.append('<span class="govuk-visually-hidden">'+this.#config.activator_text+'</span>');
       return $node;
   }
 
--- a/app/javascript/src/components/menus/connection_menu.js
+++ b/app/javascript/src/components/menus/connection_menu.js
@@ -11,7 +11,8 @@ class ConnectionMenu extends ActivatedMenu {
     super($node, mergeObjects({
       activator_classname: $node.data("activator-classname"),
       container_id: $node.data("activated-menu-container-id"),
-      activator_text: $node.data("activator-text")
+      activator_text: $node.data("activator-text"),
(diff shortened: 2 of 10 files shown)


app/javascript/src/components/menus/activated_menu_activator.js
```diff
@@ -35,9 +35,13 @@ class ActivatedMenuActivator {
     this.#config = config;
     this.#className = "ActivatedMenuActivator";
     this.menu = menu;
+
```
stevenburnell-moj, line 38: The three dots are a stylistic addition and part of the design, so should be controlled by the CSS and not become part of the code in the component.

app/javascript/src/components/menus/activated_menu_activator.js
```diff
@@ -64,10 +68,12 @@ class ActivatedMenuActivator {
   #createNode() {
       const $node = $(
         createElement("button",
```
stevenburnell-moj, line 70: Isn't this change counter-productive to Accessibility in that you're replacing (presumably) useful text with nothing?

app/javascript/src/components/menus/activated_menu_activator.js
```diff
@@ -64,10 +68,12 @@ class ActivatedMenuActivator {
   #createNode() {
       const $node = $(
         createElement("button", 
-                      this.#config.activator_text, 
+                      undefined, 
                       this.#config.activator_classname
         )
       );
```
stevenburnell-moj, line 74: I see what you're going for here. but it's not good to put this much HTML in the JS component code, especially when it's aiming to resolve a design-only issues (the icon) or hardcoding GDS (so visual only) classnames. 

Also, as per the notes on Configuration above, you can pass in an activator element to use, so all this could be in the template. When nothing is specified the component uses the above createElement() function to inject a basic <button>

stevenburnell-moj (changes requested):

app/javascript/src/components/menus/activated_menu_activator.js
chrispymm: The accessible text is exactly the same - the activator_textis what will be used as an accessible label and read out by screen readers.

app/javascript/src/components/menus/connection_menu.js
```diff
@@ -11,7 +11,8 @@ class ConnectionMenu extends ActivatedMenu {
     super($node, mergeObjects({
       activator_classname: $node.data("activator-classname"),
       container_id: $node.data("activated-menu-container-id"),
-      activator_text: $node.data("activator-text")
+      activator_text: $node.data("activator-text"),
+      activator_icon: $node.data("activator-icon")
```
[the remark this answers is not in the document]
chrispymm: But the issue here is that it can't be controlled by css - any icon injected by css will be read out by a screen reader, so it is not a 'design only' element...  

Maybe it is simply the naming... I've called it icon which implies a 'design' element, but it can essentially be thought of as a 'visible' label and an `accessible` label.

The only other solution I can see to this would be using SVG... ...which may work and would keep presentational things out of the markup as I think I can use an svg url in the `::after' content attribute.

I'll have a play and see what I can come up with...

PR #1467, opened 2022-04-25
Update acceptance tests to use fixtures

brenetic:
This updates many of the editing of pages and components acceptance
tests to use fixtures instead of constantly recreating services before
each scenario by interacting with the UI.

We still keep some of the tests which do it that way, but it is
unecessary for all of them.


njseeto (approved):
🚀

PR #1480, opened 2022-04-27
Flow view accessibility audit fixes

chrispymm:
This PR addresses a couple of errors on the flow view raised by the WAVE accessibility audit tool.

1. Adds some alt text to `imagemap` and `area` tags.
2. Updates the expander component to use a button for the expander.

1.  The alt text on these elements is slightly redundant, as it is entirely ignored by screenreaders, and to all intenets an purposes its a decorative element, with the main link text doign all the work.  However I've opted to amke the errors go away, as a t some point we might introduce automated accessibility tests into CI so it's easier if they're gone!

2. This change was made as an H2 is an invalid element for the aria-expanded attribute. It must be an interactive element, and should really be a button. This means it will now be announced correctly by screenreaders.  This whole thing could proabbly be replaced with `<details>/<summary>` to use antive behaviour, but it's in there now so I'm leaving it alone!


brenetic (dismissed):
🚀

app/views/partials/_standalone.html.erb
```diff
@@ -1,5 +1,5 @@
 <section id="flow-standalone-pages" data-component="Expander">
-  <h2 class="govuk-heading-s"><%= t('pages.footer') %></h2>
+  <button class="govuk-heading-s"><%= t('pages.footer') %></button>
```
stevenburnell-moj, line 2: This is not very good. Assuming a non-JS environment (yes, I know...) we can easily see the HTML is no longer semantic. H2 provides something that button does not.

We should look for an enhancement technique that neither duplicates the word output (so no H2 + button combo), and does not simply add a link within the H2 (yes, not illegal in HTML and sometimes used by others but really just a horrible cop-out solution).

Thoughts on a solution... (not saying any are correct or suitable): 
- Possibly, we can adjust the Expander element so that it does not completely hide the content (thinking something like we do with access-hidden text, maybe if the container area just had height of 1px rather than 0px, if that's the case). 
- Allowing things to be controlled by CSS so an Accessible user can easily override (e.g. without the Expander class or Expander.open - not sure if a thing - would be it is classed as visible). Someone with a screen reader (presumably the target audience for changes) might switch CSS off by default, for obvious reasons. Also, I think there might be some media query that relates to screen readers. Perhaps that is an option.
- Some kind of aria-something (haspopup, controls, expanded, role=button ??) solution that might help to work round issues. 
- Not closing the area by default so it is alway open for screen readers (TBH, I can't see there being much point in closing anyway).
- Maybe put aria-hidden on the H2 and use the class="hidden" technique for the button. Actually I don't like this option because it's still not semantic!
- Perhaps as per the above but the H2 doesn't have aria-hidden, so works as a heading, and the button is class="hidden' thing that says something suitable about revealing the related content.

Sorry, I'm out of ideas now.

stevenburnell-moj (changes requested):

stevenburnell-moj:
> 🚀

[steven inserts a frowning emoji]

app/views/partials/_standalone.html.erb
chrispymm: You're right. I hadn't thought this through properly. 

Well we're sort of both right. It should have an H2 for semantics, but it should also have the correct interactive elements for accessibility.

I think the proper structure for this is probably 'details/summary` but I'll investigate further as to what we can do.

chrispymm: Have reverted this section of the PR, and will do the changes to the expander in a separate Pr as it's a bit larger in scope now.

app/views/services/_flow_layout.html.erb
```diff
@@ -20,10 +20,10 @@
   <% else %>
 
     <% if item[:type] == 'flow.branch' %>
-      <article class="flow-item flow-branch" aria-label="<%= item[:title] %>" data-fb-id="<%= item[:uuid] %>">
-        <%= image_pack_tag "diamond.svg",  :usemap => "#thumbnail-#{item[:uuid]}" %>
+      <article class="flow-item flow-branch" data-fb-id="<%= item[:uuid] %>">
+        <%= image_pack_tag "diamond.svg",  usemap: "#thumbnail-#{item[:uuid]}", alt: "Branching diamond", role: "presentation" %>
```
stevenburnell-moj, line 24: The alt text does not provide meaningful content, and therefore benefit, to a user. It is acceptable to leave the alt value empty and, since you've added the role=presentation attribute this should (guess) still provide a way for programmatic analysis of a page to determine the empty alt is valid for accessibility.

app/views/services/_flow_layout.html.erb
```diff
@@ -20,10 +20,10 @@
...
         <map name="thumbnail-<%= item[:uuid] %>" aria-hidden="true">
-          <area shape="poly" coords="0,72.5 100,0 200,72.5 100,145" alt="" href="<%= edit_branch_path(service.service_id, item[:uuid]) %>">
+          <area shape="poly" coords="0,72.5 100,0 200,72.5 100,145" alt="Branching diamond" role="presentation" href="<%= edit_branch_path(service.service_id, item[:uuid]) %>" tabindex="-1">
```
stevenburnell-moj, line 26: Since there's aria-hidden=true on this element, I'd have to look up to see if anything inside is even considered relevant/exposed to Accessible users??

I would expect the alt value to be same 'redundant' issue as the one specified above, however. 

My read aloud (guessing) expectation of this area of code (image+map) would be something like:
"blah... Branching diamond... blah... Branching diamond"

If that's what I'd get as a user experience then I'd file it under 'somewhat annoying' :-)

stevenburnell-moj (changes requested):

stevenburnell-moj (approved):

PR #1483, opened 2022-04-28
Fix brief flash of unstyled content issue

stevenburnell-moj:
Prevent this (no JS dialog implemented yet) view from flashing...
Screenshot 2022-04-28 at 14 52 00

...before this one (with JS dialog) is ready.
Screenshot 2022-04-28 at 14 52 12


chrispymm (approved):
👍

PR #1496, opened 2022-05-03
Remove unnecessary form-group class from move template

brenetic:
There were one too many divs with govuk-form-group as a class in the
move template

### DIFF
--- a/app/views/api/move/_new.html.erb
+++ b/app/views/api/move/_new.html.erb
@@ -1,24 +1,23 @@
 <div class="component-dialog-api-request" id="move_targets_list">
   <%= form_for @move, url: api_service_flow_move_path(service.service_id, @move.to_move_uuid) do |f| %>
+    <%= f.hidden_field :previous_flow_uuid, value: @move.previous_flow_uuid %>
+    <%= f.hidden_field :previous_conditional_uuid, value: @move.previous_conditional_uuid %>
+    <%= f.hidden_field :target_conditional_uuid, value: '' %>
+
     <div class="govuk-form-group ">
-      <%= f.hidden_field :previous_flow_uuid, value: @move.previous_flow_uuid %>
-      <%= f.hidden_field :previous_conditional_uuid, value: @move.previous_conditional_uuid %>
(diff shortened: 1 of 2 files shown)


stevenburnell-moj (changes requested):
be good to keep the hidden inputs out of the div (just in case/readability, etc)

Also, am I reading it right that there is a lede paragraph inside the form group?

app/views/api/move/_new.html.erb
```diff
@@ -1,24 +1,23 @@
...
-        <select class="govuk-select" name="move[target_uuid]" id="move_target_uuid">
-          <% @move.targets.each do |target| %>
-            <% if target[:selected] %>
-              <option value="<%= target[:target_uuid] %>" data-conditional-uuid="<%= target[:conditional_uuid] %>" selected="selected">
-            <% else %>
-              <option value="<%= target[:target_uuid] %>" data-conditional-uuid="<%= target[:conditional_uuid] %>">
-            <% end %>
-              <%= target[:title] %>
-            </option>
+      <p><%= t('dialogs.move.content') %></p>
```
chrispymm, line 9: Sorry to jump in here, but in taking a quick look at this I'm not sure it works well for screenreaders - as they will not easily hear the 'Move to after' text in the paragraph.

image


In the above only the text 'Move the page "Question"' is used as the label for the field.  Voiceover for mac reads this as:

*Apply for a Dog license, Move the page Question, popup button*

This is getting beyond the scope of your initial PR here and probably needs fabien/mark to comment.  But I can see three possible options:

* add an id to the `<p>` and add it as an aria-describedby on the form field, but this probably isn't robust, as aria-describedby behaves differently across different screenreaders
* the main header/label should be changed to 'Move the page X to after', and lose the `<p>` tag
* Make the heading an actual H3, and then change the `<p>` to be a `label`, maybe with the content  'move page to after', relying on the user to keep the context of the page they just clicked on as they may not hear the heading.

brenetic:
i think this "simple" PR has moved way out of my wheel house so i am going to close this and run away 🙈

PR #1501, opened 2022-05-04
Progressively enhance branching diamond with clip-path for nicer focu…

chrispymm:
This PR makes the focus state for the branching diamonds much nicer in browsers that support the `clip-path' property.

With clip-path:
image

Without clip-path (fallback):
image

While the picture at [caniuse](https://caniuse.com/?search=clip-path) doesn't look great, that is mostly due to browsers only supporting part of a fairly large spec for clip-path.  In reality every browser that isn't red on caniuse supports what we're using it for (everything except IE)

image

By wrapping all of the new rules in `@supports` we can provide the old `imagemap` functionality to IE and then in browsers that support it we hide the imagemap and rely on `clip-path` to provide the diamond shaped link area and styling an inline svg to control the colors on focus.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,6 +25,18 @@ module ApplicationHelper
     end
   end
 
+  def flow_branch_link(item)
+    link_to edit_branch_path(service.service_id, item[:uuid]), class: 'govuk-link' do
+      concat(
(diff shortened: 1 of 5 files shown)


stevenburnell-moj (approved):
Shame CSS needs so much and makes developers jump through so many hoops to get this kind of thing :-(

End result though...
Nice one, looks good 👍

PR #1502, opened 2022-05-05
2508 component formdialog

stevenburnell-moj:
JS testing for FormDialog

### DIFF
--- a/app/javascript/src/component_dialog_form.js
+++ b/app/javascript/src/component_dialog_form.js
@@ -24,13 +24,7 @@ const mergeObjects = utilities.mergeObjects;
  * @config (Object) Configurable key/value pairs.
  **/
 class FormDialog {
-
-  static STATE_ATTR = "state";
-  static STATE_OPEN = "open";
-  static STATE_CLOSED = "closed";
-
   #config;
-  #state;
 
   constructor($node, config) {
     var dialog = this;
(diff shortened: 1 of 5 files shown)


app/javascript/src/component_dialog_form.js
```diff
@@ -75,21 +76,21 @@ class FormDialog {
...
 
     $node.on("dialogclose", function( event, ui ) {
-      dialog.$container.attr(STATE_ATTR, STATE_CLOSED);
+      dialog.$container.attr(FormDialog.STATE_ATTR, FormDialog.STATE_CLOSED);
     });
 
     $node.on("dialogopen", function( event, ui ) {
-      dialog.$container.attr(STATE_ATTR, STATE_OPEN);
+      dialog.$container.attr(FormDialog.STATE_ATTR, FormDialog.STATE_OPEN);
     });
```
chrispymm, line 87: Should this not set `#state`?

app/javascript/src/component_dialog_form.js
```diff
@@ -104,7 +105,7 @@ class FormDialog {
   }
 
   get state() {
-    return this.$node.attr(STATE_ATTR);;
+    return this.$container.attr(FormDialog.STATE_ATTR);
```
chrispymm, line 108: Is there a reason why you're reading the state from the html element rather than simply returning the value of the class private property? Is it because that's how the underlying jQuery UI dialog does it? 

Maybe because `#state` doesn't appear to get set/changed after the constructor?

If we're not using the `#state` property and just using the attribute, then we should probably remove `#state` to avoid confusion.

app/javascript/src/component_dialog_form.js
```diff
@@ -75,21 +76,21 @@ class FormDialog {
     ];
 
     // Sort out easy state reporting.
-    this.#state = STATE_CLOSED;
+    this.#state = FormDialog.STATE_CLOSED;
 
     $node.on("dialogclose", function( event, ui ) {
-      dialog.$container.attr(STATE_ATTR, STATE_CLOSED);
+      dialog.$container.attr(FormDialog.STATE_ATTR, FormDialog.STATE_CLOSED);
```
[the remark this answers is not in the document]
stevenburnell-moj: Have now changed to data-state.

app/javascript/src/component_dialog_form.js
stevenburnell-moj: #state is left over from my original effort/plan but is no longer used. I have now removed.

app/javascript/src/component_dialog_form.js
stevenburnell-moj: See earlier comments about previous efforts and #state no actually being used. 

Also, at risk of repeating my first reply, I've simply replaced everything with a dialog.isOpen() wrapper to the (discovered) already existing dialog.$node.dialog("isOpen") jQueryUI Dialog provided route. 

Other than simplicity of access, I also replaced state() with isOpen() to make this component more consistent with your recent Expander change.

chrispymm (approved):

PR #1507, opened 2022-05-05
Number minimum component validator

brenetic:
Create component validations GET and POST endpoints.

The base validation checks for a valid validator for a given component as well as the presence of the value.

The minimum validation doesn't do anything special at this point other than check it's a number.

When setting a component validation if the user input is valid it will just pass back a JSON representation of that validation to be written to the underlying page as well as a status code of 202.

If for some reason the configuration is invalid then the same modal will be represented along with the errors.

Use MetadataPresenter 2.16.3


njseeto (approved):
LGTM! 🚀

PR #1511, opened 2022-05-06
Component validation menu

brenetic:
## Create component question menu template

This adds a template for the component question menu. It gets the
supported validations for a given component and then iterates over the
ones that are currently enabled in the Editor to create the links to
configure them.

## Surface component validations from Presenter

The Component model in the presenter now surfaces the supported
validations for a given component.

### DIFF
--- a/app/views/partials/_template_component_question_menu.html.erb
+++ b/app/views/partials/_template_component_question_menu.html.erb
@@ -1,0 +1,21 @@
+<% @page.components.each do |component| %>
+  <script type="text/html"
+          data-component-template="ComponentQuestionMenu_<%= component.type %>"
+          data-component-uuid="<%= component.uuid %>"
+          data-activator-text="<%= t('question.menu.activator') %>">
+    <ul class="govuk-navigation component-activated-menu">
+      <li data-action="required"><span><%= t('question.menu.required')%></span></li>
+      <li data-action="remove"
+          data-api-path="<%= URI.decode_www_form_component(api_service_page_question_destroy_message_path(service.service_id, @page.uuid, component.uuid, method: :delete)) -%>">
+        <span class="destructive"><%= t('question.menu.remove')%></span>
+      </li>
+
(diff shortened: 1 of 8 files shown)


app/views/partials/_template_component_question_menu.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    <li data-action="remove"
+        data-api-path="<%= URI.decode_www_form_component(api_service_page_question_destroy_message_path(service.service_id, @page.uuid, '#{question_uuid}', method: :delete)) -%>">
+      <span class="destructive"><%= t('question.menu.remove')%></span>
+    </li>
+
+    <% @page.components.each do |component| %>
+      <% enabled_validations(component).each do |validation| %>
+        <li data-action="<%= validation %>"
+            data-api-path="<%= URI.decode_www_form_component(api_service_page_component_validations_path(service.service_id, @page.uuid,'#{question_uuid}', validation)) -%>">
+          <span><%= t("question.menu.#{validation}") %></span>
```
brenetic, line 15: @chrispymm to be honest the `component` is available here so we could just do

`api_service_page_component_validations_path(service.service_id, @page.uuid, component.uuid, validation)`

straight away instead of having to do the find and replace in JS

app/views/partials/_template_component_question_menu.html.erb
```diff
@@ -0,0 +1,21 @@
+<% @page.components.each do |component| %>
+  <script type="text/html"
+          data-component-template="ComponentQuestionMenu_<%= component.type %>"
+          data-component-uuid="<%= component.uuid %>"
```
chrispymm, line 4: We probably don't need both of these - if we have the uuid, we don't really need the type.  I think we can probably manage with
```
data-component-template="ComponentQuestionMenu_<%= component.uuid -%>"
```
But I reckon leave it as it is for this PR and I can tweak if needed.

chrispymm (approved):

PR #1526, opened 2022-05-10
2515 component activatedformdialog

stevenburnell-moj:
Updated JS tests for ActivatedFormDialog


stevenburnell-moj:
Investigations into failing JS test that doesn't happen on local has revealed possible (multiple) randomly failing tests, that vary in result, depending on which combination of tests are called. *sigh*

Closing this PR until whatever is causing issues can be fixed.

PR #1527, opened 2022-05-10
Copy changes to Publish authentication modal

njseeto:
[Trello](https://trello.com/c/k5EntNtE/2501-threat-form-published-to-test-is-missing-basic-auth-m)

- Changes to the title of the modals
- Changes to error messages

These are custom validations as we are unable to use the built in validations without having to manipulate
the `${attribute}`. Since the `${attribute}` passed is the label, any changes to the label could potentially break the error messages.

### Before
Screenshot 2022-05-10 at 18 28 13

### After
Screenshot 2022-05-10 at 18 27 39


brenetic (approved):
🚀

PR #1529, opened 2022-05-10
Refactor base component validator to use Rails validates

brenetic:
Rails has some built in validators that we can use instead of a custom
validtor class


njseeto (approved):
🚀

PR #1537, opened 2022-05-16
Reduce dependabot PR noise

brenetic:
Set dependabot back to the default dependency config. This will only do
the direct dependencies of the app.

It raises too many PRs that we cannot get through reviewing.

Also add chris as a reviewer.


njseeto (approved):
🚀

PR #1541, opened 2022-05-17
Create Enabled Validations Presenter

brenetic:
The majority of validations use a single modal when a user is
interacting with the configuration. However for the text and textarea
components their validation modals are grouped into pairs.

There is a separation between the min and max config, however the same
modal is used to set the character ('*_length') and the word ('*_word')
validations. As a result of this it is not possible to have a direct
mapping of validation type when the API is called.

In terms of the list of menu items instead of the text and textarea
components having:

- max_length
- min_length
- max_word
- min_word

they will have:

- max_string_length
- min_string_length

The BaseComponentValidation model will have to be able to assign the correct
component validation object during the API request.

### DIFF
--- a/app/presenters/enabled_validations_presenter.rb
+++ b/app/presenters/enabled_validations_presenter.rb
@@ -1,0 +1,31 @@
+class EnabledValidationsPresenter
+  def initialize(component)
+    @component = component
+  end
+
+  STRING_VALIDATIONS = %w[max_length min_length max_word min_word].freeze
+  STRING_LENGTH_SUFFIX = '_string_length'.freeze
+
+  def enabled_list
+    validations = enabled_validations.map do |validation|
+      if validation.in?(STRING_VALIDATIONS)
+        "#{prefix(validation)}#{STRING_LENGTH_SUFFIX}"
+      else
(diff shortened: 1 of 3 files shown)


app/presenters/enabled_validations_presenter.rb
```diff
@@ -0,0 +1,32 @@
+class EnabledValidationsPresenter
+  def initialize(component)
+    @component = component
+  end
+
+  STRING_VALIDATIONS = %w[max_length min_length max_word min_word].freeze
+  STRING_LENGTH_POSTFIX = '_string_length'.freeze
```
chrispymm, line 7: Maybe its a convention I'm unaware of, but I would have expected this to be `STRING_LENGTH_SUFFIX` as the opposite to a prefix...

Sorry, very petty I know.

chrispymm (approved):

app/presenters/enabled_validations_presenter.rb
brenetic: no problemo :) have updated the PR

njseeto (approved):
Looks good! 🚀

PR #1545, opened 2022-05-18
Adapt expander to handle checkbox revealed inputs

chrispymm:
Makes changes to the expander component to support using a checkbox or radio button as the activator.

Allows the expander component to be used for conditionally-revelaed inputs pattern used in the validation modals.

Doing this required moving the initializer for the flow view expanders from `controller_default` into `controller_services` to enable the different kinds of expander to be initialized without conflict.

### DIFF
--- a/app/javascript/src/component_expander.js
+++ b/app/javascript/src/component_expander.js
@@ -130,6 +130,7 @@ class Expander {
   #createActivator() {
     var source = this.#config.activator_source;
     var $activator;
+    var activatorEventType;
 
     if(typeof source == 'string') {
       $activator = $('<button>' + source + '</button>');
@@ -137,7 +138,7 @@ class Expander {
     }
     else {
       // Assume source to be jQuery element from this point.
-      if(source.is('button')) {
+      if(this.#isValidActivatorElement(source)) {
         $activator = source;
       }
       else {
@@ -160,7 +161,9 @@ class Expander {
       'aria-controls': this.#id
     });
 
-    $activator.on("click", () => {
+    activatorEventType = this.#isInput($activator) ? 'change' : 'click';
+
+    $activator.on(activatorEventType, () => {
       this.toggle();
     });
 
@@ -215,7 +218,14 @@ class Expander {
     return $node;
   }
 
-}
+  #isValidActivatorElement(source) {
+    return source.is('button') || source.is('input[type="checkbox"]') || source.is('input[type="radio"]')
+  }
 
+  #isInput($activator) {
+     return $activator.is('input[type="checkbox"]') || $activator.is('input[type="radio"]')
+  }
+
+}
 
 module.exports = Expander;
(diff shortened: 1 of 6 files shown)


app/javascript/src/controller_pages.js
```diff
@@ -294,32 +295,14 @@ function addQuestionMenuListeners(view) {
...
-              $content.addClass('govuk-checkboxes__conditional--hidden');
-              checkbox.attr('aria-expanded', false);
-            }
+        var $revealedInputs = dialog.$node.find('[data-component="Expander"]');
+        $revealedInputs.each(function() {
+          var $activator = $(this).parent().find('input[type="checkbox"]');
+          new Expander($(this), {
+            activator_source: $activator,
+            wrap_content: false,
           });
```
stevenburnell-moj, line 304: >20 lines down to < 10 👍

app/javascript/src/component_expander.js
```diff
@@ -160,9 +160,16 @@ class Expander {
...
+
+    if(this.#isInput($activator)) {
+      $activator.on("change", () => {
+        this.toggle();
+      });
+    } else {
+      $activator.on("click", () => {
+        this.toggle();
+      });
+    }
```
stevenburnell-moj, line 172: Think `isInput()` is used a couple of times so I'm wondering why we don't set the event type inside the `createActivator()` method, or whatever suitable place. Then we can reduce lines with like: 
```
$activator.on(activator_event_type, () => {
        this.toggle();
      });
```
Might even be able to ditch the curly braces but not 100% sure on syntax.
Readability over streamlined is probably more important than clipping two characters, anyway :-)

app/javascript/src/component_expander.js
```diff
@@ -215,7 +222,22 @@ class Expander {
     return $node;
   }
 
-}
+  #setInitialState() {
+    if(this.#isInput(this.$activator)) {
+      this.$activator.prop('checked') ? this.open() : this.close();
+    } else {
+      this.#config.auto_open ? this.open() : this.close();
+    }
```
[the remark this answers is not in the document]
chrispymm: 👍 good point. All makes sense. 

The state actually gets set within the `open()` and `close()` methods. But if we're going back to a single `auto_open` config value then I can probably drop the `#setInitialState` method.

app/javascript/src/controller_services.js
```diff
@@ -910,5 +912,20 @@ function calculateAndCreatePageFlowConnectorPath(points, config) {
...
+ */
+function createExpanderComponents() {
+  $("[data-component=Expander]").each(function() {
+    var $node = $(this);
+    new Expander($node, {
+      activator_source: $node.find('> h2').first(),
+      wrap_content: true,
+    });
+  });
+}
```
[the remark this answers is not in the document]
chrispymm: I did a grep for expander so I'm pretty sure not. 

With this change we probably **can** now use expander on those modals.

app/javascript/src/component_expander.js
chrispymm: Personal preference maybe, but I can't stand arrow functions without curly braces! 😀

PR #1546, opened 2022-05-18
Fix self-referential page deletion

njseeto:
[Trello](https://trello.com/c/GK7AgK0Y/2577-bug-form-breaks-when-you-delete-a-page-that-points-to-itself)
When a page that has itself as its default next is deleted, the app does not know how to deal with this.
This is because the page preceding the deleted page is trying to find the default next of the deleted page, which does not exist.
This commit replaces the default next of the preceding page with an empty string if the page that has been deleted is self-referential.

Screenshot 2022-05-18 at 15 08 45

At present, deleting the 'Transport' page in the above example would break the app. With this commit the form is able to render the following:

Screenshot 2022-05-18 at 15 39 50

### DIFF
--- a/app/services/page_destroyer.rb
+++ b/app/services/page_destroyer.rb
@@ -37,7 +37,7 @@ class PageDestroyer
 
     @latest_metadata['flow'].each do |_flow_uuid, properties|
       if properties['next']['default'] == uuid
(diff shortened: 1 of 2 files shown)


brenetic (approved):
🚀

PR #1550, opened 2022-05-20
Dialog refactoring

chrispymm:
WIP

A start on refactoring dialogs to be more consistent...


app/javascript/src/component_dialog_validation.js
```diff
@@ -107,15 +97,77 @@ class DialogValidation {
...
+      // Allow a function to be specified in dialog config 
+      utilities.safelyActivateFunction(dialog.#config.onLoad, dialog);
+      dialog.#enhance();
+      if(dialog.#config.autoOpen) {
+        dialog.open();
+      }
+    });
+
+  }
+
```
stevenburnell-moj, line 167: Seems like some optimisation could happen here. 

Not saying this is correct (see below) but something like this, perhaps:
```
#initialise(source) {
  var dialog = this;
  if(typeof source == 'string') {
    dialog.$node = source;
    dialog.#build();
    dialog.#enhance();

    if(dialog.#config.autoOpen) {
      dialog.open();
    }
  }
  else {
    $.get(source, function(response) {
      dialog.$node = $(response);

      // Allow a passed function to run against the created $node (response HTML) before creating a dialog effect
      utilities.safelyActivateFunction(dialog.#config.build, dialog);

    }).done(() => {
      dialog.#build();

      // Allow a function to be specified in dialog config 
      utilities.safelyActivateFunction(dialog.#config.onLoad, dialog);

      dialog.#enhance();

      if(dialog.#config.autoOpen) {
          dialog.open();
      }
    });
  }
}


#build(html) {
    var dialog = this;
    var $node = $(html);

    // Allow a passed function to run against the created $node (response HTML) before creating a dialog effect
    utilities.safelyActivateFunction(dialog.#config.build, dialog);

    // stuff
    $node.on("dialogcreate", function(event, ui) {
        dialog.$container = $node.parents(".ui-dialog");
        dialog.$container.addClass("DialogForm");
    });

    // Add the jQueryUI dialog functionality.
    $node.dialog({
        autoOpen: false,
        closeOnEscape: true,
        height: "auto",
        modal: true,
        resizable: false,
    });

    dialog.$node = $node;
    dialog.$node.data("instance", this);
    dialog.$form = dialog.$node.find('form');
}

```
The auto-open could easily be inside `enhance()` function but name didn't suit so I did not do that in example. Also, the `safelyActivateFunction()` calls look more than needed but haven't investigated what are purpose so left both in this as example code only. Might be one could be ditched/combined into the other.

Ideally, maybe something like this...
```
#initialise(source) {
    var dialog = this;
    if(typeof source == 'string') {
        dialog.$node = source;
        dialog.#build();
    }
    else {
        $.get(source).done((response) => {
            dialog.$node = $(response);
            dialog.#build();
        });
    }
}
```


Just food for thought.

stevenburnell-moj (changes requested):

PR #1553, opened 2022-05-23
Add warning message when there is no service output configuration set

brenetic:
Whenever a service has no output configuration set then we should show a
warning message to let the user know. At this moment there is only
output using email so we check with `SERVICE_EMAIL_OUTPUT` config is
saved for a given environment or not.

Screenshot 2022-05-23 at 13 46 18


njseeto (approved):
🚀

njseeto (approved):
🚀

PR #1557, opened 2022-05-24
Re-publish live radio buttons

njseeto:
[Trello](https://trello.com/c/QruUxdH5/2547-radio-button-on-live-republish-not-displaying-correctly-prioritisation-score-7)
When a form user publishes their form to live without authentication, and then clicks the 'Publish to Live' button again, neither radio buttons are selected.

This is because the `existing_authentication?` method checks whether:
- A service has previously been published or
- If authentication has been previously set

We want the 'Set a username and password' radio to be selected when:
- A user has not previously published a form
- A user has published a form with authentication

We want the 'Allow anyone with the link to view' radio to be selected when:
- A user has published a form without authentication

### Before
Screenshot 2022-05-25 at 10 48 22

### After
Screenshot 2022-05-25 at 10 48 41

### DIFF
--- a/app/views/publish/_form.html.erb
+++ b/app/views/publish/_form.html.erb
@@ -6,7 +6,7 @@
   </h2>
   <div class="govuk-radios">
     <div class="govuk-radios__item">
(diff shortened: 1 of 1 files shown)


brenetic (approved):
🚀

PR #1560, opened 2022-05-25
Remove acceptance test sleeps

brenetic:
We can use cabybara in a way that it is blocking until a page or element
is on the page. This will wait until the default timeout is breached.


njseeto (approved):
🚀

PR #1570, opened 2022-05-30
Export services to CSV from admin dashboard

brenetic:
This creates the ability to export some details about all the services
from the Editor. The details include:

- service name
- email address of user who created the service
- Whether it is published to Test
- Whether it is published to Live

Also bump the version of Administrate gem

Screenshot 2022-05-30 at 16 39 38


njseeto (approved):
🚀

PR #1577, opened 2022-06-07
Add Form Analytics Configuration Settings Page

brenetic:
This adds the ability to add the currently available three types of
Google Analytics products:

- Universal Analytics
- Google Tag Manager
- Google Analytics 4

The same form is used for both Test and Live environments. Those
environment names map to `dev` and `production` in terms of
`deployment_environment` when they are saved as ServiceConfigurations
in the DB.

This is currently only available in the Test environment as the link and page are
behind a feature flag.

Screenshot 2022-06-15 at 13 03 29

Screenshot 2022-06-15 at 13 03 44

Screenshot 2022-06-15 at 13 03 55

Screenshot 2022-06-15 at 13 04 08

Co-authored-by Steven Burnell <steven.burnell@digital.justice.gov.uk>

### DIFF
--- a/app/javascript/src/component_expander.js
+++ b/app/javascript/src/component_expander.js
@@ -62,6 +62,7 @@ const {
  * @param condif (Object) Key/value pairs for config (see above)
 **/
 class Expander {
+  #browser_controlled;
   #config;
   #state;
   #id;
@@ -75,26 +76,48 @@ class Expander {
     }, config);
 
     this.#config = conf;
-    this.#state = 'open';
+    this.#state = 'open'; // This value always get overwritten by openState
     this.#id = uniqueString("Expander_");
+    this.#browser_controlled = false; // Will be changed in createNode if $content is <details> element
     this.$node = this.#createNode($content);
     this.$activator = this.#createActivator();
 
     conf.auto_open ? this.open() : this.close();
   }
 
-  open() {
-    this.$node.slideDown( this.#config.duration, () => {
+  set openState(open) {
+    if(open) {
       this.$activator.attr("aria-expanded", "true");
       this.#state = 'open';
-    });
+    }
+    else {
+      this.$activator.attr("aria-expanded", "false");
+      this.#state = 'closed';
+    }
+  }
+
+  open() {
+    if(this.#browser_controlled) {
+      this.$node.attr("open", true); // <details> native control for open
+      this.openState = true;
+    }
+    else {
+      this.$node.slideDown( this.#config.duration, () => {
+        this.openState = true;
+      });
+    }
   }
 
   close() {
-    this.$node.slideUp( this.#config.duration, () => {
-      this.$activator.attr("aria-expanded", "false");
-      this.#state = 'closed';
-    });
+    if(this.#browser_controlled) {
+      this.$node.attr("open", false); // <details> native control for close
+      this.openState = false;
+    }
+    else {
+      this.$node.slideUp( this.#config.duration, () => {
+        this.openState = false;
+      });
+    }
   }
 
   isOpen() {
@@ -134,6 +157,7 @@ class Expander {
 
     if(typeof source == 'string') {
       $activator = $('<button>' + source + '</button>');
+      $activator.attr("type", "button");
       this.$node.before($activator);
     }
     else {
@@ -142,11 +166,12 @@ class Expander {
         $activator = source;
       }
       else {
-        if(source.children().length < 0) {
+        if(source.children().length < 0 || this.#browser_controlled) {
           $activator = source; // No extra support for Empty element, yet.
         }
         else {
           $activator = $('<button>' + source + '</button>');
+          $activator.attr("type", "button");
           $activator.text(source.text()); // Only want text not child elements.
           source.empty();
           source.append($activator);
@@ -161,11 +186,13 @@ class Expander {
       'aria-controls': this.#id
     });
 
-    activatorEventType = this.#isInput($activator) ? 'change' : 'click';
-
-    $activator.on(activatorEventType, () => {
-      this.toggle();
-    });
+    // If required, add the Expander.toggle() for open/close control.
+    if(!this.#browser_controlled) {
+      activatorEventType = this.#isInput($activator) ? 'change' : 'click';
+      $activator.on(activatorEventType, () => {
+        this.toggle();
+      });
+    }
 
     return $activator;
   }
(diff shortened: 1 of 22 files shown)


app/javascript/src/component_expander.js
```diff
@@ -219,11 +265,11 @@ class Expander {
...
-    return source.is('button') || source.is('input[type="checkbox"]') || source.is('input[type="radio"]')
+    return source.is('button') || source.is('summary') || this.#isInput(source);
   }
 
   #isInput($activator) {
-     return $activator.is('input[type="checkbox"]') || $activator.is('input[type="radio"]')
+   return $activator.is('input[type="checkbox"]') || $activator.is('input[type="radio"]')
   }
 
 }
```
chrispymm, line 275: Initially I was going to ask why we were adding in a bunch of stuff to specifically handle `details/summary` when that does it natively, but I think I understand now:

 - We use details/summary because 'design system'
 - We enhance it with an `Expander` component because we want to be able to control the `details` to expand when we check the checkbox

Correct? 

In which case I think this is OK. The only downside I can see is that there is no indication for screenreader users on the checkbox that it will reveal more content - but I think that's solvable... I'll add it in another comment

app/javascript/src/controller_form_analytics.js
```diff
@@ -0,0 +1,67 @@
...
+      var $checkbox = $("input[type=checkbox]", $this);
+      var expander = new Expander($("details", $this), {
+        auto_open: $(".govuk-form-group--error", $this).length
+      });
+
+      $checkbox.on("click", function() {
+        if(this.checked && !expander.isOpen()) {
+          expander.open();
+        }
+      });
```
chrispymm, line 60: In here you could grab the `id` of the `Expander` component, and add an `aria-controls` attribute to the checkbox input, and also add an `aria-expanded` attribute on checking the box (and remove when unchecking)

```
$checkbox.attr('aria-controls', $("details", $this).attr('id'));
```

Not sure how great it is to have two things controlling the same widget accessibility wise, but it is probably best to let users know that something is going to happen when they check the box.

chrispymm (changes requested):

app/javascript/src/component_expander.js
```diff
@@ -61,44 +61,71 @@ const {
...
+  set openState(open) {
+    if(open) {
       this.$activator.attr("aria-expanded", "true");
-      this.#state = 'open';
-    });
+      this.#state = Expander.STATE_OPEN;
+    }
+    else {
+      this.$activator.attr("aria-expanded", "false");
+      this.#state = 'closed';
```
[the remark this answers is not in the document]
stevenburnell-moj: I added the STATE_OPEN simply because it was a little confusing seeing "open" for that value and "open" for the attribute value in close proximity. Felt too easy to confuse the two and introduce a simple error. That's why I also put in the NATIVE_CONTROL_ATTRIBUTE. We do use it twice, so there is also some gain in using a constant in the traditional sense as well, but that was not a driving factor.

I nearly put in STATE_CLOSED for the consistency like you suggest, but didn't bother because the value of what would be STATE_CLOSED is somewhat irrelevant.

The internal state tracking value is never exposed. We're interested (inside isOpen() method only) to find if state is equal to STATE_OPEN but the "closed" value could just as well be an empty string, or "fred", or "anything else". In fact, if I'd used an empty string it might be more correct and you may not have even queried it.

app/javascript/src/component_expander.js
stevenburnell-moj: You are correct.

Without augmenting the native browser functionality of details/summary to, at least, track the internal openState, I found the Checkbox could get confused as to whether it was open, or not. 

I didn't add any 'aria-controls' (for example) to the Checkbox because it doesn't really control it, at least in the proper sense. If you're using a screenreader, after checking that box to say you want analytics, by the time you reach the details/summary, I am assuming the details/summary will report its state as open (or whatever) and work as though it was already in an open state, etc. The checkbox is not an open/close control, as such, but more a  convenience feature that means they do not need to perform an extra click by the time they read it. If they have not clicked the checkbox and go straight to the details/summary, presumably they would interact with it in whatever normal fashion.

The intended use was to make sure (and I guess it is the same for all users) that, once the checkbox is ticked, users do not miss the (hidden but also required) configuration options.

app/javascript/src/controller_form_analytics.js
stevenburnell-moj: We could but that might be confusing. As mentioned above, the Checkbox is not really a control. Nothing happens when it is unchecked, so removing 'expanded' would not be correct. The actual control, `<summary>`, is presumably doing all that but, I believe, its 'expanded' value does get updated by the Checkbox, so that it is kept in sync.

Actually, I doubt the native functionality of details/summary even needs any aria attribute for this, but it's there because the open method simply sets the state, and the state updates the actual activator, which is not always a `<summary>` tag. I'm taking the position of 'no harm in having both' and 'saves having extra browserControlled checks around that attribute update'.

chrispymm: I'll respond to both comments here:

The TLDR is that I think I agree and that on reflection the best solution* is not to add the extra attributes.

You're right in that the checkbox is not a 'full' control in that it only opens, and doesn't close the expander.  

However, I think your assumption of how a screenreader user navigates the page is possibly slightly off.  My understanding is that they won't naively tab through a document like a sighted user, they will often 'scan' a page before navigating it - to find all headings, or form elements or interactive elements - so if they don't know more content has been revealed (by a screenreader announcing an element controlling an expanding area) then they may not know it is there, and continue with their mental model of the page as they've scanned it.

So in this instance a screenreader user may have discovered the forms with the checkboxes and submit button. So after checking the box, may just hit return to submit the form.

In this instance, as it is a weird case, I think we'll err on not adding the extra aria - but there is the potential of non-discovery of content.

On another note, adding aria attributes to a native details summary is almost certainly an anti-pattern, as the best aria is no aria! (where possible of course) and a native details/summary shouldn't need aria.

* In actual fact I think the best solution here is that we should't have both the checkbox reveal and the details summary.  The checkbox state should just control the visibility of the config.  Even if they're both enabled it's only 6 fields, which is not that many, and that would be a simple, easy to understand interface for all users (and consistent with other areas of the UI too)

app/javascript/src/component_expander.js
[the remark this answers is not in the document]
chrispymm: Understand the reasoning, but I actually still think it would be slightly better with it even if it is slightly redundant. Mostly because as it is, reading the code makes me have to think *why* is it not consistent and reason it out, whereas if both were there it would be obvious and consistent (if a little OTT).

chrispymm (approved):

PR #1579, opened 2022-06-08
2579 JS test FlowConnectorPath

stevenburnell-moj:
JS tests for FlowConnectorPath, base class to other connector paths (which will follow with their own tests).


chrispymm (approved):
LGTM

PR #1580, opened 2022-06-09
Fix multiple results found issue by adding more specific selector

stevenburnell-moj:
Fix non-isolated element lookup causing duplication in EditableComponent element funcitonality

(second element mirror the correct one)
Screenshot 2022-06-09 at 16 39 05
Screenshot 2022-06-09 at 16 39 15


brenetic (approved):
🚀

brenetic (approved):
🚀

PR #1586, opened 2022-06-16
Services horizontal autoscaling configuration

brenetic:
Add the ability to have set per namespace horizontal autoscaling
configuration.

The three configs possible to set are:

- min replicas
- max replicas
- target cpu utilisation

They are mostly in line with the other platform applications.


mattempty (approved):
🚀

mattempty (dismissed):
🚀

mattempty (approved):
🚀

PR #1595, opened 2022-06-22
Run component validation only when enabled

brenetic:
This fixes a bug whereby the component validation of the user answer
would occur even if they had disabled that validation.


spec/models/component_validations/base_component_validation_spec.rb
```diff
@@ -68,13 +70,11 @@
         end
 
         it 'returns truthy' do
```
chrispymm, line 72: Should probably change the name of the expectation 🙂
```suggestion
        it 'returns falsey' do
```

chrispymm (changes requested):

PR #1604, opened 2022-06-30
Add autocomplete feature flag

njseeto:
### Add feature flag at autocomplete entrypoint
We already have a filter for autocomplete to be visible to devs only however would like to add an extra layer to prevent pollution of the live database.


brenetic (approved):
🚀

PR #1605, opened 2022-06-30
Add in sanitise-html library to prevent XSS attacks in content areas

chrispymm:
Adds sanitize-html library to handle sanitization of content in markdown content areas.  

Should prevent XSS attacks

### DIFF
--- a/app/javascript/src/editable_components.js
+++ b/app/javascript/src/editable_components.js
@@ -31,6 +31,7 @@ const converter = new showdown.Converter({
                     tables: true,
                     disableForced4SpacesIndentedSublists: true
                   });
+const sanitizeHtml = require('sanitize-html');
 const EditableCollectionItemMenu = require('./components/menus/editable_collection_item_menu');
 
 showdown.setFlavor('github');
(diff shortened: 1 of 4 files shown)


app/javascript/src/editable_components.js
```diff
@@ -801,8 +802,7 @@ function convertToMarkdown(html) {
  * For that reason, we are trying to be minimalistic in approach.
  **/
 function sanitiseHtml(html) {
-  html = html.replace(/<([\/\s])?script[^\<\>]*?>/mig, "&lt;$1script&gt;");
-  return html;
+  return sanitizeHtml(html);
 }
```
chrispymm, line 806: I'm not wild about a function `sanitiseHtml()` calling a library `sanitizeHtml()` as its only one letter different.

However I wanted to leave our function intact and call the library from one place, as we are likely to add args to the sanitizeHtml library call.

resolve would be to rename our function, or to alias the library on require

stevenburnell-moj: Hmm... yes, it's not ideal. My first thought was to suggest rename our function but we also have a sanitiseMarkdown(), so our code is being consistently with that. Perhaps go with the alias idea, then, or just live with it. I'll approve the PR and simply revisit if you change your mind.

stevenburnell-moj (approved):

PR #1612, opened 2022-07-08
2579 jstest flowconnectorpath

stevenburnell-moj:
Base class for FlowConnectorPath items.
More (variants) to follow but currently handling those in separate branches to keep the PR smaller and control through tickets.

Note:
This has been an on/off task taking a backseat to recent Analytics tasks so has felt a bit scrappy, with some old commits in there, but everything passes where tests can be applied.


brenetic (approved):
🚀

PR #1618, opened 2022-07-18
2727 global analytics

stevenburnell-moj:
https://trello.com/c/h3595yEp/2727-show-cookie-banner-when-global-analytics-is-configured-for-a-form

Use latest Metadata Presenter 2.17.9
(Workaround) Fix for JS errors when in Preview

PR #1620, opened 2022-07-19
2727 global analytics 2

stevenburnell-moj:
Replacing PR https://github.com/ministryofjustice/fb-editor/pull/1618 for better history.

https://trello.com/c/h3595yEp/2727-show-cookie-banner-when-global-analytics-is-configured-for-a-form

Use latest Metadata Presenter 2.17.9
(Workaround) Fix for JS errors when in Preview

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'delayed_job_active_record'
 #    github: 'ministryofjustice/fb-metadata-presenter',
 #    branch: 'add-branching-title'
 # gem 'metadata_presenter', path: '../fb-metadata-presenter'
-gem 'metadata_presenter', '2.17.8'
+gem 'metadata_presenter', '2.17.9'
 
 gem 'faraday'
 gem 'faraday_middleware'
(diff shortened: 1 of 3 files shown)


brenetic:
The cookie banner will be shown in preview for a form, even when there are no measurement IDs present. Is this the intended behaviour?

stevenburnell-moj:
> The cookie banner will be shown in preview for a form, even when there are no measurement IDs present. Is this the intended behaviour?

That's the intended behaviour on Runner. Preview in Editor is a more interesting question. With the Cookie banner showing is a true representation of the form view, at least until the user makes a choice. I would listen to arguments that it is not necessary to show in just a preview, however. I guess the correct way to handle it would be to have an environment switch that knows it is Preview and update the show_banner method to logic along the lines of 'if not preview and there is no banner cookie to indicate a user choice' show the banner. We could also, put in a workaround of just using CSS to hide it using the Editor stylesheets. What do you think?

chrispymm (approved):
From reading the conversations, I think the JS solution is probably the best.  As we don't really want editor stylesheets in preview - that could cause all sorts of problems.  

The only other way around it I can think is that the banner is hidden by default and is shown by js in the runner. 

This seems to be a simple solution that works though.

PR #1624, opened 2022-07-20
Disable form name update (expect further changes)

stevenburnell-moj:
No design available.

Changes made for https://trello.com/c/nNadIkVp/2720-change-form-name-field-to-read-only
Have disabled the name input and 'Save' button. 
At this point, expecting more tweaks and on-going changes so this is just a start


stevenburnell-moj:
Didn't mean to create full PR, replacing with draft only.

PR #1631, opened 2022-07-22
Preview mechanism for autocomplete component

njseeto:
[Trello](https://trello.com/c/82s692KP/2714-autocomplete-preview-mechanism)

To enable preview in the editor we require an API call to get the items related to a component. This will be different in the
Runner, whereby the items will be present in an environment variable.
The items will also be available in the autocomplete template once a user uploads their items.

### Bump presenter 2.17.10


brenetic (approved):
🚀

PR #1632, opened 2022-07-25
Allow users to send answers as a CSV attachment

brenetic:
This adds the checkbox to the Email Settings page to allow the Editor
user to enable and disable the sending of the CSV attachment of user
answers in both the 'dev' and 'production' environments.

Also add service_csv_output column to the SubmissionSetting table

Screenshot 2022-07-21 at 15 27 40


njseeto (approved):
🚀

PR #1636, opened 2022-07-25
Inject autocomplete items into container during publishing

brenetic:
This adds the ability for the publisher to request the autocomplete
items for a given service and then set them onto the environment
variable AUTOCOMPLETE_ITEMS before adding that to the deployment
configuration in Kubernetes.

For times when there are no autocomplete items for a service we just set
"{}" as the value. The Runners initialiser will just create an empty
hash on start up in those instances.

### DIFF
--- a/app/services/publisher/service_provisioner.rb
+++ b/app/services/publisher/service_provisioner.rb
@@ -27,6 +27,13 @@ class Publisher
       service.to_json.inspect
     end
 
+    def autocomplete_items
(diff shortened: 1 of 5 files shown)


chrispymm (dismissed):
🚀

njseeto (approved):
🚀

chrispymm (approved):
🚀

njseeto (approved):

PR #1637, opened 2022-07-26
Add Crown Presecution Service domain to allow list

brenetic:
Adding in the Crown Prosecution Service domain into the allow list as
they would like to start using the Editor.


mattempty (approved):
🚀

PR #1638, opened 2022-07-26
Handle autocomplete component with no uploaded items

brenetic:
The ResourceNotFound will always be a possibility, especially as it
is always the case immediately after a page has been created. Therefore
there is no need to keep notifying Sentry and we can downgrade the
logger message to be an info instead of a warn.


chrispymm (approved):
🚀

PR #1639, opened 2022-07-26
Update sample csv file

brenetic:
The file previously was actually an example of a CSV uploaded for the
autocompete component.

The actual sample needs to better represent what is submitted at the end
of the form.


mattempty (approved):
🚀

PR #1646, opened 2022-07-28
2720 read only form name

stevenburnell-moj:
Changes made for https://trello.com/c/nNadIkVp/2720-change-form-name-field-to-read-only
Have removed the name input and disabled the 'Save' button. 

These changes are a temporary workaround to avoid issues with generating new form URLs.
A later follow-up task (to be designed) will re-enable the Form Name change functionality but separate it from the URLs.

**Should now look something like:**
Screenshot 2022-07-28 at 13 00 44

### DIFF
--- a/app/views/settings/form_information/index.html.erb
+++ b/app/views/settings/form_information/index.html.erb
@@ -4,7 +4,7 @@
   <div class="govuk-form-group <%= !f.object.errors.empty? ? 'govuk-form-group--error' :'' %>">
     <%= f.label :service_name, class: "govuk-label" do %>
       <%= f.object.class.human_attribute_name :service_name %>
-      <span class="govuk-hint"><%= t('settings.form_information.form_name..hint') %></span>
+      <span class="govuk-hint"><%= t('settings.form_information.form_name.hint') %></span>
     <% end %>
     <% f.object.errors.each do |error|  %>
       <span class="govuk-error-message"><%= error.message %></span>
--- a/app/views/settings/form_information/index.html.erb
+++ b/app/views/settings/form_information/index.html.erb
@@ -9,8 +9,7 @@
     <% f.object.errors.each do |error|  %>
       <span class="govuk-error-message"><%= error.message %></span>
     <% end %>
-    <%# f.text_field :service_name, class: "govuk-input width-responsive-two-thirds", readonly: true %>
-    <p><%= service.service_name %></p>
+    <%= f.text_field :service_name, class: "govuk-input width-responsive-two-thirds", readonly: true, tabindex: -1 %>
     <%= render partial: 'partials/template_warning', locals: { message: t('settings.form_information.form_name.warning_message').html_safe } %>
   </div>
   <%= f.submit t('actions.save'), class: "govuk-button fb-govuk-button", disabled: true %>
--- a/app/views/settings/form_information/index.html.erb
+++ b/app/views/settings/form_information/index.html.erb
@@ -9,8 +9,8 @@
     <% f.object.errors.each do |error|  %>
       <span class="govuk-error-message"><%= error.message %></span>
     <% end %>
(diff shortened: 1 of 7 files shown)


app/views/settings/form_information/index.html.erb
```diff
@@ -1,16 +1,18 @@
-<h1 class="govuk-heading-xl"><%= t('settings.form_information') %></h1>
+<h1 class="govuk-heading-xl"><%= t('settings.form_information.name') %></h1>
 
 <%= form_for @settings, as: :service, url: settings_form_information_index_path(service.service_id) do |f| %>
   <div class="govuk-form-group <%= !f.object.errors.empty? ? 'govuk-form-group--error' :'' %>">
     <%= f.label :service_name, class: "govuk-label" do %>
       <%= f.object.class.human_attribute_name :service_name %>
-      <span class="govuk-hint"><%= t('settings.form_information_help') %></span>
+      <span class="govuk-hint"><%= t('settings.form_information.form_name..hint') %></span>
```
chrispymm, line 7: Got a double `.` in the translation string. Which I'm assuming would casue issues.

chrispymm: Also, I know this is only a temporary change, but this is now some very odd markup.

Firstly, the hint is in the label which it shouldn't be. It should be in its own element attached to an input with `aria-describedby`.

Which leads on to... we have a label that doesn't label anything cos we've removed the thing it's labelling. Which even though a screenrewader user would still be able to hear it if they navigate to it with the rotor, iot would still be announced as a label, they wouldn't be able to find the input it is labelling which would be confusing.

I think the best fix for this would be to keep the input, make it `readonly` and `aria-disabled` then add styles to make it not look like an input.

You could also make the input described by the warning message as well as the hint, enabling a screenreader user to gain the full context on focusing the input.

Something like this:

```
<%= f.label :service_name, class: "govuk-label" do %>
    <%= f.object.class.human_attribute_name :service_name %>
<% end %>

<span id="service_name_hint" class="govuk-hint">
  <%= t('settings.form_information.form_name..hint') %>
</span>

<% f.object.errors.each do |error|  %>
    <span class="govuk-error-message"><%= error.message %></span>
<% end %>

<%= f.text_field :service_name, class: "govuk-input width-responsive-two-thirds", readonly: true, aria-disabled: true, aria-describedby: "service_name_hint service_name_warning" %>

<%= render partial: 'partials/template_warning', locals: { id: "service_name_warning", message: t('settings.form_information.form_name.warning_message').html_safe } %>
```

```
input[readonly] {
  // styles to make it look like a <p>
}
input[readonly]:focus-visible {
 // remove focus styles for those who are not keyboard navigating
}
```

This way the majority of users would see the same as they do now, and if they clicked on the element it would not behave / focus as an input. 

Screenreader users would still be able to focu the input, would hear that the input was disabled and readonly, but would also hear the label, hint and warning which would give them the context to understand why.

Appreciate this is a lot for a temporary fix (but we don't know **how** temporary so it is probably worth doing still)  but at the very least we should sort out the hint being correctly associated to the input for the future.

chrispymm (changes requested):

app/views/settings/form_information/index.html.erb
stevenburnell-moj: Made all the changes but the :focus-visible didn't work. Couldn't find a reason but guessing it might be because I'm using a trackpad and not a mouse (why should that be)?

Even examples on https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible didn't work, when looked for reasons/answers.

Last commit shows, in the end, I turned it into regular text, which solves many of the context issues and more closely matches the design output, until we turn it back into full form-enabled functionality.

Have left form elements (with some of the above changes) in place due to the temporary nature and expectation of form element being re-enabled at some point.

chrispymm (approved):

PR #1648, opened 2022-07-28
Allow autocomplete components to work in preview

chrispymm:
Include required JS / Styles for autocomplete components to work in preview.

Bumps Metadata Presenter to 2.17.13

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,8 +14,8 @@ gem 'delayed_job_active_record'
 # gem 'metadata_presenter',
 #    github: 'ministryofjustice/fb-metadata-presenter',
 #    branch: 'add-branching-title'
-# gem 'metadata_presenter', path: '../fb-metadata-presenter'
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -31,15 +31,6 @@
...
 
-  create_table "global_service_configurations", force: :cascade do |t|
-    t.string "name", null: false
-    t.string "value", null: false
-    t.string "deployment_environment", null: false
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
-    t.index ["deployment_environment"], name: "index_global_service_configurations_on_deployment_environment"
-  end
-
```
brenetic, line 42: i think a rebase has gone wrong somewhere. the schema file shouldn't have changed

brenetic (changes requested):
need to check why the db schema file has changed

brenetic (approved):
🚀

PR #1664, opened 2022-08-04
Fix settings screen accessibility issues

chrispymm:
Fixes for a accessibility issues raised by automated chwecking tools.
- Use `<h1>` instead of `<h2>` for main heading
- Remove `<br />` elements in the list (invalid html) and add margin instead
- Add unique ids to all elements on the settings send by email screen


brenetic (approved):
🚀

PR #1669, opened 2022-08-08
Fix failing test after html changes

chrispymm:
- Fixes a failing test after chaging ids in the template
 - Removes `minimmum: 1` count from a finder as it isn't valid capybara syntax and is ignored.


brenetic (approved):
🚀

PR #1672, opened 2022-08-09
Updates in tests to match html changes

chrispymm:
I'm a muppet and was running the wrong rpsec command locally so the tests needed adjustments to pass.

Couple of small changes to make these accessibility fixes pass tests


brenetic (approved):
🚀

PR #1682, opened 2022-08-10
Autocomplete publish warning

njseeto:
[Trello](https://trello.com/c/dJAvQ1Iy/2755-autocomplete-show-warning-in-publishing-menu-if-an-autocomplete-component-has-no-items)

We want to have warnings on the publishing page when an autocomplete component does not have any associated items.
These warnings should link to the autocomplete component without items.

Screenshot 2022-08-10 at 14 38 03

### DIFF
--- a/app/controllers/publish_controller.rb
+++ b/app/controllers/publish_controller.rb
@@ -5,6 +5,7 @@ class PublishController < FormController
     @published_dev = published?(service.service_id, 'dev')
     @published_production = published?(service.service_id, 'production')
     @publish_warning = PublishPresenter.new(service)
+    @autocomplete_warning = AutocompleteItemsPresenter.new(service, service_autocomplete_items)
   end
 
   def create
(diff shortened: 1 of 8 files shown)


chrispymm (changes requested):
All looks good.  I remember we discussed the backend exposing a boolean `publishable?` in order for me to determine whether or not the `Publish to Live` button should be enabled/disabled. This would be super helpful, should that be in this PR? or should it be a separate thing?

njseeto:
> All looks good. I remember we discussed the backend exposing a boolean `publishable?` in order for me to determine whether or not the `Publish to Live` button should be enabled/disabled. This would be super helpful, should that be in this PR? or should it be a separate thing?

Yep definitely something we should do, probably best we do it in a separate PR as we'll probably need to make a new model rather than a method within the controller that checks these thing. So it might be best to tackle once we know what the `send_from` work will look like

chrispymm (approved):

PR #1686, opened 2022-08-11
Update editor ingress name

brenetic:
Second round migration done so we can remove the '-new' from the ingress
name


njseeto (approved):
🚀

PR #1689, opened 2022-08-12
Error warning messages

chrispymm:
* Adds in new warning message style for blocking errors on the publishing page.
* Fixes error where @autocomplete_items was `nil` on the `publish_controller#create` action

image

### DIFF
--- a/app/controllers/publish_controller.rb
+++ b/app/controllers/publish_controller.rb
@@ -1,11 +1,10 @@
 class PublishController < FormController
-  before_action :assign_form_objects
+  before_action :assign_form_objects, :assign_autocomplete_warning
 
   def index
     @published_dev = published?(service.service_id, 'dev')
     @published_production = published?(service.service_id, 'production')
     @publish_warning = PublishPresenter.new(service)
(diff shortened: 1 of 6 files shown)


app/controllers/publish_controller.rb
```diff
@@ -1,11 +1,10 @@
 class PublishController < FormController
-  before_action :assign_form_objects
+  before_action :assign_form_objects, :assign_autocomplete_warning
```
brenetic, line 2: is this because it is also required for the `create`?

brenetic: just a thought, but to save making the API call until only when it's necessary we could also do this just before we render `:index` on line 20.

brenetic (approved):
🚀 

one comment, happy for it to stay as is if you prefer

app/controllers/publish_controller.rb
njseeto: oh yeah this is a good suggestion. I don't have a strong preference out of the two options, happy with either

njseeto (approved):

PR #1690, opened 2022-08-12
Accessibly disable publish button

chrispymm:
Adds the ability for the 'Publish to Live' button to be accessibly disabled if there are blocking errors^.

Accessibly disabled means using `aria-disabled="true"` which means the button can still be focused by AT users and will be announced as disabled. We then have to ensure the button has no action on click/keyboard interaction.

In order to avoid having to pass the aria-disabled attribute through the `PublishController`, `PublishForm` and `FormDialog` class into the `DialogActivator` class it has been moved into the main view.

The button is initially given the `hidden` attribute, which is then removed by the `DialogActivator` class in order to provide the same progressive enhancement as before. If there is no JS there wil be no activator button, just the form, and the button will be `disabled` normally (we need JS to prevent the submit of an `aria-disabled` button)

^ currently just missing autocomplete items, but soon to include more things.

image

### DIFF
--- a/app/views/publish/index.html.erb
+++ b/app/views/publish/index.html.erb
@@ -32,6 +32,14 @@
 
     <%= render partial: 'autocomplete_warning' %>
 
+    <button class="govuk-button fb-govuk-button"
+        data-module="govuk-button"
+        type="button" 
+        data-fb-action="publish"
+        hidden >
+        <%= t('actions.publish_to_test') %>
+    </button>
+
     <%# TODO: Need an indicator from BE code that shows whether this is first-time publish   %>
     <%#       Currently have hardcoded 'false' value to avoid dialog showing but this should %>
     <%#       be dynamic to control dialog visibility.                                        %>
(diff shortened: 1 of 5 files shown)


app/views/publish/index.html.erb
```diff
@@ -72,12 +80,21 @@
 
     <%= render partial: 'autocomplete_warning' %>
 
+    <button class="govuk-button fb-govuk-button"
+            data-module="govuk-button"
+            type="button" 
+            aria-disabled="<%= @autocomplete_warning.messages.blank? ? 'false' : 'true' %>"
```
brenetic, line 86:
```suggestion
            aria-disabled="<%= ActiveModel::Type::Boolean.new.cast(@autocomplete_warning.messages.blank?).to_s %>"
```

brenetic (approved):
🚀 

one comment, but again not a PR stopper

app/views/publish/index.html.erb
brenetic: actually might not even be possible. am unsure if `ActiveModel` is available in a view

chrispymm: I believe your replacement would be the wrong way round, so I'd need:
`ActiveModel::Type::Boolean.new.cast(!@autocomplete_warning.messages.blank?).to_s`
or maybe 
`ActiveModel::Type::Boolean.new.cast(@autocomplete_warning.messages.length).to_s`

brenetic: ah yeah quite right:

`ActiveModel::Type::Boolean.new.cast(@autocomplete_warning.messages.present?).to_s`

either way, don't worry :)

PR #1694, opened 2022-08-16
Fix autocomplete error for malformed CSV

njseeto:
The CSV::Malformed error in the `file_contents` method is triggered too late, we have moved this method to the CsvValidator so it can be triggered earlier.


brenetic (approved):
🚀

PR #1696, opened 2022-08-16
Update cookie notice with content from Mark's form

stevenburnell-moj:
Cookie notice now directly added by fb-metadata-presenter but not a final solution.
Footer links still rely on the loop generated from this en.yml file. 
Have blanked the cookie body content to try and reduce some confusion as to the origin.


njseeto:
This change has been rolled into this PR: https://github.com/ministryofjustice/fb-editor/pull/1741

PR #1699, opened 2022-08-17
Trigger validation if autocomplete answer is changed to be invalid

chrispymm:
This matches the PR adding this functionalityinto the runner.

The autocomplete component only updates the underlying `<select>` element when a valid option is chosen. If a user sets an answer, then returns to change their answer and either blanks the autocomplete or enters an invalid/partial string the submitted value will not be changed. 

This is strange UX and confusing for the user.

On autocomoplete pages, on submit, we compare the value in the autocompleteinput with the underlying select if they are different we set the selects value to empty in order to trigger validation

### DIFF
--- a/app/javascript/packs/runner_application.js
+++ b/app/javascript/packs/runner_application.js
@@ -8,9 +8,9 @@ require("@rails/ujs").start()
 const accessibleAutocomplete = require("accessible-autocomplete")
 import 'accessible-autocomplete/dist/accessible-autocomplete.min.css' 
 
-const elements = document.querySelectorAll('.fb-autocomplete');
-
-Array.prototype.forEach.call(elements, function(element) {
+// Initialise autocomplete components
+const autocompleteElements = document.querySelectorAll('.fb-autocomplete');
+Array.prototype.forEach.call(autocompleteElements, function(element) {
   accessibleAutocomplete.enhanceSelectElement({
     defaultValue: '',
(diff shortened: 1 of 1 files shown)


app/javascript/packs/runner_application.js
```diff
@@ -19,6 +19,34 @@ Array.prototype.forEach.call(elements, function(element) {
...
+    const autocompleteSelect = form.querySelector('.fb-autocomplete');
+    
+    // if the select is not empty and the values do not match or if the
+    // autocomplete is empty trigger validation
+    if(autocompleteSelect.value != '' && !autocompleteSelect.value.includes(autocompleteField.value) || autocompleteField.value == '') {
+      autocompleteSelect.value = '';
+    }
+  });
+}
+
```
stevenburnell-moj, line 49: This looks like duplicated code effort of the stuff I've just been commenting on in the Runner. Presumably, if it were shifted to the component (as per comments on other PR) then you also wouldn't have to duplicated here ??

stevenburnell-moj (commented):
We really need to figure out how to share the code between Editor and Runner. I'm guessing (after I added the other comment), that you've simply cut-and-pasted the file like we did on something else that needed to be shared. This is going to become a maintenance issue at some point, I'm sure :-(

stevenburnell-moj (approved):

PR #1700, opened 2022-08-18
Encode special characters and account for Byte Order Mark in CSV file

njseeto:
[Trello](https://trello.com/c/Zxje7j0O/2826-autocomplete-parsing-special-characters-in-csv-files)

## Ensure csv files are utf-8 encoded when read

Prior to this change, certain characters (eg: Å or é) would not pass validation due to them not being encoded appropriately. This would mean csv files that contained these characters would be rejected and unable to be uploaded as options in the autocomplete component.

Screenshot 2022-08-18 at 09 23 04

## Additional csv validator test for Byte Order Mark characters
We want to allow csv's which are created with Byte Order Mark characters to be uploaded especially as this seems to be the way that Excel saves CSV files by default.

### DIFF
--- a/app/models/autocomplete_items.rb
+++ b/app/models/autocomplete_items.rb
@@ -20,6 +20,6 @@ class AutocompleteItems
   end
 
   def file_contents
-    @file_contents ||= CSV.read(file.path)
+    @file_contents ||= CSV.read(file.path, encoding: 'utf-8')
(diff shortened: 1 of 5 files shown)


brenetic (approved):
🚀

brenetic (approved):
🚀

PR #1701, opened 2022-08-18
Make admin area service search work

chrispymm:
Pass the search param on the administrate serviecs page through to the metadata api to be able to search the services.


brenetic (approved):
🚀

PR #1710, opened 2022-08-19
Update cookies acceptance tests

stevenburnell-moj:
Cookie page is no longer editable so:
- Fix standalone page test by using Privacy page
- Add test to ensure Cookie page is not editable


brenetic:
can we squish the commits down please

brenetic (approved):
🚀

PR #1712, opened 2022-08-22
Do not allow empty fields when there is a Value column in CSV file

H3ll3m4:
See [Trello](https://trello.com/c/Tl4aHGip/2827-autocomplete-do-not-allow-empty-fields-when-there-is-a-value-column) and [Trello](https://trello.com/c/7ySh2qbd/2834-autocomplete-update-error-message-on-csv-upload)

## Do not allow empty fields when there is a Value column in CSV file
Currently when a user add an file with an empty cell in the value
column, we use the corresponding Text column value.
We assume there could be a mistake in the csv file and would prefer to
confirm with the user.
Adding tests, fixture and implementation in the validator.
 
## Remove conditional Filling the Value property to be the Text property
The validator doesn't allow empty cells anymore and this condition is not required anymore

## Add error when there is an empty 'Text' row
We currently do not have a helpful error message for when a user attempts to upload a CSV with empty Text rows.
For the associated test we have created a csv file rather than add another fixture to the project.

### DIFF
--- a/app/controllers/api/autocomplete_controller.rb
+++ b/app/controllers/api/autocomplete_controller.rb
@@ -13,7 +13,7 @@ module Api
           service_id: params[:service_id],
           component_id: params[:component_id],
           created_by: service.created_by,
-          data: items_hash(@items.file_values)
+          data: items_hash(@items.file_rows)
         )
         return head :created unless response.errors?
 
--- a/app/controllers/api/autocomplete_controller.rb
+++ b/app/controllers/api/autocomplete_controller.rb
@@ -33,6 +33,10 @@ module Api
 
     private
 
+    def has_headers?
+      @items.file_headings == %w[text value]
+    end
+
     def autocomplete_items_file
       @autocomplete_items_file ||= params.dig(:autocomplete_items, :file)
     end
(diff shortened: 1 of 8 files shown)


app/controllers/api/autocomplete_controller.rb
```diff
@@ -33,6 +33,10 @@ def create
 
     private
 
+    def has_headers?
+      @items.file_contents.first.count == 2
```
brenetic, line 37: this would also return true if the first row of the csv is `Scooby,Do` as opposed to `Text,Value`.

maybe something like `@items.file_headings == %w[text value]` or similar if we specifically want to check the headers are in fact correct and present?

njseeto: oh yes, this is indeed an oversight! Have updated.

app/controllers/api/autocomplete_controller.rb
```diff
@@ -33,6 +33,10 @@ def create
 
     private
 
+    def has_headers?
+      @items.file_headings == %w[text value]
```
brenetic, line 37: forgot about this! 😆 

`@items.files_headings` now doesn't return things that are downcased already

njseeto: 😅 good spot. Have changed this to look for the number of headings rather than the specific strings as the validator should weed out any header issues beforehand.

brenetic (approved):
🚀

PR #1715, opened 2022-08-23
Update to govuk-frontend 4.3.1 (latest)

chrispymm:
Changes required to update to the latest version of govuk-frontend (4.3.1)
changes mostly consist of
* changing `.govuk-hint` elements from `<span>` to `<div>`
* changing `.govuk-error-message` elements from `<span>` to `<p>`

### DIFF
--- a/app/javascript/packs/runner_application.js
+++ b/app/javascript/packs/runner_application.js
@@ -64,4 +64,7 @@ if(autocompleteComponent) {
 
 // Little bit hacky but we want to prevent the
 // Cookie banner from showing in preview mode.
-document.getElementById("govuk-cookie-banner").style.display = "none";
+const cookieBanner = document.getElementById("govuk-cookie-banner");
+if(cookieBanner) {
+  cookieBanner.style.display = "none";
+}
--- a/app/javascript/styles/_index.scss
+++ b/app/javascript/styles/_index.scss
@@ -1002,7 +1002,7 @@ html {
 }
 
 // For when a user is signed in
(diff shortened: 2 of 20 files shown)


app/javascript/packs/runner_application.js
```diff
@@ -64,4 +64,7 @@ if(autocompleteComponent) {
 
 // Little bit hacky but we want to prevent the
 // Cookie banner from showing in preview mode.
-document.getElementById("govuk-cookie-banner").style.display = "none";
+const cookieBanner = document.getElementById("govuk-cookie-banner");
+if(cookieBanner) {
+  cookieBanner.style.display = "none";
+}
```
chrispymm, line 70: This change is not directly related, but I was getting a console error locally due `document.getElementById('govuk-cookie-banner')` returning nil, so I wrapped it to ensure the eleemnt exists before we try to style it.

app/javascript/styles/_index.scss
```diff
@@ -1002,7 +1002,7 @@ html {
 }
 
 // For when a user is signed in
-.govuk-navigation-header {
+.fb-navigation-header {
```
chrispymm, line 1005: Renamed this class, as it is not a design system class (our header differs from the design system)  

Figured that it would help avoid confusion if the prefix was `fb` to indicate it is a formbuilder/mojforms class.

There's more to do along these lines.  But restricted it to just this one in this PR.

stevenburnell-moj (approved):

PR #1726, opened 2022-08-25
Fix FlowConnectorPath refactor (after issues caused by first attempt)

stevenburnell-moj:


test/flow_connector_path/down_forward_down_backward_up_path/component_test.js
```diff
@@ -0,0 +1,94 @@
...
+
+  describe("Component", function() {
+    var created;
+    const POINTS = {
+                     from_x: 1951,
+                     from_y: 125,
+                     to_x: 1850,
+                     to_y: 562.5,
+                     via_x: 525,
+                     via_y: 312.5
```
chrispymm, line 17: I'm very willing for this to be me no understanding how 'via' works, but I don;t see how these co-ordinates result in a `DownForwardDownBackwardUpPath`

chrispymm (commented):
Couple of queries, probably no issue, just wanted to double check.

test/flow_connector_path/down_forward_down_backward_up_path/component_test.js
stevenburnell-moj: I understand why you would think this but I believe it is the result of needing to calculate element position/dimensions in order to get the via_x/y coordinate. Essentially, it is relative to the starting (from_x/y) point, not the container.

**Longer explanation**
Since the from_x/y and and to_x/y show it needs to go backward, but the design says it needs to go forward first (it's a branching line so needs to pass through end of condition comment element), the via_x/y is a point that tells it how far to go forward (somewhere around the end of that comment element), before it goes backward, if you see what I mean.

It's the via_x/y is the point between forward and down, in the path, and is relative to the from_x/y, not the whole container (which I think from_x/y and to_x/y are - so that's why they are larger numbers). It could probably be calculated using it's position from the container, but that is a larger calculation than would also involve subtracting the starting (from_x/y) point, just to see how far forward the line should go. Instead it's pretty much the value we want (but we need also to subtract the curve size numbers).

I tapped into the code to output the points data of a real paths to get (not calculate - I'm not that smart!) most of these testing points you see in the code. I've just done that again for this DownForwardDownBackwardUpPath example to show them in the console (similar to what you see above). I've also attached a couple screenshots that hopefully help to illustrate what I'm trying to explain in case none of this is very clear.

**Path showing points data and element outline**
Screenshot 2022-08-31 at 11 36 20

**Browser ruler showing (approximate) dimensions of the via_x/y position** (bottom right blue area)
Screenshot 2022-08-31 at 11 36 50

test/flow_connector_path/down_forward_down_forward_path/component_test.js
```diff
@@ -0,0 +1,95 @@
...
+    var created;
+    // Warning: made up coordinates because failed to create line in browser
+    const POINTS = {
+                     from_x: 701,
+                     from_y: 125,
+                     to_x: 1550,
+                     to_y: 662.5,
+                     via_x: 525,
+                     via_y: 312.5
+                   }
```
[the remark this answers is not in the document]
stevenburnell-moj: See above for lengthy explanation. Basically via_x/y is relative to starting point, from_x/y.

chrispymm (approved):

PR #1728, opened 2022-08-26
Fix layering issue caused by svg overlapping thumbnail

stevenburnell-moj:
Originally, svg elements on the Flow page had been given a 100px by 100px sizing. This was not desired but helped avoid a bug in Chrome where anything less (ie. the preferred 1x1) resulted in a loss of FlowConnectorLine visibility. As this was recently noticed and flagged to be a bug (due to causing overlapping issues when attempting to click the first thumbnail in a form overview), it was given another look. Seems Chrome (probably some update since) can now work with the preferred 1x1 sizing, so the CSS has been updated and bug eliminated.

See screenshots for Firefox, Chrome, and Safari.

Cursor cannot be seen on capture but it is placed in top left corner of first image, for each, where you can see it is triggering the hover style that reveals the three button menu activator. Lines are clearly still visible.

**Firefox**
Screenshot 2022-08-26 at 10 09 02

**Chrome**
Screenshot 2022-08-26 at 10 08 13

**Safari**
Screenshot 2022-08-26 at 10 09 42

### DIFF
--- a/app/javascript/styles/_index.scss
+++ b/app/javascript/styles/_index.scss
@@ -575,19 +575,13 @@ html {
   // allow for use of hover over event in making Conneciton menus appear.
   // However, although this setting works fine in Firefox, there seems to
   // be a bug in Chrome that simply hides FlowConnectorPath elements from
-  // being seen. By accidental tinkering, it seems that giving them a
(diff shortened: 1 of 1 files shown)


brenetic (approved):
🚀

PR #1730, opened 2022-08-26
Fix broken links by stripping out injected angle brackets in markdown…

stevenburnell-moj:
… notation

Recent changes to strip out HTML from markdown has left a bug where HTML converted to markdown that produces link format like this:

1.
`[link text](<http://some.url/here>)`

Results in broken HTML link output that looks like this:

2.
`[link text]()`

Rather than revert the HTML sanitising function that has likely caused this, a simple filter line has been added to convert the markdown, stripping out those angle brackets. The markdown (1) above will now look like this:

3.
`[link text](http://some.url/here)`

This produces the required HTML output with a link that is just the visible text "link text" (as per this example).

See screenshots for actual examples in play.

**Original HTML output**
Screenshot 2022-08-26 at 15 02 28

**Converted to this on edit**
(note the angle brackets)
Screenshot 2022-08-26 at 15 03 04

**Converted back to HTML**
(without editing - shows original link now broken)
Screenshot 2022-08-26 at 15 03 18

### DIFF
--- a/app/javascript/src/editable_components.js
+++ b/app/javascript/src/editable_components.js
@@ -806,11 +806,27 @@ function sanitiseHtml(html) {
 }
 
 /* Opportunity safely strip out anything that we don't want here.
- * 1. Something in makeMarkdown is adding <!-- --> markup to the result so we're trying to get rid of it.
- * 2. ...
+ *
+ * 1. Something in makeMarkdown is adding <!-- --> markup to the result
+ *    so we're trying to get rid of it.
+ *
+ * 2. sanitizeHTML is altering automatic link syntax by removing
+ *    everything in (including) angle brackets added by showdown.
+ *
+ *    e.g. [link text](<http://some.url/here>)
+ *
+ *         results in this incorrect markdown (and output):
+ *         [link text]()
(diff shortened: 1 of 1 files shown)


app/javascript/src/editable_components.js
```diff
@@ -811,6 +811,7 @@ function sanitiseHtml(html) {
  **/
 function sanitiseMarkdown(markdown) {
   markdown = markdown.replace(/\n<!--.*?-->/mig, "");
+  markdown = markdown.replace(/\]\(\<(.*?)\>\)/mig, "]($1)");
```
chrispymm, line 814: Can we add a comment here as to what this regex is doing (and why) for future reference, and to the one above would be helpful too.

chrispymm (changes requested):

app/javascript/src/editable_components.js
stevenburnell-moj: Done.
Commit: abd3ee858202a39f1c74ca14e9a16dda22f66f78

chrispymm (approved):

PR #1734, opened 2022-09-01
Updating database with autocomplete items

H3ll3m4:
[Trello](https://trello.com/c/AQN8Fcmj/2845-autocomplete-add-versionid-column-in-autocomplete-db)

To ensure proper auditing of published forms, we're adding identifier for autocomplete items: autocomplete ids. They are rows ids in the metadata api items table. They are different from the autocomplete component uuids. We're saving them to the database at publishing stage. Tests performed locally.


brenetic (approved):
🚀

njseeto (approved):
🚀

PR #1743, opened 2022-09-07
Testable temporary 2900 email address markdown

stevenburnell-moj:
Apparently our documentation recommends adding email in markdown in this form:
<my@email.address>

After the introduction of sanitize-html script, because that is within angle brackets, instances of those found in the markdown will get removed. 

This fix attempts to bypass that drawback of the sanitize-html script by replacing instances with different identifiers, to hide from sanitize-html, then running a reversal action after the sanitize-html script has done its thing.

Note: We could also have simply found and replaced <my@email.address> with [my@email.address](mailto:my@email.address) which, presumably, would also work, but the decision against doing this was to avoid confusion by changing the user input before they save anything. It can be argued that confusion still exists since, after saving, the returned page will now show markdown formatted with the long form [my@email.address](mailto:my@email.address)

For the sake of recording, the regex built to fix this issue was coded and tested in the browser console first. We do not currently have any JS tests for Editable Component code (and not sure this will be possible to capture) but it would be useful to included the development script within such tests (again, if possible).

To avoid losing the development script/test, I'm including below...

### Testing markup
```
Use this service to:

- do something
- update your name, address or other details
- do something else

<img src="foo">

<something>

[Google](www.google.com)

<steven@steven.com>

Registering takes around 5 minutes.
<not@wanted.com/>

some text <with@some.email> in it
<not @wanted.com >
```

### Browser console script
```
var input = "Use this service to:\n\
\n\
- do something\n\
- update your name, address or other details\n\
- do something else\n\
\n\
<img src=\"foo\">\n\
\n\
<something>\n\
\n\
[Google](www.google.com)\n\
\n\
<steven@steven.com>\n\
\n\
Registering takes around 5 minutes.\n\
<not@wanted.com/>\n\
\n\
some text <with@some.email> in it\n\
<not @wanted.com >\n";

console.clear();
console.log(input, "\n\n");

input = input.replace(/^(.*?)$/mi, "\n\ADD SYNTAX...\n==================\n$1");
input = input.replace(/\<([\w\-]+@{1}[\w\-\.]+)>/mig, "$mailto$1$mailto")
console.log(input, "\n\n");


input = input.replace(/ADD SYNTAX\.\.\.\n==================\n/, "");
input = input.replace(/^(.*?)$/mi, "\n\AND CLEAR OTHERS...\n==================\n$1");

input = input.replace(/\<.*?\>/mig, "")
console.log(input, "\n\n");


input = input.replace(/AND CLEAR OTHERS\.\.\.\n==================\n/, "");
input = input.replace(/^(.*?)$/mi, "\n\AND BACK TO BRACKETS...\n==================\n$1");

input = input.replace(/\$mailto([\w\-]+@{1}[\w\-\.]+)\$mailto/mig, "<$1>")
console.log(input, "\n\n");

```


stevenburnell-moj:
Closing this PR because I've added it to https://github.com/ministryofjustice/fb-editor/pull/1738 due to things getting complicated over shared changes.

PR #1745, opened 2022-09-09
Remove Global Service Configuration table

brenetic:
It looks like this table was accidentally added to the schema. There is no related ActiveRecord model. The global config for a service is set in an initialiser on the Rails config object.


njseeto (approved):
🚀

PR #1746, opened 2022-09-09
Add From Address

njseeto:
[Trello](https://trello.com/c/YZFcCLeN/2904-add-route-for-from-address-and-rough-template-for-settings-screen) 
[Trello](https://trello.com/c/jbZihH38/2820-add-table-to-record-from-address-email-and-status) 
### Add From Address configuration page to Settings
This adds the From Address page with the input field for updating the Send From Email for a form.

### Create `from_address` table and Initialise all new services with a default `from_address`
We would like all forms to have a default from address, a from_address record will be created in the from_address table when a user creates a new service.
When a user adds a new email address, it is given a status of 'unverified'. These statuses will change once we start interacting with the AWS SES API.
When a user does not enter an email we will use the default email address, ie: the existing `from_address` record will be overwritten with the default email address.

### On the Submission Settings Page
Screenshot 2022-09-09 at 13 09 43

### 'From' Address page
Screenshot 2022-09-09 at 13 09 49


chrispymm (approved):

brenetic (approved):
🚀

PR #1755, opened 2022-09-15
From address settings screen

chrispymm:
This PR: 
* Adds in all the required markup for the 'From Address' settigns screen.
* Tidies up the settings screen lists to match the designs (increased spacing) 

It also introduces 2 new gems:

**GovUK Frombuilder gem**
We use this on the runner, but for some reason it doesn;t seem to have been used in the editor.  Adding this in allows us to use some nice helpers for generating design system forms, and keep templates less cluttered.

**GovUK Components**
Also from the DfE and built to pair with the formbuilder gem.  This provides a set of design system components using [ViewComponent](https://viewcomponent.org/). Like the formbuilder gem this will also enable us to include GDS compatible components in our templates, and avoid lots of markup repetition which we currently have.  It will also allow us to potentially create our own reusable components in a similar way.

The intention is that going forward, views can be tidied up to utilise these components where possible. Which should hopefully improve consistency and avoid repetition (and having to constantly copy/paste out of the design system)

**One further thing to note:**
```
<%= @presenter.message[:text].html_safe %>
```
This line above was triggering a Brakeman XSS error, as it believed it (incorrectly) to be directly outputting an unescaped model property.

I have currently mitigated this by adding `--report-direct` to the brakeman command [as suggested here](https://brakemanscanner.org/docs/warning_types/cross_site_scripting/)  this means brakeman will only report direct use of unescaped params.


.circleci/config.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     steps:
       - checkout
       - ruby/install-deps
-      - run: bundle exec brakeman -q --no-pager -i brakeman.ignore
+      - run: bundle exec brakeman -q --no-pager --report-direct -i brakeman.ignore
```
brenetic, line 52: i think it might be better to add something to the brakeman.ignore file instead of doing a global flag. we have precedence for this.

i think it's bundle exec brakeman -i and then point it to the existing file when prompted

brenetic (changes requested):

.circleci/config.yml
chrispymm: Done.

njseeto (approved):
LGTM! 💯

PR #1756, opened 2022-09-15
Implementation of From address setting with AWS SES

H3ll3m4:
### Created a From Address model

In our model we validate for correct domain. The database FromAddress record is updated by the AWS SES client.

### AWS SES Client

We set up the AWS SES SDK. We have injected the AWS secrets in the test environment only. We added an AWS SES adapter for test and live; as well as a local adapter to test in development.

### Handling AWS SES errors and adding tests

When we call get_email_identity on the AWS SES client we expect to receive an EmailServiceNotFound error. This is because the email submitted by the user doesn't exist in AWS yet.

When AWS call fails we expect to get the EmailServiceError.


app/controllers/settings/from_address_controller.rb
```diff
@@ -21,4 +29,12 @@ def assign_from_address
...
     @from_address = FromAddress.find_or_initialize_by(service_id: params[:id])
   end
+
+  def email_service
+    if Rails.env.development?
+      EmailService::Adapters::Local.new
+    else
+      @email_service ||= EmailService::Adapters::AwsSesClient.new
+    end
+  end
```
brenetic, line 39:
```suggestion
@email_service ||= begin
  if Rails.env.development?
    EmailService::Adapters::Local.new
  else
    EmailService::Adapters::AwsSesClient.new
  end
end
```

deploy-eks/fb-editor-chart/templates/deployment.yaml
```diff
@@ -245,6 +256,16 @@ spec:
...
+          - name: AWS_SES_SECRET_ACCESS_KEY
+            valueFrom:
+              secretKeyRef:
+                name: fb-editor-secrets-{{ .Values.environmentName }}
+                key: aws_ses_secret_access_key
+          - name: AWS_SES_ACCESS_KEY_ID
+            valueFrom:
+              secretKeyRef:
+                name: fb-editor-secrets-{{ .Values.environmentName }}
+                key: aws_ses_access_key_id
```
brenetic, line 268: the background task is now a kubernetes cronjob as opposed to a delayed job queue therefore these will not be needed in the worker config.

they can be injected into the config for the cronjob much in the same way the sessions_trim config injects things

brenetic (changes requested):

deploy-eks/fb-editor-chart/templates/deployment.yaml
njseeto: Good shout. Done.

brenetic (dismissed):
🚀

app/controllers/settings/from_address_controller.rb
```diff
@@ -1,14 +1,22 @@
...
+      from_address: @from_address,
+      from_address_params: from_address_params,
+      email_service: email_service
+    )
 
-    redirect_to settings_from_address_index_path(params[:id])
-  rescue ActiveRecord::RecordInvalid
-    render :index
+    if from_address_creation.save
+      redirect_to settings_from_address_index_path(params[:id])
```
brenetic, line 16: if we are going to use the service_id from the service object a la line 30 then we should be consistent in our controller and not use params[:id]

app/controllers/settings/from_address_controller.rb
```diff
@@ -1,14 +1,22 @@
 class Settings::FromAddressController < FormController
   before_action :assign_from_address
 
-  def index; end
+  def index
+    @from_address_presenter = FromAddressPresenter.new(@from_address)
```
brenetic, line 5: if we are already assigning the presenter in the before_action we dont need to do it again here

config/locales/en.yml
```diff
@@ -283,7 +283,7 @@ en:
       lede: Emails sent by your form, either to you or to people filling it in, will come from this address
       messages:
         default: You need to change the default no-reply-moj-forms address before your form can go live
-        pending: <strong>Now check the email you set for a validation link</strong><br /><br />The email will come from <b>Amazon Web Services</b>. The link is valid for 24 hours. 
+        pending: <strong>Now check the email you set for a validation link</strong><br /><br />The email will come from <b>Amazon Web Services</b>. The link is valid for 24 hours.
```
brenetic, line 286: oh i see, hence the question from @chrispymm about brakeman. 

an alternative is to check the from_address_status in the view and apply the strong there, keeping html out of the locales. could probably then remove the brakeman issue? 🤔 

up to you both

deploy-eks/fb-editor-chart/templates/deployment.yaml
```diff
@@ -133,6 +133,19 @@ spec:
...
+{{- if eq .Values.from_address "enabled" }}
+          - name: AWS_SES_SECRET_ACCESS_KEY
+            valueFrom:
+              secretKeyRef:
+                name: fb-editor-secrets-{{ .Values.environmentName }}
+                key: aws_ses_secret_access_key
+          - name: AWS_SES_ACCESS_KEY_ID
+            valueFrom:
+              secretKeyRef:
+                name: fb-editor-secrets-{{ .Values.environmentName }}
```
brenetic, line 145: yeah this might fail when going to Live as there are no creds for Live in fb-editor-deploy

brenetic (changes requested):

config/locales/en.yml
njseeto: Correct me if I'm wrong here @chrispymm but I'm not sure we can apply the strong in the view with the 'FromAddressPresenter' in it's current state. Currently the whole message is passed into the view as `@presenter.message[:text]`. If we want to apply strong in the view we would have to rethink how to pass that particular message (status: 'pending') in from the presenter as it is only the first half of the message that is bold - I'm happy to give this a shot if we'd like but perhaps we should tackle is in a separate PR?

brenetic (approved):
🚀

PR #1758, opened 2022-09-16
Fix CSV files submitted with carriage return at the end

H3ll3m4:
[Trello](https://trello.com/c/9Nu15mlx/2876-carriage-return-texting-fix-for-autocomplete)

Valid CSV files should not have a carriage return at the end. We strip any blank rows at the end of the file to format submitted CSV files correctly. Added unit tests and moved all autocomplete fixtures used for tests into a folder named Autocomplete.


app/models/autocomplete_items.rb
```diff
@@ -21,7 +21,8 @@ def to_partial_path
   end
 
   def file_contents
-    @file_contents ||= CSV.read(file.path, encoding: 'bom|utf-8')
+    contents = File.read(file.path, encoding: 'bom|utf-8').strip
+    @file_contents ||= CSV.parse(contents)
```
brenetic, line 25:
```suggestion
    @file_contents ||= begin
          contents = File.read(file.path, encoding: 'bom|utf-8').strip
          CSV.parse(contents)
     end
```

H3ll3m4: 🚀

brenetic (approved):
🚀

njseeto (approved):
💯

PR #1760, opened 2022-09-16
2918 (update and add to) component ActivatedMenu JS tests

stevenburnell-moj:
Precursor action to creating JS tests for other menu components that inherit from the ActivatedMenu class.


chrispymm (commented):
I can't help but think that `currentFocusIndex` should actually be a private property on `ActivatedMenu` as it shoudln't really be set from outside the component. All focus manipulation should be done via the public `focus`-related methods

I can't think of a reason why we would want to be able to directly set the focus index from outside the component, that you couldn't achieve using the `focus(index)` method.

It's not a massive issue, and I'm not going to block merge if you don't want to resolve it (especially as it's my oversight in the first place!)

stevenburnell-moj:
> currentFocusIndex should actually be a private property 
Yes, I don't think I was looking to optimise/improve anything but that sounds like a sensible suggestion and now would be as good a time as any to make such a change. Will have a look and come back to you when done.

chrispymm (approved):

PR #1762, opened 2022-09-20
Allow `justice.gsi` domain for FromAddress emails

njseeto:
[Trello](https://trello.com/c/1MBG56LO/2922-from-address-add-gsi-domain-to-email-validation)
There are some teams in MoJ that use the `justice.gsi` domain. As such, we should allow teams to set up their From Address with this domain.


njseeto:
Closing as this domain has been decommissioned.

PR #1763, opened 2022-09-20
Add back links to all settings screens

chrispymm:
This PR adds a few visual improvements to match the Figma designs:
* Add backlinks to all the setting screens
* Correct the top and bottom title spacing (both with and without back button)

image

image

This is implemented by adding 2 helpers:
* A back link helper 
* A setting screen helper - a wrapper for the page grid container divs and the back button (if needed) allowing new settings screens to be easily created with the correct layout/formatting and back link.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -66,6 +66,15 @@ module ApplicationHelper
     flow_object.title || service.find_page_by_uuid(flow_object.uuid).title
   end
 
+  def fb_back_link(path)
+    link_to path, class: 'fb-back-link' do
+      concat tag.span '<svg width="7" height="12" viewBox="0 0 7 12" fill="none" role="image" aria-hidden="true">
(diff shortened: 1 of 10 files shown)


app/helpers/application_helper.rb
```diff
@@ -66,6 +66,15 @@ def flow_title(flow_object)
     flow_object.title || service.find_page_by_uuid(flow_object.uuid).title
   end
 
+  def fb_back_link(path)
```
brenetic, line 69:
```suggestion
  def editor_back_link(path)
```
❓

chrispymm: Yeah, I toyed with this....

In a lot of places we use `fb` as the prefix still, which is obviously outdated, but continuing to use it adds consistency... so do you change and have differing prefixes across the product or keep it the same and have an old one? 🤷‍♂️

On reflection though I think here your change makes sense, as we don't actually have any fb prefixed helpers, and editor is probably more clear as to where it is used.

brenetic (approved):
🚀

PR #1764, opened 2022-09-20
Fix to deployment fails due to acceptances tests not updated

H3ll3m4:
Acceptances tests failed due to move of fixtures files in a specific autocomplete folder.

Update to correct path in acceptances tests was required.


brenetic (approved):
🚀

PR #1766, opened 2022-09-21
Add email domain allowlist for FromAddress

njseeto:
[Trello](https://trello.com/c/uG6KFmbT/2952-from-address-move-justice-and-digitaljustice-domains-into-an-allowlist)

Moved allowed domains into a validator that checks against an allow list. We want to be able to easily add domains to the allow list as the app grows.

### DIFF
--- a/app/validators/email_domain_validator.rb
+++ b/app/validators/email_domain_validator.rb
@@ -1,0 +1,21 @@
+class EmailDomainValidator < ActiveModel::Validator
+  ALLOWED_DOMAINS = [
+    'justice.gov.uk',
+    'digital.justice.gov.uk'
+  ].freeze
+
+  def validate(record)
+    return if record.email.blank?
+
+    unless allowed_domain?(record.email)
+      record.errors.add(:base, I18n.t('activemodel.errors.models.from_address.invalid_domain'))
+    end
+  end
+
+  private
+
+  def allowed_domain?(email)
+    domain = email.split('@').last
+    domain.in?(ALLOWED_DOMAINS)
+  end
+end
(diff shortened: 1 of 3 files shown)


app/validators/email_domain_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    unless allowed_domain?(record.email)
+      record.errors.add(:base, I18n.t('activemodel.errors.models.from_address.invalid_domain'))
+    end
+  end
+
+  private
+
+  def allowed_domain?(email)
+    domain = email.split('@').last
```
brenetic, line 18:
```suggestion
    domain = Mail::Address.new(email).domain
```

app/validators/email_domain_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+    ALLOWED_DOMAINS.any? { |allowed_domain| allowed_domain[domain] }
```
brenetic, line 19: maybe the reverse?

```suggestion
    domain.in?(ALLOWED_DOMAIN)
```

app/validators/email_domain_validator.rb
njseeto: ooo this is handy! Had no idea this existed.

njseeto: Unfortunately the `Mail::Address` class requires a correctly formatted email, otherwise it throws an error. One way around this is to maybe add a proc to run if the email has passed the email format validation?

```
 validates_with EmailDomainValidator, if: proc { |obj| obj.errors.empty? }, allow_blank: true
```

brenetic (approved):
🚀

PR #1767, opened 2022-09-22
Refactor data controller for extraction

chrispymm:
Extracts the `DataContoller` class into a separate `SubmitHandler` class that does the following:

* Enables/disables the submit button as needed
* Updates the content of the save button label / accessible description as needed
* Optionally handles prevention of pageunloading when there are unsaved changes

This allows this same functionality to be re-used across other forms in the interface.

### DIFF
--- a/app/javascript/src/controller_pages.js
+++ b/app/javascript/src/controller_pages.js
@@ -19,6 +19,7 @@
 const  { 
   updateHiddenInputOnForm,
   stringInject,
+  mergeObjects
 }  = require('./utilities');
 
 const {
@@ -61,6 +62,12 @@ class PagesController extends DefaultController {
         break;
     }
   }
+
+  updateComponents() {
+    $(".fb-editable").each(function() {
+      $(this).data("instance").save();
+    });
+  }
 }
 
 /* ------------------------------
@@ -68,7 +75,16 @@ class PagesController extends DefaultController {
  * ------------------------------ */
 PagesController.edit = function() {
   var view = this;
-  var dataController = new DataController(view);
+  var $form = $("#editContentForm");
+  var dataController = new DataController($form, {
+    text: {
+      clean: view.text.actions.saved,
+      dirty: view.text.actions.save,
+      busy: view.text.actions.saving,
+      description: view.text.aria.disabled_save_description,
+    },
+    preventUnload: true,
+  });
 
   this.$editable = $(".fb-editable");
   this.dataController = dataController;
@@ -128,7 +144,7 @@ PagesController.edit = function() {
 
   dataController.saveRequired(false);
   this.$document.on("SaveRequired", () => dataController.saveRequired(true) );
-
+  this.$document.on("Save", () => this.updateComponents() );
 
   // Bit hacky: Cookies page is going through this controller but content is static.
   // The static content is wrapped in [fb-content-type=static] to help identify it.
@@ -149,17 +165,24 @@ PagesController.create = function() {
 
 
 class DataController {
-  constructor(view) {
-    var $form = $("#editContentForm");
-
-    this.text = view.text;
+  constructor($form, config) {
+    this.config = mergeObjects({
+      text: {
+        clean: '',
+        dirty: '',
+        busy: '',
+      },
+      preventUnload: false,
+    }, config);
+  
+    this.text = config.text;
     this.$form = $form;
     this.$submitButton = this.$form.find(':submit');
     this.$saveDescription = this.$form.find('#save_description');
 
     this.$form.on('submit', (event) => {
       if(this.submitEnabled) {
-        this.update();
+        $(document).trigger('Save');
         this.removeBeforeUnloadListener();
       } else {
         event.preventDefault();
@@ -169,7 +192,7 @@ class DataController {
     this.$submitButton.on('click', (event) => {
       if(this.submitEnabled) {
         this.removeBeforeUnloadListener();
-        $(event.target).prop("value", this.text.actions.saving );
+        $(event.target).prop("value", this.config.text.busy );
       } else {
         event.preventDefault();
       }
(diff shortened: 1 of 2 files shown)


app/javascript/src/controller_pages.js
```diff
@@ -245,12 +199,9 @@ class AddComponent {
 AddComponent.MenuSelection = function(event, data) {
   var action = data.activator.data("action");
   if( action != "none" ) {
-    var submitEnabled = this.dataController.submitEnabled;
-
-    updateHiddenInputOnForm(this.dataController.$form, "page[add_component]", action);
-    this.dataController.submitEnabled = true;
-    this.dataController.$form.submit();
-    this.dataController.suibmitEnabled = submitEnabled;
+    console.log('addCompoennt');
```
stevenburnell-moj, line 202: Dev output left in

app/javascript/src/submit_handler.js
```diff
@@ -0,0 +1,108 @@
...
+
+  set required(required) {
+    if(required) { 
+      this.#enable();
+    } else {
+      this.#disable();
+    }
+  }
+
+  forceSubmit() {
```
stevenburnell-moj, line 45: Is there a reason this is called 'forceSubmit' rather than just 'submit'?
To me, the 'force' bit sounds like it is overriding something, or submitting regardless of any reason it shouldn't. 

Then again, since it looks like I could call `mySubmitHandler.forceSubmit()` directly in code, maybe that's the whole point. 

Should that be an option though? 

Makes me wonder if this introduces the possibility to (accidentally??) bypass the whole `this.required` checking in a potentially dangerous way. Was there a use-case or expectation driving a thought process here?

app/javascript/src/submit_handler.js
```diff
@@ -0,0 +1,108 @@
...
+    const currentState = this.#submitEnabled;
+    this.#submitEnabled = true;
+    this.$form.submit();
+    this.#submitEnabled = currentState;
+  }
+
+  #handleClick(event) {
+    if(this.required) {
+      this.#removeBeforeUnloadListener();
```
stevenburnell-moj, line 54: If a click of the button, when required is true, submits the form and the form has a submit handler that, when required is true, calls this same line, is it still needed here?

app/javascript/src/submit_handler.js
```diff
@@ -0,0 +1,108 @@
...
+      $(event.target).prop("value", this.#config.text.submitting );
```
stevenburnell-moj, line 55: This is looking for the button, right? 
If so, we already have that with `this.$button` so assume we don't need to find it again with a lookup.

Also, just out of curiosity, is there a a different with `this.$button.val('blah')` and `this.$button.prop('value', 'blah')` that made you choose the latter?

One last thing (maybe it's the answer to previous 'curiosity' question)...

Does this work when `this.$button` is either:
a). `<input type="submit" value="I am an input submit" />`
b). `<button type="submit">I am a button submit</button>`

app/javascript/src/submit_handler.js
```diff
@@ -0,0 +1,108 @@
...
+      this.$buttonDescription.text("");
+    }
+    this.#addBeforeUnloadListener();
+  }
+
+  #disable() {
+    this.#submitEnabled = false;
+    this.$button.prop("value", this.#config.text.submitted );
+    this.$button.attr("aria-disabled", true);
+    if(this.$buttonDescription) {
```
stevenburnell-moj, line 101: Not a important but if you prefer, I think if we initialised button description element as empty jQuery collection, 
(e.g. `this.$buttonDescription = $();`) before we checked the config to initialise it properly in the constructor, then it would (should??) be safe to remove the check here and just have the `button.text('blah')` bit. I could be wrong but think jQuery handles it fine, with or without a proper element.

app/javascript/src/submit_handler.js
```diff
@@ -0,0 +1,108 @@
+
+const {
+  mergeObjects
+} = require('./utilities');
+
+class SubmitHandler {
```
stevenburnell-moj, line 6: I think we should add a little description to explain this class. We both know what it's doing but I'm not convinced it would be all that obvious for any future developer, having just scanned through properties and methods. Also, I've seen you do a good job of explaining config options elsewhere, so would benefit from something similar here, before you consider it a final version, of course.

app/javascript/src/submit_handler.js
chrispymm: Yep, if you check the PR description I explained this.  There is absolutely a rerquirement to be able to bypass the `required` functionality.

On a multiQ page, adding a new question or content area, triggers a relaod of the page.  If we don't bypass the `required` functionality then this fails.

So we either need to be able to do somehting like this from outside the class:
```
var tmp = submitHandler.enableSubmit;
submitHandler.enablesubmit = true
submitHandler.$form.submit();
submitHandler.enablesubmit = tmp
```

Which feels very messy, or we need a way for the class itself to be able to do this, which felt much neater... hence `forceSubmit()` or possibly `saveAndSubmit()`?  Although it doesn't save itself, just emits the event... I mean technically all it needs to do is remove the `window.unload` eventListeners and then trigger a submit.

app/javascript/src/submit_handler.js
chrispymm: as to `.prop('value', 'x')` vs `val('x')` there's no difference... not sure there was a reason why I chose one over the other.

As to your second point. That's a good one.  I'll handle that.

app/javascript/src/submit_handler.js
stevenburnell-moj: Ok, I missed that description text (note to self: always re-read the description!)
Also, I didn't notice the relevance of the change where you had...

```
$button.on("click.AddContent", () => {
      var submitEnabled = this.dataController.submitEnabled;
      updateHiddenInputOnForm(config.$form, fieldname, "content");
      this.dataController.submitEnabled = true;
      this.dataController.$form.submit();
      this.dataController.suibmitEnabled = submitEnabled;
    });
```

and updated it to...

```
$button.on("click.AddContent", () => {
      updateHiddenInputOnForm(config.$form, fieldname, "content");
      this.submitHandler.forceSubmit();
    });
```

I was going to mention in this comment that it need only have been changed to...

```
$button.on("click.AddContent", () => {
      updateHiddenInputOnForm(config.$form, fieldname, "content");
      this.submitHandler.required = true;
      this.submitHandler.submit(); 
    });
```

Since `this.submitHandler = true` already alters the value of private `enableSubmit` and, the resulting page reload would to the reset back to default `false` value. However, I can see now that the getter also activates the onunload handler, which would ruin the current action. Fair enough.

To address the this.submitHandler.forceSubmit() naming issue, I wondered if it could simply be...

```
$button.on("click.AddContent", () => {
      updateHiddenInputOnForm(config.$form, fieldname, "content");
      this.submitHandler.submit(true); 
    });
```

So that you could explicitly pass and optional value true (which presumably would change only the `enableSubmit` directly), thus indicating you knew why you were programatically submitting but then, especially given the use-case here, I'm wondering if either `submitHandler.forceSubmit()` or `submitHandler.submit(true)` (whatever variant) are actually necessary.

Maybe `submitHandler.submit()` is all that is required and the inner functionality actively handling the `enableSubmit` value is not a problem. The expected behaviour is that passing a Form (which finds the button) is enough to attach the required `SubmitHandler` functionality but, by simply providing a `submitHandler.submit()` method it can be assumed that it is safe to use - which it is, right? The `forceSubmit()` name, perhaps unnecessarily just implies there might be a reason to think twice or avoid using.

chrispymm: I had also come to a similar conclusion that `submit()` was probably the best method name.

stevenburnell-moj (approved):

PR #1768, opened 2022-09-22
Re-send email verification link for From Address 

njseeto:
[Trello](https://trello.com/c/uazpEFwF/2923-from-address-re-send-email-verification-link)
### Re-send verification link for FromAddress
When a user is unable to verify their email in time, we need to be able to re-send this email. AWS does not have a 're-send' API endpoint as such we need to first delete and then create the email identity again, this will trigger the AWS verification email.

### Move shared FromAddress methods into a FromAddressObjects module
These methods are now used in two controllers. Move the common methods out into a shared module to dry up the code.

### DIFF
--- a/app/controllers/api/from_address_controller.rb
+++ b/app/controllers/api/from_address_controller.rb
@@ -1,14 +1,10 @@
 module Api
   class FromAddressController < ApiController
+    include FromAddressObjects
+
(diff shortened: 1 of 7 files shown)


brenetic (approved):
🚀

brenetic (approved):
🚀

PR #1769, opened 2022-09-23
From Address Publish mechanism

njseeto:
[Trello](https://trello.com/c/eG608how/2928-publishing-add-send-from-address)

### Remove service email from from email updater
We no longer need the `SERVICE_EMAIL_FROM` service config here, as we will move this into the `PublishServiceCreation` class.
We need to separate the `SERVICE_EMAIL_FROM` out of the email settings page to be able to surface up to three different emails (ie: if the user has different emails published to Test, Live and saved in the `FromAddress` database) on the email settings page.

### Add service configuration for service email from
Now that we have removed the `SERVICE_EMAIL_FROM` from the `EmailUpdater`, we need to create the `SERVICE_EMAIL_FROM` config in the `PublishServiceCreation`. The `PublishServiceCreation` is only ever instantiated when a user clicks publish for a given service.
If the `SERVICE_EMAIL_FROM` config does not exist for a service, we create one with the existing email in the FromAddress record for that environment.

### Add the manual gate back into the deployment pipeline
This is because we are changing the Publishing mechanism, which we would like to test fully before deploying to Live.
Once we are happy with the behaviours in testing, we can move back to continuous deployment to Live.

### DIFF
--- a/app/services/publish_service_creation.rb
+++ b/app/services/publish_service_creation.rb
@@ -40,6 +40,7 @@ class PublishServiceCreation
 
     ActiveRecord::Base.transaction do
       create_publish_service
+      assign_service_email_from
 
       if require_authentication?
         create_service_configurations
@@ -162,5 +163,15 @@ class PublishServiceCreation
       service_id: service_id,
       deployment_environment: deployment_environment
     ).try(:service_csv_output?)
+  end
+
+  def assign_service_email_from
+    create_or_update_configuration(
+      name: 'SERVICE_EMAIL_FROM', value: from_address_email
+    )
+  end
+
+  def from_address_email
+    FromAddress.find_or_initialize_by(service_id: service_id).email_address
   end
 end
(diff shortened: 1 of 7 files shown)


app/services/publish_service_creation.rb
```diff
@@ -163,4 +170,16 @@ def service_csv_output
       deployment_environment: deployment_environment
     ).try(:service_csv_output?)
   end
+
+  def service_email_from
```
brenetic, line 174: where is this being used?

app/services/publish_service_creation.rb
```diff
@@ -75,6 +76,12 @@ def no_service_output?
     send_by_email.blank? || (send_by_email.present? && service_email_output.blank?)
   end
 
+  def assign_service_email_from
```
brenetic, line 79: can probably move it to be private

app/services/publish_service_creation.rb
njseeto: good spot, forgot to remove in the clean up.

brenetic (approved):
🚀

PR #1775, opened 2022-09-26
Resend validation always returns 200 status

njseeto:
### Resend validation always returns 200 status
We have decided to always return a 200 status.
We won't be able to do much when we can not reach AWS except have the user try the link again.
The link comes back on page reload, so in the event a user does not receive a validation email from AWS, they will be able to access the resend link and try again.
Have discussed with product and design who are happy with this behaviour.


brenetic (approved):
🚀

PR #1776, opened 2022-09-26
Reinstate continuous deployment to Live

njseeto:
### Reinstate continuous deployment to Live
Now that we have tested the publishing mechanism, we can remove the manual gate and allow continuous deployment to live.


brenetic (approved):
🚀

PR #1777, opened 2022-09-26
Publish warning messages for From Address

njseeto:
[Trello](https://trello.com/c/PFcpcgAo/2956-from-address-publish-warnings-object)
### Publish warning messages for From Address
For the Publish page to know which warnings are required, it needs to know whether an email has been set and verified, and which messages to show.

We already have an existing `FromAddressPresenter`, this commit refactors the presenter's messages object to allow us to pass the correct message according to which controller is being presented to the user.

### Add to brakeman report
This is for an existing warning that is already being ignored. Have amended the report as we now pass a second argument into the `FromAddressPresenter`.

### DIFF
--- a/app/controllers/publish_controller.rb
+++ b/app/controllers/publish_controller.rb
@@ -1,5 +1,5 @@
 class PublishController < FormController
-  before_action :assign_form_objects, :assign_autocomplete_warning
+  before_action :assign_form_objects, :assign_autocomplete_warning, :assign_from_address_presenter
 
   def index
     @published_dev = published?(service.service_id, 'dev')
@@ -53,6 +53,14 @@ class PublishController < FormController
 
   def assign_autocomplete_warning
     @autocomplete_warning = AutocompleteItemsPresenter.new(service, service_autocomplete_items)
+  end
+
+  def assign_from_address_presenter
+    @from_address_presenter = FromAddressPresenter.new(from_address, :publish)
+  end
+
+  def from_address
+    @from_address ||= FromAddress.find_or_initialize_by(service_id: service.service_id)
   end
 
   def published?(service_id, environment)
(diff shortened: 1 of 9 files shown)


app/services/publish_service_creation.rb
```diff
@@ -76,6 +76,8 @@ def no_service_output?
     send_by_email.blank? || (send_by_email.present? && service_email_output.blank?)
   end
 
+  delegate :verified?, to: :from_address, prefix: true
```
brenetic, line 79: using the `prefix: true` means that it will be called by using `from_address_verified?` instead of just `verified?`. is that intentional?

brenetic (dismissed):
🚀

app/controllers/publish_controller.rb
```diff
@@ -55,6 +55,14 @@ def assign_autocomplete_warning
     @autocomplete_warning = AutocompleteItemsPresenter.new(service, service_autocomplete_items)
   end
 
+  def assign_from_address_presenter
+    @from_address_presenter = FromAddressPresenter.new(from_address)
```
chrispymm, line 59: Could we not provide the `page` into the presenter here?  Then the presetner couls initialize the messages array with the right "page" of messages.  That way the view can just access `@from_address_presenter.message[text]`

This doesn't feel too icky to me, as the controller knows which 'page' it is presenting... so seems reasonabel to pass it in.  Could potentially even change `page` to `controller` / `controller_name` ?

app/services/publish_service_creation.rb
njseeto: yep, we are only doing this in the view where we only need to know if form address verified. I don't mind either way - happy to change it back to a specific method if you have a preference?

app/controllers/publish_controller.rb
njseeto: done!

chrispymm (approved):

PR #1780, opened 2022-09-27
Add from address email to send by email page

njseeto:
[Trello](https://trello.com/c/l3YsIX7S/2929-send-by-email-from-address)

### Add from address email to send by email page
Since this will be the first time users will be able to change their from address email, we will need to surface this email in the Send data by Email page.

We want to allow users to see the from address email that will be used when the form is next published rather than what is currently set as the `servce_email_from` value for the particular environment. This means we want to provide the email address that is in the `FromAddress` table, rather than the `service_email_from` address in the `ServiceConfiguration` table.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -394,6 +394,7 @@ en:
         send_by_email_dev: Send by email on Test
         send_by_email_production: Send by email on Live
         service_email_output: Send email to
+        service_email_from: 'From:'
         service_email_subject: Email subject
         service_email_body: Email text
         service_email_pdf_heading: Heading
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -407,6 +408,11 @@ en:
         csv_hint: A comma-separated values (CSV) lists all the answers in plain text, separated by commas. It can be imported into spreadsheets and other applications.
         csv_sample: See sample CSV
         csv_checkbox: Send a separate email with answers in a CSV
+        from_address:
+          default: You need to set and validate a team email address.
+          pending: You need to validate this email address.
+          pending_link: 'Get validation link'
+          link: Change 'from' address
```
chrispymm, line 415: This should have 'curly' quotes not straight ones.  If you have a look elswhere in the yml file you'll find some you can copy.

chrispymm (changes requested):

config/locales/en.yml
njseeto: done

chrispymm (approved):
🚀

PR #1782, opened 2022-09-28
2792 JS tests for ConnectionMenu

stevenburnell-moj:


test/menus/connection_menu/events_test.js
```diff
@@ -0,0 +1,116 @@
...
+    });
+
+
+    /* TEST EVENT: menuselect
+     **/
+    describe("menuselect", function() {
+
+      it("should not call the selection method when menuselect is triggered with action 'none'", function() {
+        var called = testSelectionCalled(created.item, "li[data-action=" + c.TEXT_ACTION_NONE + "]");
+        //expect(called).to.be.false;
```
chrispymm, line 67: Should this be commented?

chrispymm (approved):

test/menus/connection_menu/events_test.js
stevenburnell-moj: Well spotted. An earlier effort had me changing code to match that test, which resulted in a positive (ie. expect false) result. However, I back-tracked on that change (for whatever reason I no longer remember) and clearly didn't reverse the mid-development-comment-out effort. 

It's a little pointless since 'none' doesn't do anything inside `selection()` but I'll alter the test to match what is currently does, e.g.
```
      it("should call the selection method when menuselect is triggered with action 'none'", function() {
        var called = testSelectionCalled(created.item, "li[data-action=" + c.TEXT_ACTION_NONE + "]");
        expect(called).to.be.true;
      });
```

chrispymm (approved):

PR #1788, opened 2022-10-03
Fix issues with save button tests

chrispymm:
Original deploy for this had issues where it was expecting dynaimic save button behaviour on the branching page.

Updated the tests so the dynamic behaviour is only tested on the eidt question pages.


brenetic (approved):
🚀

PR #1789, opened 2022-10-03
Add Controller to add Expander to send by email settings page

chrispymm:
Adds the expander component to the `Send data by email` settings page.

This adds the functionlality to open the following configuration block when the 'Send by email on test/live' checkbox is checked (if it is not already open).

If the page is saved with errors then the configuration block with errors in will open automatically on page load, allowing the user to see and resolve the issues.


brenetic (approved):
🚀

PR #1793, opened 2022-10-05
Publish page warnings

njseeto:
### Add the initial Publishing Page Presenter
Create a Publishing Page Presenter to encapsulate all the different possible warnings that we need to show our users.

We now pass in an array of potential messages for the from address and submission pages presenters. Depending on the 'state' at any given moment the presenters will decide whether or not show a message.

Renamed the presenters better describe what they do. There are now:
- Publishing Page Presenter - orchestrates all presenters
- Submission Pages Presenter - only deals with the Check Answers page and Confirmation page
- From Address presenter - only deals with the From Address and the potential states it can be in
- Autocomplete Items Presenter - relates to any autocomplete components on pages and any issues they may have

### Front End changes for publish warnings
- Refactor publish index template to be environment agnostic
- Fix email settings headings
- Update locale properties

### Refactor Publishing Page presenters
Use keyword arguments.
Add additional test for when the Publish button is shown or not

### Correctly disable the Publish button
We always allow Publishing to the Test environment.
We disable the Publishing button in Live when:
- service email output is disabled
- there is no service email output saved
- there are submission warnings (from address or submission pages)
- there are autocomplete warnings

### Check whether there are any submission warnings
If there are no warnings then we do not want to show the alert icon.

### Update Publishing Page acceptance tests
We have heavily changed the behaviour of the Publishing page. Therefore, we have decided to re-write the acceptance tests for this page. We now have a catch all warning based on whether the user has their 'Send by Email' enabled.

If the 'Send by Email' has been enabled, we show an overall warning message, which is followed by more specific errors that relate to the submission pages, from address and autocomplete errors.


chrispymm (approved):
🚀

chrispymm (approved):

PR #1800, opened 2022-10-07
Ensure AWS calls are made only if the email is valid

njseeto:
[Trello](https://trello.com/c/C2nWHT9z/2984-from-address-bug-aws-verification-email-is-sent-when-errors-are-triggered)

### Ensure AWS calls are made only if the email is valid
We do not want to call AWS with an invalid email address. This commit passes the `from_address_params` into the `from_address` initialiser which allows us to check whether the email submitted is valid before we attempt to make the AWS requests.

### Remove unnecessary test
This test looks at the scenario when the AWS console shows the email as not verified, but in our database the email is verified.
This can never happen as we cannot change the verification status to be unverified, ie: change the sending_enabled status.
We only have the power to delete or create email identities. Therefore, we can remove this test.

### DIFF
--- a/app/models/from_address_objects.rb
+++ b/app/models/from_address_objects.rb
@@ -2,7 +2,6 @@ module FromAddressObjects
   def from_address_creation
     @from_address_creation ||= FromAddressCreation.new(
       from_address: @from_address,
-      from_address_params: from_address_params,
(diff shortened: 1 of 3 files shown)


brenetic (approved):
🚀

PR #1801, opened 2022-10-07
Cleanup test service configurations

brenetic:
Add a rake task which removes all the service configurations, publish services, from addresses and submission settings for services that are not owned by members of the MoJ Forms teams.

For the most part this would be just service configurations generated by acceptance tests when they run on deployment. It can also include those generated by old members of the team who have now left.


njseeto (approved):
🚀

PR #1804, opened 2022-10-11
Allow list message for From Address email domains

njseeto:
[Trello](https://trello.com/c/DjU0jKGX/2996-from-address-warning-to-add-domain-to-allow-list)

### Show warning message if email domain is not in allow list
Currently, we validate email domains - if the domain is not in the allow list then we do not save and we provide an error message.

We now want to allow users to save their emails regardless of whether the domain is on the allow list, but provide an opportunity for users to request their domain be added to the allow list.

This commit removes the domain validation and provides template changes for the two journeys, email domain on the allow list and email domains not on the allow list.

### Cater for 404 in AWS `delete_email_identity` call
Since we now allow users to save their email without creating an email identity in AWS, ie: when a user enters an email with a domain that is not on the allow list, we need to cater for instances where we call `delete_email_identity`and the email_identity is not found.

Screenshot 2022-10-11 at 15 02 36


spec/services/from_address_creation_spec.rb
```diff
@@ -59,6 +59,28 @@
       end
     end
 
+    context 'with email not on allow list' do
+      let(:email) { 'galadriel@valinor.me' }
```
brenetic, line 63: have you been watching The Rings of Power? 👀

app/validators/email_domain_validator.rb
```diff
@@ -1,20 +1,12 @@
-class EmailDomainValidator < ActiveModel::Validator
+class EmailDomainValidator
```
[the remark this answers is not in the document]
njseeto: All done, added this change to a separate commit.

brenetic (changes requested):

app/views/settings/from_address/index.html.erb
```diff
@@ -9,19 +9,28 @@
...
+                  remote: true,
+                  method: 'post',
+                  url: api_service_settings_from_address_resend_path(service.service_id),
+                  success: t('settings.from_address.link_resent_success')
+                } %>
+            </p>
+        <% else %>
+          <%= t('warnings.from_address.settings.contact_us').html_safe %>
+            <p>
+              <span role="alert"></span>
```
chrispymm, line 28: I need to check why this empty `<span>` is here. It has an alert role, which indicates we're using JS to inject content we want announced by screenreaders... but I'm not sure that's true anymore.

chrispymm: Just checked further.  This empty span is defintily redundant in the 'contact us' message.

In the block above, the 'resend email' link is done via ajax, and on success we udate the hidden span with the success message the `role=alert` means it gets read out by screenreaders.

njseeto: Ahh I see, have removed the redundant `span` in the contact us message.

app/services/from_address_creation.rb
```diff
@@ -18,28 +18,30 @@ def verify_email
...
+      # Although we save all email addresses with the status as 'pending',
+      # we only want to call AWS if the domain is on the allow list
+      email_service.create_email_identity(from_address.email) if allowed_email?
 
       Rails.logger.info("Created email identity for service #{from_address.service_id}")
       :pending
 
-    elsif email_identity.sending_enabled && !from_address.verified?
+    else
+      email_identity.sending_enabled && !from_address.verified?
```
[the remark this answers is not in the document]
njseeto: ah yes silly mistake, have fixed now

chrispymm (approved):

app/models/from_address.rb
```diff
@@ -33,6 +36,13 @@ def default_email?
     email_address == DEFAULT_EMAIL_FROM
   end
 
+  def allowed_domain?(email)
```
brenetic, line 39: sorry should have brought this up yesterday. the `from_address` object already has the email as an accessor therefore it doesn't need to be passed into this method, especially as the `from_address_creation` has access to the specific `from_address` object containing the email.

Therefore this can be a method that acts on the attribute inside the `from_address` object and then you can just delegate the `allowed_domain?` method to the `from_address` object in the `from_address_creation` model.

app/services/from_address_creation.rb
```diff
@@ -68,4 +69,8 @@ def use_default_email?
     from_address.email.blank? ||
       from_address.email == FromAddress::DEFAULT_EMAIL_FROM
   end
+
+  def allowed_email?
```
brenetic, line 73: `delegate :allowed_domain? to: :from_address`

as mentioned above. then this method can come out

njseeto: all done

brenetic (approved):
🚀

PR #1805, opened 2022-10-11
Unpublish services in the Test environment

brenetic:
This adds a rake task that will unpublish services in the test-dev and test-production namespaces every Friday at 20:00.

This is simply to stop randome services just sitting there using up resources in our Test environments. Users can republish any time they like.

Move some of the helper methods used by the Admin Services Controller into a Helper module.

### DIFF
--- a/deploy-eks/fb-editor-chart/templates/unpublish_test_services.yaml
+++ b/deploy-eks/fb-editor-chart/templates/unpublish_test_services.yaml
@@ -49,16 +49,6 @@ spec:
                   secretKeyRef:
                     name: {{ .Values.bearer_token }}
                     key: token
-              - name: AWS_SES_ACCESS_KEY_ID
-                valueFrom:
-                  secretKeyRef:
-                    name: fb-editor-secrets-{{ .Values.environmentName }}
-                    key: aws_ses_access_key_id
-              - name: AWS_SES_SECRET_ACCESS_KEY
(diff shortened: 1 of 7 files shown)


deploy-eks/fb-editor-chart/templates/unpublish_test_services.yaml
```diff
@@ -0,0 +1,73 @@
...
+                valueFrom:
+                  secretKeyRef:
+                    name: rds-instance-formbuilder-editor-{{ .Values.environmentName }}
+                    key: url
+              - name: EDITOR_SERVICE_ACCOUNT_TOKEN
+                valueFrom:
+                  secretKeyRef:
+                    name: {{ .Values.bearer_token }}
+                    key: token
+              - name: AWS_SES_ACCESS_KEY_ID
```
njseeto, line 52: Do we need the AWS secrets for this cronjob?

njseeto (changes requested):

deploy-eks/fb-editor-chart/templates/unpublish_test_services.yaml
brenetic: good spot! i have removed them, and also taken them out of the test service configs rake task as well.

njseeto (approved):
🚀

PR #1807, opened 2022-10-12
From Address acceptance tests

njseeto:
Acceptance tests for the From Address settings page and Send data by email page. For both we check:
- The presence of the default email and the correct warning
- When an email is not on the allow list and the correct warning

We do not test for when an email is valid as we do not want to make a unnecessary calls to AWS nor do we want to pollute the AWS console with a large number of emails.


brenetic (approved):
🚀

PR #1808, opened 2022-10-12
Change the `service_email_from` default

njseeto:
[Trello](https://trello.com/c/JvwtxVit/2954-from-address-set-default-email-address-to-no-reply-moj-formsdigitaljusticegovuk)

As part of the From Address work, we want to ensure the default email we use clearly states there will be no monitoring of the email inbox.


brenetic (approved):
🚀

PR #1810, opened 2022-10-12
Remove cleanup of service configuration from acceptance tests

brenetic:
We now have the `remove:test_services_configs` rake task that takes care of doing this so we do not need to do it during the test runs.


mattempty (dismissed):
🚀

njseeto (approved):
🚀

PR #1814, opened 2022-10-14
Rake task to update default from address email

njseeto:
With the release of From Address, old forms will have the old 'moj-forms' default email. We want to update these to use the new 'no-reply-moj-forms' email. This task will be run manually as it will only need to be done once.

The email is saved as an empty string in the rake task as the `FromAddress` model allows this and will update and encrypt records accordingly.

### DIFF
--- a/lib/tasks/from_address.rake
+++ b/lib/tasks/from_address.rake
@@ -5,4 +5,17 @@ namespace 'from_address' do
   rescue StandardError => e
     Sentry.capture_exception(e)
   end
+
+  desc "Update default email to 'no-reply-moj-forms'"
+  task update_email: [:environment, 'db:load_config'] do
+    FromAddress.all.each do |record|
+      if record.email_address == 'moj-forms@digital.justice.gov.uk'
+        record.email = ''
+        record.save!
+      end
+    end
+
+  rescue StandardError => e
+    Sentry.capture_exception(e)
+  end
 end


lib/tasks/from_address.rake
```diff
@@ -5,4 +5,17 @@ namespace 'from_address' do
   rescue StandardError => e
     Sentry.capture_exception(e)
   end
+
+  desc "Update default email to 'no-reply-moj-forms'"
+  task update_email: [:environment, 'db:load_config'] do
+    FromAddress.all.each do |record|
+      if record.email_address == 'moj-forms@digital.justice.gov.uk'
+        record.email = ''
+        record.save!
```
brenetic, line 14: what if we destroyed the record instead? am i right in thinking that coz we do `find_or_initialize_by` that it would be fine?

or perhaps just set the email to be the new email address?

njseeto: I hadn't thought about destroying the record so have just tested this. It looks like a record is created when a user clicks the 'save' button - running `FromAddress.count` gives me 3 records but I have over 10 forms. Is this expected? I think if a user has created/saved a from address I think we should update rather than delete.

I also tried setting the email to be the new email address in the rake task, but the record would not save - I think it is because the updated email was not encrypted? Saving as blank seemed to be a way around that.

brenetic: saving an empty email is probably not want we want? it should never be empty coz it should at the very least save the default email address to the db.

catering for no from address record means that all existing services with no from address would continue to work as expected. similarly for all new services a from address record is created by the new service generator. 

currently the controller doesnt mind if there is no from address record; just make one and show the user whatever is the default email address.

also the publishing mechanism should t care if there is no from address for a service as it will just make one and then use the default email address.

the rake task is checking specifically for the old email so will only remove the record if its the old email. if the user has changed it then it will ignore the record.

njseeto: yep agreed and understand the rake task is only checking for the old email. Saving an empty email, will trigger the model before save to check, replace and encrypt with the default email - it won't ever empty, at least that's what I've seen when testing with the local database. But happy to change it to a delete and leverage the `find_or_initialize by` mechanism.

brenetic (approved):
🚀

PR #1816, opened 2022-10-14
Send by email from address formatting

njseeto:
Small change to format link when email address is verified.

### Before
Screenshot 2022-10-14 at 12 33 56

### After
Screenshot 2022-10-14 at 12 35 34


chrispymm (approved):
🚀

PR #1817, opened 2022-10-14
From address remove feature flag

njseeto:
We currently have the from address feature behind a feature flag.
This PR removes the flag for when we are ready to go live.


H3ll3m4 (approved):
🚀

PR #1818, opened 2022-10-14
Add private key to cronjob configs

brenetic:
Both the removal of test services configs and the unpublishing of test services require the private key in order to sign the JWT token when sending requests to the Metadata API.


njseeto (approved):
🚀

PR #1820, opened 2022-10-17
Undo Redo for Move page and Change next page

H3ll3m4:
[Trello](https://trello.com/c/rMBqTlnQ/2886-undo-redo-rails-ootb)

## Context
We want to provide the user the possibility to undo some actions. It is possible to get the previous version of metadata by calling the previous methods on the metadata api endpoint. This allow to get the the previous version of the form.

## Scope
Scope is limited to two user actions Move page and Change Next Page.

## Implementation
Move page is handled by Move controller and Change next page by Destinations Controller. Once the user choose to undo or redo an action, this action is passed as params to the Undo Controller. The Undo Controller handle the calls to the previous version from the Metadata API Client. It also store current undo/redo action information in the session. Then the Service Controller pass this information into a button for the view to display. 

## Tests
Adding unit tests, integration tests as request spec and acceptance tests.


app/controllers/api/move_controller.rb
```diff
@@ -8,6 +8,15 @@ def targets
...
       @move.change
+
+      if @move.valid?
+        session[:undo] = {
+          action: 'undo',
+          undoable_action: 'move',
+          text: t('actions.undo_redo.undo_move')
+        }
+      end
+      # TODO: do we want to add here in this path toos
```
brenetic, line 19: do we need this?

app/controllers/api/destinations_controller.rb
```diff
@@ -8,6 +8,11 @@ def new
 
     def create
       destination.change
+      session[:undo] = {
+        action: 'undo',
+        undoable_action: 'change_next_page',
```
brenetic, line 13: i read `undoable` as something that cannot be undone 😆

app/controllers/api/move_controller.rb
brenetic, line 19: do we need this comment?

app/controllers/api/undo_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+          text: t('actions.undo_redo.undo_change_next_page')
+        }
+      when 'move'
+        session[:undo] = {
+          action: 'undo',
+          undoable_action: 'move',
+          text: t('actions.undo_redo.undo_move')
+        }
+      end
+      previous
```
brenetic, line 18: can we name this something a little more descriptive please

brenetic:
Coolio, couple of comments. 

Seeing all the logic in the controller now I realise that that's a fair amount of logic. How would you feel about breaking it out into a class? For use in both the `undo` and `undo_redo` methods. If so that could mean a reduction in request specs and just have unit tests instead.

Also please do squash those commits down. They probably don't need to be that granular.

Other than that, great stuff! 👍

app/controllers/services_controller.rb
```diff
@@ -17,6 +17,8 @@ def create
   end
 
   def edit
+    @undo_redo_button = session.delete(:undo)
```
brenetic, line 20:
```suggestion
    if ENV['undo_redo'] == 'enabled'
      @undo_redo_button = session.delete(:undo)
    end
```

assuming you don't want to launch the feature without the error messages?

H3ll3m4: Added feature flag

app/controllers/api/undo_controller.rb
H3ll3m4: Changed to call_previous_version

app/controllers/api/move_controller.rb
H3ll3m4: comments removed

app/controllers/api/destinations_controller.rb
H3ll3m4: will keep until we find a better name

app/controllers/api/move_controller.rb
H3ll3m4: comments removed

brenetic (approved):
🚀

PR #1821, opened 2022-10-18
Remove From address update email rake task

njseeto:
This rake task was used to update any forms that were created before From Address was released. Now that from address has been released and all forms have been been updated with the new default email address, we no longer require this rake task.


H3ll3m4 (approved):
🚀

brenetic (approved):
🚀

PR #1824, opened 2022-10-20
Confirmation Email

njseeto:
### Add confirmation email feature flags and settings page link
For now, add feature flag at the confirmation email entry point as we don't want to release this feature into production.

### Add email templates and `ConfirmationEmailSettings` class
The email template will need to be able to iterate through the email component ids to allow the form editor to pick
which email component will be the one to use as the confirmation email.

`ConfirmationEmailSettings` class will allow us to get the correct values that are required for the confirmation email feature. These values are primarily gathered via a form and can be adjusted per deployment environment.

Screenshot 2022-10-20 at 15 54 10

### Create common parent class for base email settings
Now that we are creating a similar email template for the confirmation email settings, there are a few methods that are
shared between the `ConfirmationEmailSettings` class and the `EmailSettings` class.
Create a `BaseEmailSettings` class with the common methods to reduce repetitive code.

### DB migration for `send_confirmation_email` to submission settings
Similar to the `send_email` column, we need to have a similar mechanism for the confirmation email. The `send_confirmation_email` value will be a boolean and signify whether the form owner wants to enable confirmation for a particular environment on a particular service.

### Add `ConfirmationEmailSettingsUpdater` and service configurations
The `ConfirmationEmailSettingsUpdater` is similar to the `EmailSettingsUpdater` in that it updates the associated
Service configurations record for a specific service in a specific deployment environment. In this case, we are looking at the configurations for the confirmation email feature.

### Add warning for when there are no email components
In the event that there are no email components on a form, the confirmation email cannot be set. We therefore, set a warning to prompt the user to add an email component and to refer to the user guide.

Screenshot 2022-10-20 at 15 53 32


config/locales/en.yml
```diff
@@ -309,6 +311,12 @@ en:
         production:
           config_details_summary: Configure <span class="sr-only">live settings</span>
           save_button: Save Live settings
+    confirmation_email:
+      heading: 'Send a confirmation email'
+      lede: Send the people who complete your form a copy of their data
+      email:
+        label: Send a confirmation email
+        help: Send the people completing your form a confirmation email with their information attached in PDF.
```
brenetic, line 319:
```suggestion
        help: Send the people completing your form a confirmation email with their information attached in a PDF.
```
?

brenetic (approved):
🚀

app/controllers/settings/confirmation_email_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def from_address
+    @from_address ||= FromAddress.find_or_initialize_by(service_id: service.service_id)
+  end
+
+  def email_components
+    @email_components ||=
+      components = service.pages.map(&:components).flatten
+    components.select { |component| component.type == 'email' }
```
brenetic, line 79:
```suggestion
      components.select { |component| component.type == 'email' }
```

brenetic: i am not sure whether this would need to be wrapped in begin end.

otherwise if you wanna single line is:

```
service.pages.map(&:components).flatten
            .select { |component| component.type == 'email' }
```

brenetic: weird spacing aside 🙈

config/locales/en.yml
njseeto: good spot 🦅 👁️

app/controllers/settings/confirmation_email_controller.rb
njseeto: done

brenetic (approved):
🚀

PR #1826, opened 2022-10-24
Confirmation email blocking modals

njseeto:
[Trello](https://trello.com/c/M2KMc4f4/3024-confirmation-email-05-block-removing-component-page-if-used)

### Add delete page modal for confirmation email
We need to prevent users from deleting any pages that have an email component that is used as the confirmation email.

Screenshot 2022-10-24 at 15 27 22

### Add delete modal for email questions
Similar to blocking on a page level, we want to prevent users from deleting the specific email component question on a multiple question page.

Since this modal triggers on a component level rather than a page level, users will be able to delete other components on their multiple question page.

Screenshot 2022-10-24 at 15 27 33


brenetic (approved):
🚀

PR #1829, opened 2022-10-25
Extract undo logic

H3ll3m4:
## Context
Undo controller was found to hold a lot of logic for a controller and a refactoring was needed.

## Implementation
It was decided to extract this logic by using an Undo Presenter class. This class is responsible for providing the text label for each case undo/redo/move/change next page. It is also responsible for changing status undo/redo with a toggle method. 

## Tests
Unit tests for the class added. Manually tested and passing undo-redo request spec integrations tests and acceptance tests.

### DIFF
--- a/app/presenters/undo_presenter.rb
+++ b/app/presenters/undo_presenter.rb
@@ -1,0 +1,37 @@
+class UndoPresenter
+  attr_accessor :action, :undoable_action
+  attr_reader :text
+
+  def initialize(action:, undoable_action:, text: '')
+    @action = action
+    @undoable_action = undoable_action
+    @text = self.class.provide_text_undo(action, undoable_action)
+  end
+
+  def self.provide_text_undo(action, undoable_action)
+    case action
+    when 'undo'
+      case undoable_action
+      when 'change_next_page'
+        I18n.t('actions.undo_redo.undo_change_next_page')
+      when 'move'
+        I18n.t('actions.undo_redo.undo_move')
+      end
+    when 'redo'
+      case undoable_action
+      when 'change_next_page'
+        I18n.t('actions.undo_redo.redo_change_next_page')
+      when 'move'
+        I18n.t('actions.undo_redo.redo_move')
+      end
+    end
+  end
+
+  def undo_presenter
+    @undo_presenter ||= UndoPresenter.new(
+      action: action,
+      undoable_action: undoable_action,
+      text: self.class.provide_text_undo(action, undoable_action)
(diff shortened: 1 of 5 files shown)


app/presenters/undo_presenter.rb
```diff
@@ -0,0 +1,67 @@
...
+  # rubocop:disable Lint/UnusedMethodArgument
+  def initialize(action:, undoable_action:, text: '')
+    @action = action
+    @undoable_action = undoable_action
+    @text = self.class.provide_text_undo(action, undoable_action)
+  end
+  # rubocop:enable Lint/UnusedMethodArgument
+
+  def toggled_presenter
+    if @text == I18n.t('actions.undo_redo.undo_move')
```
brenetic, line 14: `attr_reader :text` on line 3 automagically creates a getter so you don't need to use `@`, you can just call `text`

brenetic:
okey dokey, i haven't thought this through massively, so perhaps this isn't the right approach, but in terms of DRY'ing it a little you could have the messages ready in a constant and then just use the `action` and `undoable_action` as property keys to retrieve the text. Instantiating a new Presenter object with the `action` and `undoable_action` should be enough so just access them directly at set them in the hash returned, kinda removing the need for any case statements entirely:

```
UNDO_REDO_TEXT = {
  undo: {
    change_next_page: I18n.t('actions.undo_redo.undo_change_next_page'),
    move: I18n.t('actions.undo_redo.undo_move')
  },
  redo: {
    change_next_page: I18n.t('actions.undo_redo.redo_change_next_page'),
    move: I18n.t('actions.undo_redo.redo_move')
  }
}.freeze

# not sure about this method name :D
def toggled_presenter
  {
    action: action,
    undoable_action: undoable_action,
    text: UNDO_REDO_TEXT[action][undoable_action]
  }
end
```

OR, perhaps less safe, interpolating the action and `action` and `undoable_action` into the call to `I18n`, removing the need for any type of content constant:

```
# still not sure about the method name :P
def toggled_presenter
  {
    action: action,
    undoable_action: undoable_action,
    text: I18n.t("actions.undo_redo.#{action}_#{undoable_action}")
  }
end
```

My only other thought as that do we need to have both an instance method and a class method? It could be ok in the controllers that use the class method to return the whole hash object as well, instead of just the text.

brenetic (dismissed):
🚀

njseeto (dismissed):
🚀

brenetic (approved):
🚀

PR #1833, opened 2022-10-26
Fix publish warnign bug

chrispymm:
Fixes the issue where the 'Publish to Live' button is disabled when the 'Send Email' setting is not checked.

The message displayed is only a warning message, as form owners may intend a form that doesn not submit data.

**Before**
image

**After**
image


njseeto (approved):
🚀

PR #1835, opened 2022-10-26
Fix tests for updated publish live button behaviour

chrispymm:
Updates the tests to pass now that the publish button is enabled when no 'send by email' is configured in settings.


brenetic (approved):
🚀

PR #1838, opened 2022-10-28
Update form validation to reject name starting with numerical characters

H3ll3m4:
## Context 
We had an issue recently where Kubernetes could not deploy a form that begins with a numeric character because it has to respect the DNS label standard as defined in RFC 1035. Solution was to update form name validation.

## Implementation
Update service validation to reject name starting with numerical characters. Currently message is default and not specific : "Your answer for ‘Give your form a name' contains characters that are not allowed."

## Tests
Unit tests and local testing performed.

### DIFF
--- a/app/services/editor/service.rb
+++ b/app/services/editor/service.rb
@@ -5,7 +5,7 @@ module Editor
 
     validates :service_name, presence: true
     validates :service_name, length: { minimum: 3, maximum: 128 }, allow_blank: true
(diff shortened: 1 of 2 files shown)


brenetic:
will this still allow numbers elsewhere in the service name?

H3ll3m4:
> 

Thanks for your comment. It works well. Thanks

brenetic (approved):
🚀

PR #1840, opened 2022-10-28
Update modal focus() to allow for links or buttons.

chrispymm:
The confirmation email question deletion modal has its confirmation button as a link to the settings page, so this change allows that button to correctly gain focus on opening the modal.

Also includes a small change to the wording in the modal, and a change to the wording in a similar branching modal for consistency.

N.B. Sorry for all the trailing whitespace noise.  Now have my editor set to remove it on save, so when editing an older file from when I didn't it all gets removed! 

**Before**
image

**After**
image


njseeto (approved):
🚀

PR #1841, opened 2022-10-28
Hint form name

H3ll3m4:
## Context
Following PR #1838 solving the issue of service names to avoid deployment issues for Kubernetes ingress controllers names having to follow DNS labels standard defined in RFC1035. We raised the error for starting a form name with numbers is default and not specific. To follow up on that, we want to suggest to the user with a hint to the prohibition to use numeric characters.

## Implementation 
Adding hint with "aria-describedby". It is placed between the label name and the text field. Adding also a class "govuk-hint” to display in grey.

## Tests
Manual tests were performed and hint is displayed as this:
Screenshot 2022-10-28 at 14 25 48

### DIFF
--- a/app/views/services/index.html.erb
+++ b/app/views/services/index.html.erb
@@ -27,15 +27,14 @@
            data-component="FormCreateDialog">
         <%= form_for(@service_creation, url: services_path) do |f| %>
           <div class="govuk-form-group <%= f.object.errors.present? ? 'govuk-form-group--error' :'' %>">
-            <div id="hint">Give your form a name, your form name cannot begin with a number</div>
             <%= f.label :service_name, class: "govuk-label govuk-label--m" %>
             <% f.object.errors.each do |error|  %>
               <p class="govuk-error-message"><%= error.message %></p>
               <div id="hint">Give your form a name, your form name cannot begin with a number</div>
             <% end %>
+            <div class="govuk-hint" id="hint">Give your form a name, your form name cannot begin with a number</div>
             <%= f.text_field :service_name, class: "govuk-input", 'aria-describedby' => 'hint'%>
           </div>
-
           <%= f.button t('services.create'), class: "govuk-button fb-govuk-button", type: 'submit' %>
           <%# f.button t('actions.cancel'), class: "govuk-button govuk-button--secondary", type: 'button' %>
         <% end %>
--- a/app/views/services/index.html.erb
+++ b/app/views/services/index.html.erb
@@ -35,7 +35,6 @@
             <%= f.text_field :service_name, class: "govuk-input", 'aria-describedby' => 'service-name-hint'%>
           </div>
           <%= f.button t('services.create'), class: "govuk-button fb-govuk-button", type: 'submit' %>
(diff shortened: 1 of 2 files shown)


app/views/services/index.html.erb
```diff
@@ -30,10 +30,11 @@
             <%= f.label :service_name, class: "govuk-label govuk-label--m" %>
             <% f.object.errors.each do |error|  %>
               <p class="govuk-error-message"><%= error.message %></p>
+              <div id="hint">Give your form a name, your form name cannot begin with a number</div>
```
chrispymm, line 33:
```suggestion
```
The hint doesn't need to be duplicated within the errors block.

app/views/services/index.html.erb
```diff
@@ -30,10 +30,11 @@
...
             <% end %>
-            <%= f.text_field :service_name, class: "govuk-input" %>
+            <div class="govuk-hint" id="hint">Give your form a name, your form name cannot begin with a number</div>
+            <%= f.text_field :service_name, class: "govuk-input", 'aria-describedby' => 'hint'%>
```
chrispymm, line 36: While it's not necessarily wrong, I would use a more specific `id` than 'hint'.  `id` attributes must be unique to the whole page. Not actually an issue on this page, but its best to be a bit more specific to avoid conflicts.  Usually combining the field name and 'hint' is enough e.g. 'service-name-hint'

The actual text content of the hint should be moved to the translations file (config/locales/en.yml).

app/views/services/index.html.erb
```diff
@@ -30,10 +30,11 @@
...
           </div>
-
           <%= f.button t('services.create'), class: "govuk-button fb-govuk-button", type: 'submit' %>
           <%# f.button t('actions.cancel'), class: "govuk-button govuk-button--secondary", type: 'button' %>
```
chrispymm, line 39: While you're here, you might as weel remove this commented line.
```suggestion
```

chrispymm (changes requested):

app/views/services/index.html.erb
```diff
@@ -31,11 +31,10 @@
             <% f.object.errors.each do |error|  %>
               <p class="govuk-error-message"><%= error.message %></p>
             <% end %>
-            <%= f.text_field :service_name, class: "govuk-input" %>
+            <div class="govuk-hint" id="service-name-hint"> <%= t('services.form_name_hint')%></div>
+            <%= f.text_field :service_name, class: "govuk-input", 'aria-describedby' => 'service-name-hint'%>
```
chrispymm, line 35: I'll admit to not being sure on this one - I'm guessing this syntax works, but I personally would tend to avoid mixing the short and arrow syntax in a hash of options.

I would use either `'aria-describedby': 'service-name-hint'` or you can use a nested hash `aria: { describedby: 'service-name-hint' }`  the nesting will join  the keys together with a `-` in the attibute name in the html.  

The latter version is more useful if you have multiple `aria` attributes or `data` attributes, but it results in a cleaner, more ruby-like syntax.

Not a blocker for approval though

chrispymm (approved):

PR #1842, opened 2022-10-28
Fix confirmation page blocking modals

njseeto:
[Trello](https://trello.com/c/R1YDi5hs/3032-bug-confirmation-email-delete-modals-appear-when-form-is-not-checked)

Previously, we were only checking whether a `CONFIRMATION_EMAIL_COMPONENT_ID` had been set in either 'dev' or 'production'. This is not enough as we can have these set but then have the `send_confirmation_email` config SubmissionSetting as 'false'.

This commit rectifies this issue and checks for:
- Whether the `send_confirmation_email` is true in either environment
- If so, whether the `CONFIRMATION_EMAIL_COMPONENT_ID` is the same as the component_id of the page/question being deleted.

Unfortunately, these configs need to be checked for both 'dev' and 'production' environments.


brenetic (approved):
🚀

PR #1843, opened 2022-10-28
Fix confirmation email settings form labels and IDs

chrispymm:
- Ensure all form labels match the field ids for accessibility. 
- Move helpers from Settings::EmailHelper -> SettingsHelper so they can be used with multiple controllers 
- Rename helper method from email_settings_field_id -> settings_field_id to be more generic - doesn't have to be used just for email 
- Update translation references for 'From:' labelling.


njseeto (approved):
🚀 Thanks for cleaning up the templates 🙏

PR #1844, opened 2022-10-28
Disable Live publishing

brenetic:
This needs to happen in order to stop any forms being published to Live during the database migration


chrispymm (dismissed):
All good! (except the tests) But I guess you might force it thru as its just for the maintenance period?

chrispymm (approved):

PR #1846, opened 2022-10-28
Fix position with branching diamond page menu

chrispymm:
Trello: https://trello.com/c/G97XXlKj/3041-bug-branching-point-page-menu-position

The branching point menu activator had deserted its post and gone AWOL
This very small change brings it back into line!

**Before**
image

**After**
image


brenetic (approved):
🚀

PR #1853, opened 2022-10-31
Removing feature flag for undo-redo

H3ll3m4:
In order to have undo redo available to users it has to go in live environment. We achieve this by removing feature flag in 3 locations, in the app code in the views, in config maps and in the environment variables file. It has been tested in local and deploy OK in CircleCI.


brenetic (approved):
🚀

PR #1854, opened 2022-10-31
Updating form name hint

H3ll3m4:
Form name hint has been proposed to be : It cannot start with a number or include special characters except ’ - ( ) and keeping it straight apostrophe in code.  Rationalising the use of colon / arrow in the hash in the view. Updating colon rather than arrow, as per comments in #1841. Tested locally.


brenetic (approved):
🚀

PR #1855, opened 2022-10-31
Add Sentry logging to frontend

chrispymm:
Ticket: https://trello.com/c/zO2GeDb2/3042-flow-view-blank-screen-property-and-affairs-upfront-notification-application-v2

**Adding Sentry**
So in addition to the above fix, this PR adds Sentry logging into the frontend of the app.  This was not as straightforward as itcould be `cos Webpack.

We want a solution that:
* Makes Sentry logging available globally in any JS file 
* Only logs in test & production environments, not dev
* Logs to the correct Sentry project `fb-editor-live` or `fb-editor-test` respectively.

This is achieved using `webpack.definePlugin` to provide the `NODE_ENV` and appropriate `SENTRY_DSN` and then using `webpack.providePlugin` to provide a `SentryLogger` class globally to the app.

the `SentryLogger` class is a simple wrapper around the `@sentry/browser` module.  It defines a `send` method, which will send the provided details to sentry if not in the development environment.

### DIFF
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -1,10 +1,19 @@
 const { environment } = require('@rails/webpacker')
 const webpack = require('webpack')
+const path = require("path");
 
 environment.plugins.prepend('Provide',
   new webpack.ProvidePlugin({
     $: 'jquery',
-    jQuery: 'jquery'
+    jQuery: 'jquery',
+    SentryLogger: 'sentry-logger',
+  })
+);
+
+environment.plugins.prepend('env',
+  new webpack.DefinePlugin({
+    'NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+    'SENTRY_DSN': (JSON.stringify(process.env.NODE_ENV) == 'production' ?  JSON.stringify('https://5c37dde874b24d8988f04157fade6015@o345774.ingest.sentry.io/5684540') : JSON.stringify('https://e7caecf8f1a54fbe99a50a4801237625@o345774.ingest.sentry.io/5671754')),
   })
 );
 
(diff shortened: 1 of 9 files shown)


config/webpack/environment.js
```diff
@@ -1,10 +1,19 @@
...
-    jQuery: 'jquery'
+    jQuery: 'jquery',
+    SentryLogger: 'sentry-logger',
+  })
+);
+
+environment.plugins.prepend('env',
+  new webpack.DefinePlugin({
+    'NODE_ENV': JSON.stringify(process.env.NODE_ENV),
+    'SENTRY_DSN': (JSON.stringify(process.env.NODE_ENV) == 'production' ?  JSON.stringify('https://5c37dde874b24d8988f04157fade6015@o345774.ingest.sentry.io/5684540') : JSON.stringify('https://e7caecf8f1a54fbe99a50a4801237625@o345774.ingest.sentry.io/5671754')),
```
brenetic, line 16: i would prefer not to inject them openly. can we not create a `.js.erb` file of some kind to access `<%= ENV['SENTRY_DSN'] %>`?

chrispymm: Yeah, I guess we can do that. Didn't think of doing it that way.  Sentry don't appear to consider the dsn url to be 'secret' as their recommendation is basically to stick in in your sourcecode.  But I'll take a look at pulling it from `ENV`

chrispymm: Hmmm.... I'm not sure this will work in my current setup.  

I injected things via webpack, as that allows me to have webpack initialize and provide the SentryLogger instance to the whole app.  As such it doesn't have access to any variables that would exist in a `js.erb` template.

The advantage of this is that it makes the SentryLogger globally available.  Otherwise It would need to be instantated / required in any file we wanted to sentry log from, which feels a bit cumbersome.

chrispymm: OK, so I've got this working now using the ENV vars.  This latest method sets the `SENTRY_DSN` and `PLATFORM ENV` vars on the `window`, which are then used in the `SentryLogger` class, which is instantiatesd in the `DefaultController` that all other controllers inherit from.  This means the setnry logger  is available in all controllers

An alternative method is in commit [df8a4ed](https://github.com/ministryofjustice/fb-editor/pull/1855/commits/df8a4edb3986a5fa9f590a2525396fa36bd7b1d5) this works slightly differently, instantiating the `SentryLogger` in `index.js` and making it available on the window. The only potential advantage here is that `window` is avaialable everywhere - even outside of the controller files.... 

Let me know what you think

brenetic (approved):
🚀  great stuff!

PR #1860, opened 2022-11-03
Update confirmation email user guide link

chrispymm:
Updates the link to the user guide in the warning if the user has no email components in their form.


brenetic (approved):
🚀

PR #1861, opened 2022-11-03
Update settings screen to use components for consistency

chrispymm:
Settings screens had incorrect and inconsistent spacings between header, (optional) desciription text and page content.

This PR fixes the spacing issues, and also updates the way the settings screens are rendered.

Previously there was a helper method to handle rednering the settings screen wrapper, and another to render a back link.

These have been migrated to use ViewComponents.  Given that we are using the GovukComponents gem to provide us with consistent components from the GOVUK design system, we can start to build out our own UI components for MOJ Forms.

This has several advantages:
* Ensures consistency - of markup structure, classes and accessibility
* Reduces the need for backend devs to write out HTML
* Encapsulates more of the template logic in ruby classes
* Simplifies view templates
* Gives the potential for these components to be tested (if required)
* Should help break up css into more manageable componentized files

Components have been placed in an `MojForms` namespace, and all inherit from the `GovukComopnent::Base`.

Introduces a new `mojf-` naming convention for MOJ Forms specific css (with an aim to migrate away from the `fb-` prefix), which shoudl help delineate which classes are coming from the design system, and which are unnique to the editor.

**Next Steps:**
It might be nice to introduce a helper pattern similar to GovukComponents to allow for writing `<%= mojforms_settings_screen() %>` instead of `<%= render MojForms::SettingsScreenComponent.new() %>`

### DIFF
--- a/app/components/moj_forms.rb
+++ b/app/components/moj_forms.rb
@@ -1,1 +1,0 @@
-module MojForms; end
--- a/app/components/moj_forms.rb
+++ b/app/components/moj_forms.rb
@@ -1,0 +1,1 @@
+module MojForms; end
--- a/app/components/moj_forms/back_link_component.rb
+++ b/app/components/moj_forms/back_link_component.rb
@@ -1,29 +1,31 @@
-class MojForms::BackLinkComponent < GovukComponent::Base
-  attr_reader :text, :href
+module MojForms
+  class BackLinkComponent < GovukComponent::Base
+    attr_reader :text, :href
 
-  def initialize(href:, text: t('actions.back'), classes: [], html_attributes: {})
-    @href = href
-    @text = text
+    def initialize(href:, text: t('actions.back'), classes: [], html_attributes: {})
+      @href = href
+      @text = text
 
-    super(classes: classes, html_attributes: html_attributes)
-  end
+      super(classes: classes, html_attributes: html_attributes)
+    end
 
-  def call
-    link_to href, **html_attributes do
-      concat tag.span '<svg width="7" height="12" viewBox="0 0 7 12" fill="none" role="image" aria-hidden="true">
+    def call
+      link_to href, **html_attributes do
+        concat tag.span '<svg width="7" height="12" viewBox="0 0 7 12" fill="none" role="image" aria-hidden="true">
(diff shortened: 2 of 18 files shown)


app/components/moj_forms/back_link_component.rb
```diff
@@ -0,0 +1,29 @@
+class MojForms::BackLinkComponent < GovukComponent::Base
```
brenetic, line 1: no general consensus as namespacing  like this, but the rest of the app does namespacing like so:

```
module MojForms
  class BackLinkComponent < GovukComponent::Base
  end
end
```

app/components/moj_forms.rb
```diff
@@ -0,0 +1 @@
+module MojForms; end
```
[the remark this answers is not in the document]
chrispymm: yeah, I must admit, I just copied the setup from the GovukComponents repo.... I'll remove and see what happens

app/views/settings/confirmation_email/index.html.erb
```diff
@@ -1,6 +1,9 @@
-<%= editor_settings_screen(settings_submission_index_path) do %>
-  <h1 class="govuk-heading-xl"><%= t('settings.confirmation_email.heading') %></h1>
-  <p><%= t('settings.confirmation_email.description') %></p>
+<%= render MojForms::SettingsScreenComponent.new(
+  heading:t('settings.confirmation_email.heading'),
+  description: t('settings.confirmation_email.description')
+  ) do |c| %>
+
+  <% c.with_back_link(href: settings_submission_index_path) %>
```
brenetic, line 6: does this need to use `<%=` in order to render something?

config/locales/en.yml
```diff
@@ -34,7 +34,7 @@ en:
     branching_title: 'Branching point %{branching_number}'
     confirmation_email_subject: "Your submission to ‘%{service_name}’"
     confirmation_email_body: "Thank you for your submission to ‘%{service_name}’.\n\r\nA copy of the information you provided is attached to this email."
-  notifictaion_banners:
+  notification_banners:
```
brenetic, line 37: 🙈

app/components/moj_forms/back_link_component.rb
chrispymm: I defaulted to writing it like that, but in the midst of troubleshooting, I opted to directly copy GovukComponents just to make sure... I think I actually prefer having the `module` defnition in there.

brenetic (approved):
🚀

PR #1864, opened 2022-11-04
Fix acceptance tests

chrispymm:
This fixes the boken acceptance tests after the previous merge to sort out and tidy up the settigns screens

It also corrects test that were failing after enabling the config setting to fail if a translation is missing.


mattempty (approved):
🚀

PR #1866, opened 2022-11-04
Update publishing warnings

H3ll3m4:
## Context
Publishing warnings are being updated. Some of the content has to be changed but also the display of autocomplete warnings with the other warnings.

## Implementation
In order to change the content it is a simple update of the translation file en.yml. In order to have the publishing warnings for autocomplete uniform with the rest, we add the publish them in the list if presenters for the publishing page presenter and remove them from the publish environment view.

## Tests
Unit tests and publishing acceptance tests have been updated. Local testing with manual observation performed.

### DIFF
--- a/app/presenters/autocomplete_items_presenter.rb
+++ b/app/presenters/autocomplete_items_presenter.rb
@@ -25,19 +25,6 @@ class AutocompleteItemsPresenter
   end
 
   def messages
-    @messages ||=
-      pages_with_autocomplete_component.each_with_object([]) do |page, arry|
-        page.components.each do |component|
-          next unless component.uuid.in?(component_uuids_without_items)
-
-          arry.push(
-            { component_title: component.humanised_title, page_uuid: page.uuid }
-          )
-        end
-      end
-  end
-
-  def message
     env = ENV['RAILS_ENV'] == 'development' ? 'dev' : ENV['RAILS_ENV']
     pages_with_autocomplete_component.each_with_object([]) do |page, arry|
       page.components.each do |component|
@@ -51,5 +38,10 @@ class AutocompleteItemsPresenter
         arry.push(msg)
       end
     end
+  end
+
+  def message
+    # because submission warnings presenter expect presenter to have methods message
+    messages
   end
 end
--- a/app/presenters/autocomplete_items_presenter.rb
+++ b/app/presenters/autocomplete_items_presenter.rb
@@ -1,11 +1,12 @@
 class AutocompleteItemsPresenter
   include ActionView::Helpers
   include GovukLinkHelper
-  attr_reader :autocomplete_items, :service
+  attr_reader :autocomplete_items, :service, :deployment_environment
 
-  def initialize(service, autocomplete_items)
+  def initialize(service, autocomplete_items, deployment_environment)
     @service = service
     @autocomplete_items = autocomplete_items
+    @deployment_environment = deployment_environment
   end
 
   def autocomplete_component_uuids
@@ -25,14 +26,13 @@ class AutocompleteItemsPresenter
   end
 
   def messages
-    env = ENV['RAILS_ENV'] == 'development' ? 'dev' : ENV['RAILS_ENV']
     pages_with_autocomplete_component.each_with_object([]) do |page, arry|
       page.components.each do |component|
         next unless component.uuid.in?(component_uuids_without_items)
 
         link = govuk_link_to(component.humanised_title, Rails.application.routes.url_helpers.edit_page_path(service.service_id, page.uuid))
 
-        description = I18n.t("publish.autocomplete_items.#{env}.message", title: link)
+        description = I18n.t("publish.autocomplete_items.#{deployment_environment}.message", title: link)
 
         msg = description.html_safe
         arry.push(msg)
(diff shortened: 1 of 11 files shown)


app/presenters/autocomplete_items_presenter.rb
```diff
@@ -23,15 +26,22 @@ def pages_with_autocomplete_component
   end
 
   def messages
```
njseeto, line 28:
```suggestion
  def message
```

Are we able to rename this method to `message` and then delete the new `message` method?

app/presenters/autocomplete_items_presenter.rb
```diff
@@ -23,15 +26,22 @@ def pages_with_autocomplete_component
...
-
-          arry.push(
-            { component_title: component.humanised_title, page_uuid: page.uuid }
-          )
-        end
+    pages_with_autocomplete_component.each_with_object([]) do |page, arry|
+      page.components.each do |component|
+        next unless component.uuid.in?(component_uuids_without_items)
+
+        link = govuk_link_to(component.humanised_title, Rails.application.routes.url_helpers.edit_page_path(service.service_id, page.uuid))
```
njseeto, line 33:
```suggestion

```
could we move this method to a private method? Then we can just call it in the `description` variable. Something like:

```
  def messages
    pages_with_autocomplete_component.each_with_object([]) do |page, arry|
      page.components.each do |component|
        next unless component.uuid.in?(component_uuids_without_items)

        description = I18n.t(
          "publish.autocomplete_items.#{deployment_environment}.message",
          title: link(component, page)
        )

        msg = description.html_safe
        arry.push(msg)
      end
    end
  end

  private

  def link(component, page)
    govuk_link_to(
      component.humanised_title,
      Rails.application.routes.url_helpers.edit_page_path(service.service_id, page.uuid)
    )
  end
```

app/presenters/autocomplete_items_presenter.rb
```diff
@@ -23,15 +26,22 @@ def pages_with_autocomplete_component
...
+
+        description = I18n.t("publish.autocomplete_items.#{deployment_environment}.message", title: link)
```
njseeto, line 35:
```suggestion
        msg = I18n.t(
          "publish.autocomplete_items.#{deployment_environment}.message",
          title: link
        ).html_safe
```
Not a biggie, but we could chain the `html_safe` in the variable assignment and rename the variable, then we can delete the `msg = description.html_safe` line.

spec/presenters/publishing_page_presenter_spec.rb
```diff
@@ -34,13 +34,17 @@
 
     context 'submission warning presenters' do
       let(:presenters) do
-        [submission_pages_presenter, from_address_presenter]
+        [submission_pages_presenter,
+         autocomplete_warning,
+         from_address_presenter]
```
njseeto, line 39:
```suggestion
        [
          submission_pages_presenter,
          autocomplete_warning,
          from_address_presenter
        ]
```

Small formatting suggestion

njseeto (commented):
Looks great, just a couple of suggestions.

app/presenters/publishing_page_presenter.rb
```diff
@@ -59,6 +59,10 @@ def submission_pages_presenter
   end
 
   def submission_warning_presenters
-    [submission_pages_presenter, from_address_presenter]
+    presenters = [submission_pages_presenter, from_address_presenter]
+
+    return presenters if deployment_environment == 'dev'
```
brenetic, line 64: why do we not want the autocomplete presenter to be there if it's dev?

brenetic:
can we have a screenshot please?

i may have completely misunderstood the story, but are the autocomplete warnings now part of the list of _all_ publishing warnings on the publishing page? if that's the case then maintaining the `autocomplete_warning` partial might not be necessary as the `SubmissionWarningsPresenter` will just map over any presenters that it's initialized with and spit out the messages

H3ll3m4:
We moved the autocomplete warning in the first header group only for publishing to Live. However, as it has been clarified this morning after the stand up, we want the publishing to Test to have autocomplete warnings separated, so like before we started the work on this feature. The results looks like this, with a server running locally : 
Screenshot 2022-11-10 at 16 44 08

In addition, we added a few updates of the content, following version 1.3 of the [publishing warnings](https://docs.google.com/spreadsheets/d/11FvkKJcO1Yfn3bVOM9IPSD0_LHjqgTSJmASx33lYe-E/edit#gid=814462722), see tab 1.3, and only the text written ToDo on column K.

app/views/publish/_autocomplete_warning.html.erb
```diff
@@ -1,8 +1,6 @@
-<% autocomplete_warning.messages.each do |message| %>
+<% autocomplete_warning.message.each do |message| %>
```
brenetic, line 1: is there only one message now per presenter? if so do we need to loop? if not we could keep both methods

brenetic (changes requested):

app/presenters/autocomplete_items_presenter.rb
```diff
@@ -23,15 +26,24 @@ def pages_with_autocomplete_component
...
+        msg = I18n.t("publish.autocomplete_items.#{deployment_environment}.message", title: link(component, page)).html_safe
+        arry.push(msg)
       end
+    end
+  end
+
+  # the method message is required by the submission warning presenter
+  def message
+    messages
+  end
```
brenetic, line 42: `alias :message, :messages`

brenetic (approved):
🚀

PR #1868, opened 2022-11-07
Ensure confirmation email modal is prioritised over branching modals

njseeto:
[Trello](https://trello.com/c/0xsnJd78/3069-confirmation-email-no-warning-model-when-deleting-a-branching-destination-page)
Currently, when a user tries to delete a page that has an email component used in confirmation email and is a destination of a branch, the user is presented with the delete destination page modal. This modal allows the user to delete the page, this is not what we want, instead we want the confirmation email modal to appear as this blocks the user from deleting the page until they have changed their confirmation email settings.
This commit implements this change.

Note: We do not need to change the behaviour on the question level as the branching modals will only trigger for  radio or checkbox options and the confirmation email modals will only trigger on email components on a multiple question page.

### DIFF
--- a/app/models/destroy_page_modal.rb
+++ b/app/models/destroy_page_modal.rb
@@ -7,11 +7,11 @@ class DestroyPageModal
   delegate :expressions, :branches, to: :service
 
   PARTIALS = {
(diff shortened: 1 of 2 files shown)


brenetic (approved):
🚀

PR #1869, opened 2022-11-07
Wait for intial delay before running rake task in cronjobs

brenetic:
We have been getting a number of connection errors to the DB when the cronjob task containers are spinning up.

Add a wait time before running the task to see if this helps


njseeto (approved):
🚀

PR #1875, opened 2022-11-08
Added cica domain to both the allow list and send from allow list.

mattempty:
This change will be needed for when cica is added to auth0 to both log in and also set up the send to email.

https://trello.com/c/L3QC762B


njseeto (approved):
🚀

PR #1880, opened 2022-11-09
Update to the login allow list

mattempty:
We have added cica successfully to Auth0, however the login is rejected. Checking the logs the domain is converted to uppercase in the response which may not pass the Editor checks. Added the upper version to the allow list.

https://trello.com/c/L3QC762B


brenetic:
i think we can downcase in the validation implementation

brenetic (approved):
🚀

PR #1884, opened 2022-11-11
Do not build testable editor workers

brenetic:
This change will stop any testable editors creating corresponding worker deployments.

Having several sets of different code bases all reading off of the same delayed job queue causes uncecessary behavioural disparities.


njseeto (dismissed):
🚀

StevenLeighton21:
Can we merge this @njseeto ?

njseeto:
> 

yep can do.

njseeto (approved):

PR #1885, opened 2022-11-14
Use `do_not_send_confirmation_email?` method

njseeto:
As part of the confirmation email work, we should reject the confirmation email environment variables if the form editor has the confirmation email feature disabled.


brenetic (approved):
🚀

PR #1887, opened 2022-11-15
Do not reject service email from configuration

H3ll3m4:
When a user has enabled send by email but not enabled confirmation email. We want to make that the service email from is still injected in the form container.


brenetic (approved):
🚀

PR #1888, opened 2022-11-16
Add individual form maintenance page

chrispymm:
Adds settings in the admin area to enable maintenance mode for an individual form and customise the content of the page that is displayed.

* Create edit services route / page in admin
* Create MaintenanceModeSettings Model
* Create MaintenanceModeSettingsUpdater Model
* Inject maintenance env vars in provisioner when required
* Bumps metadata_presenter to 2.17.23 whihc includes the maintenance page template.

One thing worth noting is that the maintenance page content is stored in the DB as a base64 encoded string.  This is because it needs to be injected into the published docker container as an ENV var and that doesn't support multi-line strings.  So base64 encoding was a simple solution to making it a single line.  The long-term solution for this would be to have the maintenance settings stored within the form metadata.

image
Add Edit link to serviices list in admin

image
Edit screen contains settings for maintenance page


app/services/publisher/service_provisioner.rb
```diff
@@ -130,7 +130,7 @@ def live_production?
     end
 
     def config_map
-      service_configuration.reject(&:secrets?).reject(&:do_not_send_submission?).reject(&:do_not_send_confirmation_email?)
+      service_configuration.reject(&:secrets?).reject(&:do_not_send_submission?).reject(&:do_not_send_confirmation_email?).reject(&:not_in_maintenance_mode?)
```
brenetic, line 133:
```
To everything chain, chain, chain
There is a season chain, chain, chain
```

i feel this is ripe for a refactor. as we have found between the bug fix and your story the logic of it is not entirely easy to parse. 

i guess in another PR though

chrispymm: yeah probably.  Its definitely not the most readable line of code! I guess for a refactor we'd likely want to push more of it back into the `ServiceConfiguration`? But definitely for another PR

PR #1900, opened 2022-11-21
Fix focus issue on activated menus after modal trigger

chrispymm:
Bugfix for: https://trello.com/c/fypC35DH/3105-vertical-keyboard-navigation-in-dropdown-is-broken-after-opening-closing-a-modal

Activated menu items that triggered a modal, would no longer corectly show focus afttre the modal had opened. Using the arrow keys would update ActiveDescendant, but the item would not be highlighted (focus styles not applied)

I suspect this is somehow related to jQuery UI dialog, however the fix appears to be to call focus() on the html node itself rather then on the jQuery $node.

This commit also removes some legacy focus management code that is no longer needed.

Also lots of trailing whitespace automatically removed (sorry for the noisy diff)

### DIFF
--- a/app/javascript/src/components/menus/activated_menu.js
+++ b/app/javascript/src/components/menus/activated_menu.js
@@ -6,7 +6,7 @@
  * button to open the menu.
  *
  * This class does the following:
- *  - Takes the provided jQuery <ul> node and wraps it in an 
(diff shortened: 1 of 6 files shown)


njseeto (approved):
🚀

PR #1901, opened 2022-11-21
Publish warning updates

chrispymm:
Updates the publish warnings to the latest content / logic.

Default no submission output is replaced by 2 messages - one if Sedn by email is off, and another if send by email is off and confirmation email is on

This also pulls the from address warnings out to be top level warnings in the test environment, while leaving them as sub-messages of the Submission Warning in live.

image
'Send by Email' disabled

---

image
'Send by Email' disabled and `Confirmation Email` enabled

---

image
From address warnings moved to top-level in Test, but included in submission warnings in Live


njseeto (approved):
🚀

PR #1902, opened 2022-11-21
Add reference number feature flags, templates, models and service configuration

njseeto:
[Trello](https://trello.com/c/GvjBsA1u/3076-reference-number-add-reference-number-link-in-settings-and-reference-number-template)
[Trello](https://trello.com/c/mpdmNH93/3086-reference-number-add-referencenumber-as-a-submission-setting)

### Add reference number feature flag, templates, route and controller
We want to have the feature flags as reference number is being developed. We want the reference number feature to only be available in Test.

We will need to revisit this to ensure we have the warning message and any remaining text present on the template.

### Set up the `ReferenceNumberUpdater` and `ServiceConfiguration`
We will inject the Runner with the 'REFERENCE_NUMBER' config if the user enables this setting. If the setting is disabled, 'REFERENCE_NUMBER' will not be injected into the Runner.

We want to update this config globally for both dev and production environments as the email templates could be different for each environment.


mattempty (approved):
🚀

app/views/settings/reference_number/_form.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= f.govuk_check_boxes_fieldset :"reference_number",
+      multiple: false,
+      legend: { text: t('settings.reference_number.heading'), size: 'l'} do %>
+      <div class="govuk-hint" id="reference_number_hint"><%= t('settings.reference_number.hint') %></div>
+      <%= f.govuk_check_box :"reference_number", 1, 0,
+        multiple: false,
+        link_errors: true,
+        label: { text: "Enable reference numbers" },
```
chrispymm, line 8: The label should be in the translations

chrispymm (changes requested):

app/views/settings/reference_number/_form.html.erb
njseeto: good spot! All done now

chrispymm (approved):
🚀

PR #1905, opened 2022-11-23
Fix missing arrow condition

chrispymm:
This is a fix for a missing arrow discovered on the DDaT Assessment Eligibility form.

**Before:**
Screenshot 2022-11-23 at 09 43 50

**After:**
image

### DIFF
--- a/app/javascript/src/component_flow_connector_path.js
+++ b/app/javascript/src/component_flow_connector_path.js
@@ -571,11 +571,12 @@ class ForwardDownBackwardUpPath extends FlowConnectorPath {
 
   constructor(points, config) {
     super(points, config);
+
     var dimensions = {
       forward1: Math.round(this.points.via_x - CURVE_SPACING),
       down: Math.round(utilities.difference(this.points.from_y, config.bottom) - (CURVE_SPACING * 2)),
       backward: Math.round(utilities.difference(this.points.from_x + this.points.via_x, this.points.to_x)),
-      up: Math.round(utilities.difference(config.bottom, this.points.from_y) + utilities.difference(this.points.from_y, this.points.to_y) - (CURVE_SPACING * 2)),
+      up: Math.round( utilities.difference(config.bottom, this.points.to_y) - (CURVE_SPACING * 2)),
       forward2: 0
     }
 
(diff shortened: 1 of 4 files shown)


app/javascript/src/controller_services.js
```diff
@@ -889,9 +890,7 @@ function calculateAndCreatePageFlowConnectorPath(points, config) {
       }
     }
     else {
-      if(up) {
-        new ConnectorPath.ForwardDownBackwardUpPath(points, config);
-      }
```
chrispymm, line 894: The `if(up)` here only drew the line if the 'to' FlowItem was in a row above the 'from' FlowItem. (see line 866)

This is not what we wanted. 

Because we always go `down` in order to go `back` we always need to go `up` regardless of what row the destination item is on.

app/javascript/src/component_flow_connector_path.js
```diff
@@ -571,11 +571,12 @@ class ForwardDownBackwardUpPath extends FlowConnectorPath {
 
   constructor(points, config) {
     super(points, config);
+
     var dimensions = {
       forward1: Math.round(this.points.via_x - CURVE_SPACING),
       down: Math.round(utilities.difference(this.points.from_y, config.bottom) - (CURVE_SPACING * 2)),
       backward: Math.round(utilities.difference(this.points.from_x + this.points.via_x, this.points.to_x)),
-      up: Math.round(utilities.difference(config.bottom, this.points.from_y) + utilities.difference(this.points.from_y, this.points.to_y) - (CURVE_SPACING * 2)),
+      up: Math.round( utilities.difference(config.bottom, this.points.to_y) - (CURVE_SPACING * 2)),
```
chrispymm, line 579: Because we always go down to the bottom in order to go back up, we can simply calculate the up path as the difference between the bottom and the 'to' item. i.e. simply the opposite of a down arrow (compare line 577 and 579)

app/javascript/src/utilities.js
```diff
@@ -281,12 +281,14 @@ function maxWidth($collection) {
...
-  if(a > b)
+function difference(a, b, abs=true) {
+  if(abs) {
+    return Math.abs(a - b);
+  } else {
     return a - b;
-  else
-    return b - a;
+  }
 }
```
chrispymm, line 292: * Simplify method to use JS built-in `Math.abs` 
* Add param to indicate if we want a signed result.

njseeto (approved):
🚀

PR #1906, opened 2022-11-23
Fix arrow overlap issues

chrispymm:
This PR fixes the fact that line overlapping was broken.

It appears that during test-writing changes were made to the implementation of the `FlowConnectorPath` class that made the `id` property private.

A public `prop` accessor function was made available (not entirely sure at this point why standard getters weren't used.)

It seems that the code in `adjustOverlappingFlowConnectorPaths()` was not updated to use the `prop` accessor.  As such `path.id` and `current.id` were returning `undefined` - this was meaning no lines were being compared for overlaps.

This went unnoticed, as the 'undefined' variables didn't trigger any errors, just prevented an essential function call.

Strike this up as a reason for using Typescript 😉

**Before:**
Screenshot 2022-11-23 at 10 23 44

**After:**
image

There is still an outstanding bug / rendering issue that I can see: nudged lines at the top of the flow viewport get cut off.  Don't know if this was an existing bug, or if it is new....

image

### DIFF
--- a/app/javascript/src/component_flow_connector_path.js
+++ b/app/javascript/src/component_flow_connector_path.js
@@ -571,11 +571,12 @@ class ForwardDownBackwardUpPath extends FlowConnectorPath {
 
   constructor(points, config) {
     super(points, config);
+
     var dimensions = {
       forward1: Math.round(this.points.via_x - CURVE_SPACING),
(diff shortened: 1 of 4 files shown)


njseeto (approved):
🚀

PR #1907, opened 2022-11-23
Show placeholder reference number when reference number is enabled

njseeto:
### Show placeholder when reference number is enabled
When in the Editor or in preview mode, and reference number is enabled, we want to show a placeholder reference number.
This is done through the methods `show_reference_number` and `reference_number_enabled?`.

### Bump to presenter 2.17.24

This version of the presenter has the new confirmation email template that includes reference numbers.

Screenshot 2022-11-25 at 13 14 33


chrispymm (approved):
All loloks good!

PR #1909, opened 2022-11-24
Add aria describedby attirbute and update content

chrispymm:
Makes a couple of tweaks to the reference number templates. 

* Adds the hint text into the aria-describedby attribute of the input
* Adds the warning text into the aria-describedby attribute of the input - feels like this is important enought that we want them to know it if all they do is focus the input.
* Separates out the 'heading' text string into 'heading' and 'legend' as they will be different when we do pay
* Add `rel="noopener noferrer"` to the `target="_blank"` link - required for security
* Tweak capitalisation of headings, and ensure 'Reference numbers' is always plural (as per figma)


njseeto (approved):
🚀

PR #1910, opened 2022-11-25
Fix logic around display of autocomplete warnings

chrispymm:
Fixes some logic around the publishing warnings and how we handle autocomplete issues.

The problem was that both in dev and live, the autocomplete warnings only showed if 'collect information by email' was off.  

In dev the solution is pretty easy, as we can show both of the top-level warnings at the same time.

In live the autocomplete warnings are part of the 'submission warnings' block.

So we need to ensure the submission warnings includes the right warnings (i.e. no from address warnings when collect info by email is off), and gets shown if there are any autocomplete issues.

**Before:**
image
Form has an autocomplete without any options, but the user wouldn't know

**After:**
image
Email warning and autocomplete warnings show, but there are no 'From address' warnings in live due to 'collect info by email' being off

### DIFF
--- a/app/presenters/service_output_warning_presenter.rb
+++ b/app/presenters/service_output_warning_presenter.rb
@@ -15,6 +15,10 @@ class ServiceOutputWarningPresenter
 
   delegate :present?, to: :message
 
+  def blank?
+    !present?
+  end
+
   def message
     key = send_confirmation_email? ? :confirmation_email : :default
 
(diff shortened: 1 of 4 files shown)


app/presenters/service_output_warning_presenter.rb
```diff
@@ -15,6 +15,10 @@ def initialize(service_id:, deployment_environment:, messages:)
 
   delegate :present?, to: :message
 
+  def blank?
```
njseeto, line 18: could we use ruby's in built `blank?` method rather than defining our own?

njseeto (changes requested):
looks good, just one small question

brenetic (approved):
🚀

njseeto (approved):
🚀

PR #1914, opened 2022-11-28
Testable first publish message

chrispymm:
Thsi all works.

I'm not convinced by the amount of `PublishService` stuff there is in the `PublishController`

I feel like more of it should be in the 'PublishServicePresenter' - but a presenter needs to be instanciated with an instance*

*I've realised this morning that that is why it was setup the way it was previously - you could call the class directly, which then instanciated itself with the latest instance - so could return to something like that... ... my other thought had been whether a `LatestPublishServicePresenter` that didn't require an instance....

The other annoyance is the need to get a count of the PublishService's there  doesn;t feel a neat way to do this... without another DB call, and also it feels messy putting it in a presenter for an instance, as the presenter then has to go and do a db call to determine the count.

It might be possible to do a custom select, but that's not overly clean either....
```
publish_count = PublishService.select('count(*)').where(service_id: xxxxx,  deployment_environment: xxxx).to_sql

PublishService.select(
  '*',
  "(#{publish_count} publish_count"
).where(
  service_id: xxxxx,
  deployment_environment: xxxx
)
```

It feels like ideally we want a PublishService instance itself to know whether it is the first publish...  but that is a bit more complex to implement


acceptance/features/publishing_page_spec.rb
```diff
@@ -136,7 +136,8 @@
       when_I_visit_the_publishing_page
       then_I_should_see_from_address_warning('pending')
     end
-  end
+
+      end
```
brenetic, line 140:
```suggestion
  end
```

brenetic (approved):
🚀

PR #1916, opened 2022-11-29
Moving 2 team members to the admin list.

mattempty:
This change allows Assma and Fabien to use the admin functions.


brenetic (approved):
🚀

PR #1919, opened 2022-11-30
Add correct values in email templates when reference number is enabled/disabled

njseeto:
With the implementation of reference number, we need to be able to provide different service email and confirmation email defaults depending on whether reference number is enabled or disabled.

The configs that need to have reference number appended when reference number is enabled include:
`SERVICE_EMAIL_BODY`, `SERVICE_EMAIL_SUBJECT`, `SERVICE_EMAIL_PDF_HEADING`, `CONFIRMATION_EMAIL_BODY`, `CONFIRMATION_EMAIL_SUBJECT`

To achieve this, we have changed the default values by inserting `{{reference_number_sentence}}` and gsubbing this with either a sentence or an empty string.

If there is the phrase `{{reference_number}}` saved in the configs, they will be swapped for the actual reference number in the Runner application.

### DIFF
--- a/app/models/base_email_settings.rb
+++ b/app/models/base_email_settings.rb
@@ -1,8 +1,6 @@
 class BaseEmailSettings
   include ActiveModel::Model
 
-  REFERENCE_NUMBER_PLACEHOLDER = '{{reference_number_sentence}}'.freeze
-
   def settings_for(setting_name)
     params(setting_name).presence ||
       database(setting_name) ||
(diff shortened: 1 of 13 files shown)


brenetic:
for the email subject and the pdf subject do we also want those to include `Your reference number is`? Or does that only need to be in the body?

app/models/base_email_settings.rb
```diff
@@ -26,4 +28,25 @@ def params(setting_name)
...
+  def insert_reference_number_sentence(content)
+    reference_number_sentence = I18n.t('default_values.reference_number_sentence')
+    content.gsub(REFERENCE_NUMBER_PLACEHOLDER, reference_number_sentence)
+  end
+
+  def remove_reference_number_sentence(content)
+    content.gsub(REFERENCE_NUMBER_PLACEHOLDER, '')
+  end
+
+  def update_default_value(setting)
```
chrispymm, line 45: You updated this as I was in the middle of suggesting to refactor to having a method instead of all the ternarys!

My only thought now is that `update_default_value()` is not overly descriptive, and maybe too generic - I don't intrinsically understand what it is going to do, and also we're likely to need another method like this for payment links.

My initial thought had been `substitute_reference_number(content)`
but thinking ahead maybe `substitute_placeholder(content, placeholder)` is better?

We would then end up with something like:
```
def confirmation_email_body
  substitute_placeholder(settings_for(:confirmation_email_body), REFERENCE_NUMBER_PLACEHOLDER)
  substitute_placeholder(settings_for(:confirmation_email_body), PAYMENT_LINK_PLACEHOLDER)
end
```

njseeto:
> for the email subject and the pdf subject do we also want those to include `Your reference number is`? Or does that only need to be in the body?

yep good spot, have expanded the method to take in a  second argument, `placeholder` so we can be flexible with the text that is subbed in

chrispymm (approved):
🚀

PR #1921, opened 2022-12-01
Adding the JAC domain to both the mail to and auth allow lists

mattempty:
The JAC are being given the ability to log into the editor using their own domain. The judicialappointments.gov.uk uses the DOM1 Azure AD (same as justice). Adding this will allow access to the editor.

https://trello.com/c/JazRQJZK


brenetic (approved):
🚀

PR #1922, opened 2022-12-01
Use correct email default values when reference number is enabled/disabled

njseeto:
### Introduce `ContentSubstitutor` class to substitute placeholder text
This class is responsible for finding the `{{reference_number_placeholder}}` and replacing it with either `''` or the default reference number sentence/subject which is found in the locales.
We decided to create this class as we know there will be future features that will require a similar mechanism of substitution.

We have also used the `EmailSettingsUpdater` and `ConfirmationEmailSettingsUpdater` classes in the `ReferenceNumberUpdater` as a way to leverage the `create_or_update!` methods in these classes to make sure we save any changes to the service configurations to the database.

### DIFF
--- a/app/helpers/content_substitutor_helper.rb
+++ b/app/helpers/content_substitutor_helper.rb
@@ -1,0 +1,15 @@
+module ContentSubstitutorHelper
+  def content_substitutor
+    @content_substitutor ||= ContentSubstitutor.new(
+      service_name: service.service_name,
+      reference_number_enabled: reference_number_enabled
+    )
+  end
+
+  def reference_number_enabled
+    ServiceConfiguration.where(
+      service_id: service.service_id,
+      name: 'REFERENCE_NUMBER'
+    ).present?
+  end
+end
--- a/app/models/content_substitutor.rb
+++ b/app/models/content_substitutor.rb
@@ -73,14 +73,20 @@ class ContentSubstitutor
     )
   end
 
+  def assign(setting)
+    public_send(setting)
+  rescue NoMethodError
+    nil
+  end
+
   private
 
   def subject_content
-    I18n.t('default_values.reference_number_subject')
+    @subject_content ||= I18n.t('default_values.reference_number_subject')
   end
 
   def body_content
-    I18n.t('default_values.reference_number_sentence')
+    @body_content ||= I18n.t('default_values.reference_number_sentence')
   end
 
   def insert_placeholder_sentence(setting, placeholder, content)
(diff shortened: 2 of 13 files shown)


app/helpers/content_substitutor_helper.rb
```diff
@@ -0,0 +1,16 @@
+module ContentSubstitutorHelper
+  def content_substitutor(deployment_environment)
+    @content_substitutor ||= ContentSubstitutor.new(
```
brenetic, line 3: i am not sure we can memoize while also passing an argument

app/helpers/content_substitutor_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+    )
+  end
+
+  def reference_number_enabled(deployment_environment)
+    ServiceConfiguration.find_by(
+      service_id: service.service_id,
+      deployment_environment: deployment_environment,
+      name: 'REFERENCE_NUMBER'
+    ).present?
+  end
```
brenetic, line 15: to get round the need for passing the deployment environment we could use `where`

app/models/content_substitutor.rb
```diff
@@ -0,0 +1,107 @@
...
+  def assign(setting)
+    public_send(setting)
+  rescue NoMethodError
+    nil
+  end
+
+  private
+
+  def subject_content
+    I18n.t('default_values.reference_number_subject')
```
brenetic, line 85: can memoize this

app/models/content_substitutor.rb
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  def body_content
+    I18n.t('default_values.reference_number_sentence')
```
brenetic, line 89: can memoize this

app/helpers/content_substitutor_helper.rb
njseeto: all done!

app/helpers/content_substitutor_helper.rb
```diff
@@ -0,0 +1,15 @@
+module ContentSubstitutorHelper
+  def content_substitutor
+    @content_substitutor ||= ContentSubstitutor.new(
+      service_name: service.service_name,
+      reference_number_enabled: reference_number_enabled
+    )
+  end
+
+  def reference_number_enabled
```
brenetic, line 9:
```suggestion
  def reference_number_enabled?
```

one to add in the next PR maybe

brenetic (approved):
🚀

PR #1929, opened 2022-12-02
Update 500 page to have GOVUK Styles

chrispymm:
Updates the 500 page for the editor to have GOVUK styles.

Needs to be fully standalone and cannot rely on editor assets.

Subset of the GOVUK CSS generated using a chrome extenmsion to pull the used styles from one of our live 404 pages. Then tidied up a bit.

**Large screen**
image

**Small screen**
image

### DIFF
--- a/public/500.html
+++ b/public/500.html
@@ -39,6 +39,13 @@
       .govuk-visually-hidden{padding:0!important;border:0!important;}
       .govuk-visually-hidden{position:absolute!important;width:1px!important;height:1px!important;margin:0!important;overflow:hidden!important;clip:rect(0 0 0 0)!important;-webkit-clip-path:inset(50%)!important;clip-path:inset(50%)!important;white-space:nowrap!important;}
       #main-content{min-height:55vh;}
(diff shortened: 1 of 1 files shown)


public/500.html
```diff
@@ -1,67 +1,97 @@
...
-  <div class="dialog">
-    <div>
-      <h1>There is a problem with MoJ Forms</h1>
-      <p>Try again later.</p>
+    <div class="govuk-width-container govuk-body-m">
+      <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
+        <!-- This file lives in public/500.html -->
+        <h1>There is a problem with MoJ Forms</h1>
+        <p>Try again later.</p>
+        <p>If the problem persists, email <a href="mailto: moj-forms@digital.justice.gov.uk">moj-forms@digital.justice.gov.uk</a></p>
```
brenetic, line 90: does this need to have the `govuk-link` class?

chrispymm: Ah yes. Good catch. That measn this isalso an issue on our live 404 page, as that's where I copied the markup from.

PR #1931, opened 2022-12-02
Remove test that triggers a publish

chrispymm:
Didn't think things through and worte a test that triggers a publihs when running in circle.

This then causes failed publish jobs due to the length of the form name.

Remove the test - the only way to acceptance test the feature is to publish, so we'll have to go without for now.


brenetic (approved):
🚀

PR #1935, opened 2022-12-05
Add confirmation email warning to reference no settings page

chrispymm:
Adds the warning block into the reference number settings page if confirmation email is not enabled (in test and live)

image

### DIFF
--- a/app/controllers/settings/reference_payment_controller.rb
+++ b/app/controllers/settings/reference_payment_controller.rb
@@ -1,9 +1,7 @@
 class Settings::ReferencePaymentController < FormController
   before_action :assign_form_objects
 
-  def index
-    @reference_payment = ReferencePaymentSettings.new(service_id: service.service_id)
-  end
+  def index; end
 
   def create
     @reference_payment = ReferencePaymentSettings.new(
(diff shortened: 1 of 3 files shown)


app/controllers/settings/reference_payment_controller.rb
```diff
@@ -26,6 +24,14 @@ def create
 
   def assign_form_objects
     @reference_payment = ReferencePaymentSettings.new(service_id: service.service_id)
+    @confirmation_email_enabled = confirmation_email_setting('dev').try(:send_confirmation_email?) && confirmation_email_setting('production').try(:send_confirmation_email?)
+  end
+
+  def confirmation_email_setting(environment)
+    SubmissionSetting.find_by(
```
brenetic, line 31: `find_by` will return the first record it finds. would `where` be more appropriate, to return a relation of both environments?

brenetic: never mind me, you're calling it twice 😆

brenetic (dismissed):
🚀

brenetic (dismissed):
🚀

brenetic (approved):

PR #1936, opened 2022-12-05
Adding feature flag and settings template for payment link

H3ll3m4:
We're adding a feature flag for payment link. We updated the templates to display a separate checkbox and the translation file. See result below:

Screenshot 2022-12-05 at 12 38 42


brenetic (approved):
🚀

PR #1937, opened 2022-12-05
Reference number only update default config values

brenetic:
When the reference number is enabled or disabled we only want to update
the following service configurations:

- CONFIRMATION_EMAIL_SUBJECT
- CONFIRMATION_EMAIL_BODY
- SERVICE_EMAIL_SUBJECT
- SERVICE_EMAIL_BODY
- SERVICE_EMAIL_PDF_HEADING

Submission settings and any other service configuration records should
be left alone.

By using the EmailSettingsUpdater and the
ConfirmationEmailSettingsUpdater models all other settings and configs
were being removed as their attributes would have been nil at the time
of saving.


njseeto (approved):
🚀

PR #1939, opened 2022-12-05
Add payment link attributes to reference payment model

brenetic:
These attributes are now being passed into the reference payment model from the controller so need to have accessors


njseeto (approved):
🚀

PR #1941, opened 2022-12-06
Payment link settings validation

H3ll3m4:
Implemented payment link settings validator and the warnings displayed on the template. 
There are 3 warnings:
- If the payment link is ticked reference number has to be ticked as well.
- If the payment link is ticked the url should not be empty
- If the payment link is ticked, the url should be conform and start with 'https://www.gov.uk/payments/'

In addition, we don't cater for the case when the payment link is not ticked but the url is present and valid, this because the payment url field only appears if the payment link is ticked. Adding javascript to make the field appear and disappear  still needs to be done.

A summary of warning is displayed at the top see screenshot:

Screenshot 2022-12-06 at 15 44 50

Warnings related to reference number are displayed above the reference number checkbox, while the payment link related warnings are located below the corresponding checkbox:
Screenshot 2022-12-06 at 15 44 40

### DIFF
--- a/app/models/reference_payment_settings.rb
+++ b/app/models/reference_payment_settings.rb
@@ -8,20 +8,22 @@ class ReferencePaymentSettings
                 :deployment_environment
 
   validates :service_id, presence: true
+  validates :payment_link_url, presence: true, if: proc { |obj| obj.payment_link == '1' }
+  validates :payment_link, presence: true, if: proc { |obj| obj.reference_number == '1' }
 
   def reference_number_checked?
-    enabled? || ServiceConfiguration.exists?(service_id: service_id, name: 'REFERENCE_NUMBER')
+    reference_number_enabled? || ServiceConfiguration.exists?(service_id: service_id, name: 'REFERENCE_NUMBER')
   end
 
-  def enabled?
+  def reference_number_enabled?
     reference_number == '1'
   end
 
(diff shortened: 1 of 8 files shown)


app/models/reference_payment_settings.rb
```diff
@@ -8,20 +8,23 @@ class ReferencePaymentSettings
...
 
-  def payment_link_checked?
-    payment_enabled? || ServiceConfiguration.exists?(service_id: service_id, name: 'PAYMENT_LINK')
+  def payment_link_url_enabled?
+    payment_link_url_present? || ServiceConfiguration.exists?(service_id: service_id, name: 'PAYMENT_LINK')
   end
 
-  def payment_enabled?
-    false
+  delegate :present?, to: :payment_link_url, prefix: true
```
brenetic, line 25: nicely done 😆

brenetic: would this also intercept `present?` called on an instance of `ReferencePaymentSettings`? probably nothing to worry about but if it does and there is an expectation of `@reference_payment_settings.present?` that may return an unexpected result. just a thought

H3ll3m4: Good point we may forget and need to check the presence of reference_payment_settings instance so have put back a full definition

PR #1943, opened 2022-12-07
Fix issues with first publish message

chrispymm:
This PR fixes 3 issues with the first publish message display.

### 1 First publish method displayed every time.
We were relying on JS to check wether the form was already published.  It seems that this check was unreliable.  So the `first_publish?` method on the `PublishServicePresenter` has been reinstated.  The backend knows definitively if this form has been published more than once, so we use that to only include the `first_publish_message` template in the view if it is needed.

### 2 First publish message was showing success, but then the link didn't work.
It seemed that relying on a HTTP request 200 was not overly reliable - possibly because the backend can get a 200 response earlier than external requests.
To combat this we switch to using a DNS resolution lookup using Google's public nameservers.  This means we are relying on the DNS having propagated beyond our infrastructure before we return success.

However DNS propagation still takes time, and the internet is made of lots of parts<sup>[1]</sup> and in testing it became obvious that we could not rely on a single success response to notify the user that the service was up.

So we poll the service url every 10 seconds and require 10 consecutive success responses before showing the green OK message to the user.  Does this mean they may have to wait slightly longer than necessary? yes, but it is much more likely that the link will work when we say it will!

### 3 First publish message could be shown again on relaoding the page after seeing the green message.
Again this was due to DNS propagation fickleness.  We were relying on one request to check if the service was already live, and if it has only just been published, there is a higher chance that that request might fail.  So, instead we do 3 calls to the service url - if any of them succeeds we consider the service live.

So the full process looks like this:
1) Publish a form for the first time
2) Page reloads
3) The JS does 3 calls to the service url
4) If any are successful, the service is already live, don't show a message
5) Otherwise, show the first publish message (wait 15mins)
6) JS starts polling the service url at 10s intervals.
7) It will wait for 10 consecutive success responses, once that threshold is reached, the message will turn green.
8) Refresh the page
9) We do step 3 again - and should get at leats one successful response so the message is not shown again.

It is worth noting, that the solution to 3) may obviate the need for fix 1) - but I think it is safer and more relaible to leave that in.

### JS Refactored and modernized 
Due to the fact that we now require multiple calls to urls in 2 separate blocks, the JS code has been refactored away from jQuery `$get` which was getting unduly nested in callback hell, and replaced with more modern `async/await` based code using  `fetch` allowing for much more reradable code.

<sup>1</sup> citation needed


brenetic (approved):
🚀

brenetic (approved):
🚀

PR #1944, opened 2022-12-07
Update link to reflect correct url

njseeto:
This link is currently broken as it was pointing to the user guide section, when we need to point to the settings section. Update the link so it points to the reference number section.


brenetic (approved):
🚀

PR #1945, opened 2022-12-07
Save payment link service configuration to the database

H3ll3m4:
We implemented the logic to store and retrieve the payment url settings in the ServiceConfiguration table as PAYMENT_LINK, with the value being the payment link url. 
For removing the PAYMENT_LINK configuration we have to either untick the checkbox or to empty the payment link url.
Whereas for adding the PAYMENT_LINK configuration we check that both the checkbox is ticked and the payment url is present.
 
We also added the expander capability to the payment url field in the view template, so when you check the payment link checkbox the url field is revealed.

### DIFF
--- a/app/javascript/src/controller_reference_payment.js
+++ b/app/javascript/src/controller_reference_payment.js
@@ -1,0 +1,32 @@
+const DefaultController = require('./controller_default');
+const Expander = require('./component_expander');
+
+class ReferencePaymentController extends DefaultController {
+  constructor(app) {
+    super(app);
+
+    switch(app.page.action) {
+      case 'create':
+      case 'index':
+        this.#index();
+      break;
+    }
+  }
+
+  #index() {
+    this.#addExpanderEnhancement();
+  }
+
+  #addExpanderEnhancement() {
+    var $paymentField = $(".fb-payment-field");
+    var $revealedInput = $paymentField.find('[data-component="Expander"]');
+    var $checkbox = $("input[type=checkbox]", $paymentField);
+    new Expander($revealedInput, {
+      activator_source: $checkbox,
+      auto_open: $checkbox.attr('checked')||$(".govuk-form-group--error", $paymentField).length,
+      wrap_content: false,
+    });
+  }
+}
+
+module.exports = ReferencePaymentController;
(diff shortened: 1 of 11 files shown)


brenetic:
does the design include the grey bar next to the reference number checkbox?

Screenshot 2022-12-07 at 17 06 36

app/javascript/src/controller_reference_payment.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  #addExpanderEnhancement() {
+    var $paymentField = $(".fb-payment-field");
+    var $revealedInput = $paymentField.find('[data-component="Expander"]');
+    var $checkbox = $("input[type=checkbox]", $paymentField);
+    new Expander($revealedInput, {
+      activator_source: $checkbox,
+      auto_open: $(".govuk-form-group--error", $paymentField).length,
+      wrap_content: false,
```
brenetic, line 27: this may or may not need a bit of tweaking, depending on whether the backend forces the checkbox to be ticked or not. 

when a user is served an error back the expander correctly opens the hidden field, but then when the user goes to click the checkbox it closes the field again and it is not possible to open it after that

app/models/reference_payment_settings.rb
```diff
@@ -31,4 +33,16 @@ def payment_link_url_present?
   def payment_link_checked?
     payment_link == '1'
   end
+
+  def saved_payment_link_url
+    return ServiceConfiguration.find_by(service_id: service_id, name: 'PAYMENT_LINK')&.decrypt_value if payment_link_url.nil?
```
brenetic, line 38:
```suggestion
    return ServiceConfiguration.find_by(service_id: service_id, name: 'PAYMENT_LINK')&.decrypt_value if payment_link_url.blank?
```

`blank?` will also cover an empty string as well as nil

app/models/reference_payment_updater.rb
```diff
@@ -3,7 +3,7 @@ class ReferencePaymentUpdater
 
   attr_reader :service, :reference_payment_settings
 
-  CONFIGS = %w[REFERENCE_NUMBER].freeze
+  CONFIGS = %w[REFERENCE_NUMBER PAYMENT_LINK].freeze
```
brenetic, line 6: `CONFIGS` is no longer being used

app/models/reference_payment_settings.rb
H3ll3m4: It is nil in case the user input en empty string '' in the payment link url field

PR #1948, opened 2022-12-08
Show autocomplete warning for pages in main flow only

brenetic:
For any pages that have an autocomplete page we show a warning on the Publishing page if the user has not uploaded an autocomplete items for any of the autocomplete components in their form.

The Editor also allows pages to become unconnected (detached) should the user decide to do so. The only pages that are followed by the published form are those that are in the main flow therefore there should be no way for a public citizen to reach the unconnected page, unless they magically know its' URL.

Therefore there is no need to show a publishing warning for any autocomplete pages without autocomplete items that do NOT exist in the main flow.

To achieve this, pass the MetadataPresenter::Grid object into the AutocompleteItemsPresenter as the Grid object knows which pages are in the main flow.

Example below:

Screenshot 2022-12-08 at 16 48 07

`Autocomplete Question` page is unconnected, therefore should not block publishing

### DIFF
--- a/acceptance/features/publishing_page_spec.rb
+++ b/acceptance/features/publishing_page_spec.rb
@@ -4,8 +4,9 @@ require_relative '../spec_helper'
 feature 'Publishing' do
   background do
     given_I_am_logged_in
-    given_I_have_a_service_fixture(fixture: 'default_new_service_fixture')
+    given_I_have_a_service_fixture(fixture: fixture_name)
   end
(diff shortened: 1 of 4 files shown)


chrispymm (dismissed):
🚀

njseeto (approved):
🚀

njseeto (approved):
🚀

chrispymm (approved):
🚀

PR #1951, opened 2022-12-09
Editable element js tests

chrispymm:
Add tests for editable components `EditableBase` and `EditableElement` class.


brenetic (approved):
🚀

PR #1953, opened 2022-12-12
Adding payment link placeholder in content substitutor class

njseeto:
In the confirmation email body, we now have a `{{reference_payment_placeholder}}`.

Depending on whether reference number and payment links are enabled,  the placeholder will be swapped out for one of the following:
- Both payment link and reference number sentences
- Just the reference number sentence
- An empty string

If payment link is enabled, the resulting sentence will include `{{payment_link}}` which will be swapped out for the `PAYMENT_LINK` value (ie: the actual payment link url) in the Runner.

### Confirmation email body example:
Screenshot 2022-12-12 at 14 33 54


app/models/content_substitutor.rb
```diff
@@ -89,6 +99,10 @@ def body_content
     @body_content ||= I18n.t('default_values.reference_number_sentence')
   end
 
+  def reference_payment_body_content
+    @reference_payment_body_content ||= I18n.t('default_values.reference_payment_sentence')
```
brenetic, line 103:
```suggestion
    return body_content unless payment_link_enabled
    
    I18n.t('default_values.reference_payment_sentence')
```

app/models/content_substitutor.rb
```diff
@@ -27,11 +29,19 @@ def confirmation_email_body
...
+        placeholder: REFERENCE_PAYMENT_PLACEHOLDER,
+        content: reference_payment_body_content
+      )
+    else
+      substitute_placeholder(
+        setting: setting,
+        placeholder: REFERENCE_PAYMENT_PLACEHOLDER,
+        content: body_content
+      )
+    end
```
brenetic, line 44:
```
      substitute_placeholder(
        setting: setting,
        placeholder: REFERENCE_PAYMENT_PLACEHOLDER,
        content: reference_payment_body_content
      )
```

maybe let the `reference_payment_body_content` method decide what the content should be?

app/models/content_substitutor.rb
brenetic: only used once i think so might not be necessary to memoize it

brenetic (approved):
🚀

PR #1958, opened 2022-12-13
Styling on confirmation page when payment link is enabled

njseeto:
When payment link is enabled, we want the confirmation page to look different as it will signify the user still needs to action the actual payment.

Bump to presenter v2.17.28

### Screenshot
Screenshot 2022-12-14 at 12 20 32


brenetic (approved):
🚀

PR #1963, opened 2022-12-14
Remove feature flag for payment link

H3ll3m4:
### Update content
Final content updates - this includes changes to links, headings and descriptions. Since these changes will affect the platform if deployed to the main branch, we have opted to add this to a testable branch rather than use more feature flags in the templates.
Update the acceptance test to reflect these new content changes.

### Ensure payment url link is stripped of white space
We want to silently remove the white space in the a payment link url. Previously, we were only removing it on the client side but still saving the white space in our database.
This commit, fixes this issue and ensures we strip the whites pace in the controller before saving it to the database.

### Clean up validators
Since we now only trigger validations if payment link has been checked, we no longer need to check for this within each if statement.

### Remove payment link feature flag


brenetic:
can we put [0b4ac23](https://github.com/ministryofjustice/fb-editor/pull/1963/commits/0b4ac235c1922c6b7fa5a9b430d0979bf6f44013) into a separate PR please

then the team can test everything while the feature flag is still enabled

njseeto:
> can we put [0b4ac23](https://github.com/ministryofjustice/fb-editor/pull/1963/commits/0b4ac235c1922c6b7fa5a9b430d0979bf6f44013) into a separate PR please
> 
> then the team can test everything while the feature flag is still enabled

We can do this, but we will need to add a few more feature flags as the content changes affect headings/links/descriptions that are already in live. So thought we could get the team to test on the testable branch and only merge this in when we are happy for release. Happy to chat over a call

brenetic (approved):
🚀

PR #1964, opened 2022-12-14
Add reference param onto payment link when publishing

brenetic:
We do not want to show the user the '?reference=' request parameter in the field on the settings page. So we leave it as the value that the user initially entered and then only add the request parameter when we publish the form.


H3ll3m4 (approved):
🚀

PR #1966, opened 2022-12-14
Inject a tag into confirmation email body

brenetic:
When a user has enabled the confirmation email with payment link then we need to inject an a tag into the body so that it is clickable once it has been processed by the submitter.

The runner will gsub both the {{payment_link}} placeholders before submitting the payload.


njseeto (approved):
🚀

PR #1967, opened 2022-12-15
Inject line break tags into email bodies

brenetic:
Both the service email body and the confirmation email body need to have the line breaks and carriage returns replaced by <br /> tags so that the correct formatting is shown when they are sent by AWS SES.


chrispymm (dismissed):
🚀

njseeto (approved):

PR #1969, opened 2022-12-15
Escape double inverted commas in confirmation email a tag

brenetic:
This is required to keep valid yaml syntax especially for when a form name has a single inverted comma in it. e.g 'a123's form'


chrispymm (approved):
🚀

PR #1970, opened 2022-12-15
Add in new arrow if first condition branching destination is in lower row

chrispymm:
The branching condition arrow layout code assumed that an arrow for the first condition of a branching point would always be going up if it wasn't going stright forwards (on the same row)

This turned out not to be a correct assumption, which was resulting in missing arrows in the COP form.

This PR adds in the missing arrow code.  I'm aware that this fix solves this specific case.  It is hard to know if this will fix every case!  The arrow used assumes the destination will be in the next column so will only have to go down.  This also may be an incorrect assumtion (I don't know)  but for now, it resolves the issue at hand.

**Before:**
image

**After:**
image

### DIFF
--- a/app/javascript/src/controller_services.js
+++ b/app/javascript/src/controller_services.js
@@ -632,7 +632,6 @@ function applyBranchFlowConnectorPaths(view, $overview) {
           };
 
       if(backward || sameColumn) {
-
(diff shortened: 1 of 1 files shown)


brenetic (approved):
🚀

PR #1971, opened 2022-12-15
Flow view layout fixes

chrispymm:
This PR adds fixes for 2 small things on the flow view.

### Arrows overflowing the top of the view
Arrows were going out of viewport at the top of the flow.
The `adjustOverflowHeight()` function shoudl have been correcting for this, however it was using `margin-top` to adjust the position.  Since that method was implemented, the title section of the page has been adjusted to use `position: fixed` meaning that the `margin-top` was not being 'measured' from the title element anymore.  Adjusting to use `top` offsets it from its container - resulting in the correct offset.`

**Before:**
image

**After:**
image

### Add margin between flow-detached-groups
The main flow view has a margin bottom to separate the 'Unconnected section' however each flow group within the unconnected section had no margin/spacing applied and so could overlap.  Adjusted CSS to apply the saem margin to detached flows as to the main flow.

**Before:**
image

**After:**
image

### DIFF
--- a/app/javascript/src/controller_services.js
+++ b/app/javascript/src/controller_services.js
@@ -351,6 +351,7 @@ function adjustOverviewHeight($overview) {
   // Move the overview area down to avoid paths overlapping content above.
   if(topOverlap > 0) {
     $overview.css("margin-top", topOverlap + "px");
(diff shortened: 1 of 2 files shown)


brenetic (approved):
🚀

PR #1972, opened 2022-12-15
Put back the curly quotes into email bodies

brenetic:
The curly quotes are necessary because if there is an apostrophe in the
form name it will confuse the kubernetes cli while it is creating the
config map.

Remove space in email subject as there was one space too many in the email subject lines in the locale
file.

Co-authored-by: Chris Pymm <chris.pymm@digital.justice.gov.uk>


chrispymm (approved):
🚀

PR #1973, opened 2022-12-15
Add in arrow if branch destination is down and not in the next column

chrispymm:
Another fix for an issue on the COP Form.

When the destination page of a branch was in a lower row (down) and not in the next column, the result was an arrow drawn through a page.

**Issue:**
image

Add in the missing case within the `applyBranchFlowConnectorPaths()` method.

**Result:**
image

In the specific case shown above, it results in a slightly ridiculous arrow, but this is because the arrows don't know what they have to go around/avoid so branching conditions always go up the the top and back down when skipping columns.


brenetic (approved):
🚀

PR #1974, opened 2022-12-15
Limit service name to 57 characters in length

brenetic:
Until such time as we can separate out the URL from the service name we need to limit the service name length to 57 characters.

This is because Kubernetes has a limit of 63 characters for URL chunks as well as certain labels that are used throughout each Form's configuration.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

On top of that Cloud Platform sometimes prefix `cname-` before the hostname that is generated by the Editor therefore we can only really allow 57 characters in total.

Once we have a separate service_slug field in the metadata we can up this constraint to something a bit more reasonable.

It is necessary to change the Acceptance Test form names to not use a
uuid as that makes them way too long.

Screenshot 2022-12-19 at 14 35 56
Screenshot 2022-12-19 at 14 36 20

### DIFF
--- a/acceptance/features/create_service_spec.rb
+++ b/acceptance/features/create_service_spec.rb
@@ -99,13 +99,13 @@ feature 'Create a service' do
 
   def then_I_should_see_a_validation_message_for_min_length
     expect(editor).to have_content(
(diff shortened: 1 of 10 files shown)


chrispymm (dismissed):
🚀

njseeto (approved):
🚀

PR #1975, opened 2022-12-15
Remove and recreate the email identity on AWS when it becomes 'unverfied'

njseeto:
[Trello](https://trello.com/c/BseYevGm/3164-from-address-not-showing-as-validated)

This is to address a bug whereby a user adds a from address email and the email is not verified within the 24hr time frame. This sets the identity status on AWS dashboard to 'unverified'.
When the user attempts to save the same email address again, it does not fall into any of the conditions in the `verify_email` method. This is because:
- `get_email_identity` will return something, as the email identity has been created before
- `sending_enabled` will be false, as the email was never verified.

Since it does not fall into either condition, the status will become `nil` in our database, which in turn will not trigger the notification banner in the templates. Meaning the user cannot ever have the ability to resend the validation link. To re-trigger a verification email we have to delete the email identity and create a new email identity.

This commit fixes this by adding an extra condition, that checks if the email identity is present and has not yet been verified. In this case we ensure the status is `pending`.

### DIFF
--- a/app/services/from_address_creation.rb
+++ b/app/services/from_address_creation.rb
@@ -32,11 +32,19 @@ class FromAddressCreation
         "Updated from address status for service #{from_address.service_id} to verified"
       )
       :verified
+
+    elsif email_service.get_email_identity(from_address.email).present? && !email_identity.sending_enabled
+      resend_validation
+
+      Rails.logger.info(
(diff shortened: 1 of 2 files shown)


chrispymm (approved):
👍 LGTM

brenetic (approved):
🚀

PR #1977, opened 2022-12-16
Payment link url validation

njseeto:
### Do not save `payment_link_url` to the DB if payment_link is disabled
When payment_link is not checked we need to remove the record from the` ServiceConfiguration` table to ensure this is not injected during the publishing mechanism.

### Only validate when payment link is enabled
We only want to validate the record if payment link is enabled, this is so validation errors do not trigger if there is a pre-existing `payment_link_url` in the input filed but the user then unchecks the payment link box.

At the moment, when a user enables payment link but leaves the url input field blank, two validations trigger. One for a blank input field and another for the incorrect url format. We only want to have the error for blank input trigger in this instance.

Screenshot 2022-12-16 at 11 58 58


chrispymm (approved):
All looks good

PR #1978, opened 2022-12-16
Reference number and payment link acceptance tests

njseeto:
This PR has tests for:
- Happy path for reference number
- Confirmation email notification when confirmation emails are enabled in one or both environments
- Happy payment link path (payment and reference checked and url valid)
- Payment link validations (invalid url, blank url, reference number not checked)
- Confirmation email body content is correct depending on whether reference number or payment link is enabled
- Ensure check boxes for both reference number and payment link persist when user saves the page
- Confirmation page styling and content changes (payment link enabled vs only reference number enabled)


brenetic (approved):
🚀

PR #1979, opened 2022-12-16
Add continue to pay button customisations

chrispymm:
Bumps metadata presenter to 2.17.29 to include continue to pay button (if enabled)

Adds customisations to the confirmation page edit screen:
* Place the 'Continue to pay' button after the 'Add content' button
* Disable the link - for mouse users with `pointer-events: none` for keyboard users with `tabindex="-1"` 

image


njseeto (approved):
🚀

PR #1981, opened 2022-12-16
Do not clear the payment link url field when a user unchecks the payment link checkbox

njseeto:
### Add payment link column to `SubmissionSettings` table
We need to add a `payment_link` submission setting to enable us to keep the payment link url in the database and to keep track of whether the user has enabled or disabled payment links.

### Update record with payment link Submission Setting
We now have a `payment_link` submission setting, this will need to be updated in the database when a user enables/disables payment link checkbox.

### Update the reference payment validator
We now have a guard clause and only need to run the validator if payment link is checked.
We also only want to show one validation error when payment link is enabled but the payment link url field is blank instead of showing two errors.

### Keep `PAYMENT_LINK` service configuration
We do not want to remove the `PAYMENT_LINK` service configuration from the database as we now need to show this value back to user when they uncheck the payment link checkbox.

### Do not inject payment link service config when disabled
Since we no longer delete the 'PAYMENT_LINK' service config we need to ensure we do not inject this config into the Runner on publishing if `payment_link` submission setting is disabled.

### DIFF
--- a/app/models/reference_payment_settings.rb
+++ b/app/models/reference_payment_settings.rb
@@ -40,8 +40,10 @@ class ReferencePaymentSettings
     valid_payment_link_url
   end
 
-  def payment_link_has_been_checked
-    return ServiceConfiguration.exists?(service_id: service_id, name: 'PAYMENT_LINK') if payment_link.blank?
+  def payment_link_has_been_checked?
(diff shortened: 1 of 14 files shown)


brenetic (approved):
🚀

PR #1986, opened 2022-12-19
Use SubmissionSetting record to determine whether payment link is enabled

njseeto:
### Use payment_link `SubmissionSetting` in `ContentSubstitutor` class
Now that we will always have a 'PAYMENT_LINK' ServiceConfiguration we will need to use the SubmissionSetting record in the ContentSubstitutor helper to check whether `payment_link` is actually enabled. This will determine whether or not we should show the `{{payment_link}}` placeholder in the confirmation email body.

### Update Application Controller `payment_link_enabled?` method
Now that we use the SubmissionSetting to understand whether payment link is enabled, we need to reflect this in the Application Controller method.
This method indicates whether we should show the payment link styling on the Confirmation Page.

### SubmissionSetting requires a deployment_environment
This fixes a small bug whereby the `payment_link_has_been_checked?` method returned false consistently. This is because the SubmissionSetting class requires a deployment_environment.
We only check one environment (`deployment_environment: 'dev'`) as both environments (dev and production) will have records created and removed at the same time.
Also fix up some tests that were using the ServiceConfiguration rather than the SubmissionSetting record in determining whether `payment_link` was enabled.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -92,7 +92,10 @@ class ApplicationController < ActionController::Base
   end
 
   def payment_link_enabled?
-    payment_link_config.present?
+    SubmissionSetting.find_by(
+      service_id: service.service_id,
+      deployment_environment: 'dev'
+    ).try(:payment_link?)
   end
   helper_method :payment_link_enabled?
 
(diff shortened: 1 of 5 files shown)


chrispymm (approved):
🚀

brenetic (approved):
🚀

PR #1987, opened 2022-12-19
Amend payment link url to return blank anchor link

chrispymm:
Updates the payment_link_url method to return a blank anchor link as we don;t have a valid reference number - so there's no point sending them to GOV.UK Pay.


brenetic (approved):
🚀

njseeto (approved):
🚀

PR #1990, opened 2022-12-20
Correctly inject S3 credentials into the Editor deployment

brenetic:
We now inject the AWS S3 credentials and bucket names for all the environments using the secrets configuration.

We can remove the need for the 'TEST' and 'LIVE' from the injected environment variable names. We only need 'DEV' and 'PRODUCTION' in order to distinguish between the different bucket and which credentials to use during publishing.


njseeto (approved):
🚀

PR #1997, opened 2022-12-29
Testable increase container resources

brenetic:
This adds the resource request limits configuration.
The fb-editor-deploy repo has it for Test and Live.
If it is not there then do not set it at all.


H3ll3m4 (approved):
🚀

PR #2029, opened 2023-01-23
Move allowed domains into a custom config

njseeto:
We have two domain allow lists that hold the same email domains. Since we are about to add a third domain level allow list for the send to feature, that will also hold the same domains, we should move the list into a custom config. This means we will have a single domain allow list and reduce code repetition.

This commit moves the domains to a custom config and changes the Auth0 and FromAddress classes to reference the new custom config.

Note: the 'cps.gov.uk' domain has been removed as this was used for testing and is no longer required on the allow list.


H3ll3m4 (approved):
🚀

PR #2032, opened 2023-01-24
Allow time for autocomplete page to load in acceptance test

njseeto:
This acceptance test is flaky and fails regularly, with an error stating the expected text cannot be found.
This could be because the test is attempting to click on the autocomplete drop down box before the options have completely loaded.
Add a sleep to allow the page to load before running the assertion.


H3ll3m4 (approved):
🚀

PR #2035, opened 2023-01-25
Improve flow view render performance 🏎

chrispymm:
This turned into a bit of a beast of a PR (sorry). Hopefully each commit within should be fairly understandable.

These changes make several improvements to how the flow view is rendered in order to improve the time it takes to display the flow.  Large forms, such as the COP form could take up to 15 seconds (after the DOM has loaded) to process and render the flow view to screen.

## Use intersection checking on the lines before we do overlap comparison
Previously every single line was checked against every single other line in the form to see if there were overlaps and things needed to be nudged.  As soon as a line is nudged, we need to start the process over again.  On large forms many lines may have bounds that do not intersect, so there is no way their lines can overlap.  By checking for bounds intersection first, we can prevent a bunch of unnecessary line comparison.

## Reducing re-layout in the browser
Accessing CSS properties from JS forces the browser to re-layout (it needs to re-calculate the position of everything on the screen in order to know the information you want) this is expensive and slow, doing a lot of re-layout is called "layout thrashing" and it is bad.

image
image
Every bit of red in the screenshot above is the browser recalculating layout 🙁

One of the biggest issues was that we were causing re-layout when looping over the arrows, and the flow items.

### Batch rendering arrows
Instead of inserting the arrows into the DOM within a loop, and then later doing all the adjustments. We create them without attaching them into the DOM, store them in arrays, and do all of our adjustments on them, and insert them into the DOM in one operation at the end.

### Batch rendering of flow items.
These items start in the DOM, so we detach them from the DOM before we do our positioning of them, and then re-insert them at the end when we're done.

### Batch render Page Menus and Connection Menus
Similar approach here - separate the creation and rendering step. However as these menus are sued elsewhere, where we are less concerned about performance and can render them immediately, we add a default config option for `render` which defaults to `true`.

### Reduce reading values that cause re-layout 
Anywhere the JS calls `.position()`, `.offset()`, `.css`, `.width()` or '.height()` will cause re-layout.  
* In many places these calls could be refactored out.  In some places we were querying for the width or height of a flow item - these are fixed numbers now, so we can replace them with constants. 
* In others, we were wanting the offset position of the flow item - which had already been calculated earlier in the process. So we now store the calculated position on the object itself and query from there when we need it, not from the DOM.

### Fixing the DownForwardDownBackwardUp arrow
This arrow seemd to cause some issues and wasn't being positioned correctly (the final up line wasn't going to the correct place).  Adjusted the calculation of the line to use the already calculated down lengths.  In order to test this, a `dimensions` property getter was added to the class so that we could test the individual dimensions, rather than testing the obscure output of the SVG path.<sup>*</sup>

### Fix the top offset to include the position of long branch expressions.
If a condition expression is super long, and on the first row of the form, it could overflow the top of the form view.  The code was supposed to do this already, however the way the expressions were positioned had changed so it no longer worked correctly.  Fixed this, and improved the code to only look at expressions on the first row of the form. 

## Final result
image
image
Much less red 🙂

### Before:
**DOMContentLoaded:** 3588ms
**OnLoad:** 9737ms
**Δ DCL-OL:** 6209ms
**Total Blocking Time:** 7276ms

### After:
**DOMContentLoaded:** 3317ms
**OnLoad:** 4429ms
**Δ DCL-OL:** 1112ms
**Total Blocking Time:** 1767ms

All measurements an average of 4 page loads.

**DOMContentLoaded:** When the DOM has been loaded, parsed and is ready.
**OnLoad:** When all loading and scripting is complete.
**Total Blocking Time:** Amount of time the JS is blocking the main thread (meaning the page is not fully interactive)

These aren't the best metrics, but the old version of the page wasn;t loading well enough for devtools to report First Contentful Paint / Largest Contentful Paint metrics.

**We've shortened the OnLoad event firing by 5 seconds!** 

Worth noting that `onLoad` is a late metric - the page will have rendered and been interactive before this.  However in the old version of the form `onLoad` was roughtly equivalent to the flow view rendering, whereas in the new version the flow view is rendered ~0.5 seconds prior to `onLoad`.

<sup>*</sup> A further PR will be needed to add dimensions tests to the other line test files.

### DIFF
--- a/app/javascript/src/component_flow_condition_item.js
+++ b/app/javascript/src/component_flow_condition_item.js
@@ -28,9 +28,6 @@
  **/
 class FlowConditionItem {
   constructor($node, config) {
-    $node.data("instance", this);
-    $node.addClass("FlowConditionItem");
-
     this.$node = $node;
     this.$from = config.$from;
     this.$next = config.$next;
@@ -42,6 +39,10 @@ class FlowConditionItem {
       x2: config.x_out,
       y2: config.y_out
     };
+
+    $node.data("instance", this);
+    $node.addClass("FlowConditionItem");
+    $node.attr('data-row', this.row);
   }
 }
 
--- a/app/javascript/src/component_flow_condition_item.js
+++ b/app/javascript/src/component_flow_condition_item.js
@@ -40,6 +40,11 @@ class FlowConditionItem {
       y2: config.y_out
     };
 
+    this.position = {
+      left: this.bounds.x1,
+      top: this.bounds.y1,
+    }
+
     $node.data("instance", this);
     $node.addClass("FlowConditionItem");
     $node.attr('data-row', this.row);
--- a/app/javascript/src/component_flow_condition_item.js
+++ b/app/javascript/src/component_flow_condition_item.js
@@ -36,6 +36,12 @@ class FlowConditionItem {
     this.$next = config.$next;
     this.row = config.row;
     this.column = config.column;
+    this.bounds = {
+      x1: config.x_in,
+      y1: config.y_in,
+      x2: config.x_out,
+      y2: config.y_out
+    };
   }
 }
 
--- a/app/javascript/src/component_flow_connector_path.js
+++ b/app/javascript/src/component_flow_connector_path.js
@@ -314,7 +314,7 @@ class ForwardPath extends FlowConnectorPath {
     this.#calculatePath(dimensions);
 
     this.type = "ForwardPath";
-    this.build();
+    // this.build();
   }
 
   get path() {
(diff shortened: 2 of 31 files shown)


StevenLeighton21 (approved):
As we talked about, it's hard for me to offer meaningful code review but this looks really good and the testing/stats look great!

StevenLeighton21 (approved):

njseeto (approved):
Just had a read through, similar to Steven hard for me to give a super in depth review but from reading through the commits, the tests and the explanations, nothing glaring has stuck out. It all looks good! 🚀

PR #2036, opened 2023-01-25
Send To

njseeto:
### Domain validator for send to
We want to ensure the email address provided for send to has a domain that is on our allow list. This is to ensure user's submissions are being sent to a trusted inbox.
If it is on the allow list then we can save it to the ServiceConfiguration table. If it is not on the allow list, we do not want to save it and we should show an error message.

### Template updates
- Update the Email Settings form template to show the warning, also ensure the warning links to the correct input field.
- Add content changes to the Email Settings form

### Acceptance tests
- Update existing acceptance test to use a domain on the allow list
- Add Send To acceptance tests which test when the domain is valid and invalid

Screenshot 2023-01-25 at 15 20 57


H3ll3m4 (approved):
🚀

app/views/settings/email/_email_fieldset.html.erb
```diff
@@ -18,13 +18,25 @@
...
+        <% end %>
+      <% end %>
+      <%= f.govuk_email_field :service_email_output,
+        label: {
+          text: f.object.class.human_attribute_name(:service_email_output),
+          for: settings_field_id(f, :service_email_output, deployment_environment)
+        },
+        hint:{ text: t('activemodel.attributes.email_settings.service_email_output_hint') },
+        width: 'two-thirds',
+        id: settings_field_id(f, :service_email_output, deployment_environment) %>
```
chrispymm, line 37: OK, so this is a little complicated.  But because this block is used twice in the page for diffrent envs we need the `id`s of things to distinguish between them.  Because this uses the GOVUK formbuilder gem, its generating an id for the hint for us, using the field key (`:service_email_output`) which produces an `id` for the hint of 'email-settings-service-email-output-hint' which is resulting in duplicate `id`s on the page.

To resolve this I think, we need to use a proc for the hint, as when we do that we are responsible for generating all the html and an `id` isn't autogenerated.
See the docs: https://govuk-form-builder.netlify.app/introduction/labels-captions-hints-and-legends/#input-erb-setting-hint-content-with-a-proc
and the source code for the hint: https://github.com/DFE-Digital/govuk-formbuilder/blob/main/lib/govuk_design_system_formbuilder/elements/hint.rb

```suggestion
        width: 'two-thirds',
        id: settings_field_id(f, :service_email_output, deployment_environment),
        hint: -> do %> 
        <div class="govuk-hint" 
             id="email-settings-service-email-output-<%=deployment_environment%>-hint">
             <%=t('activemodel.attributes.email_settings.service_email_output_hint') %>
        </div>
      <% end%>
```

chrispymm (changes requested):

app/views/settings/email/_email_fieldset.html.erb
njseeto: Nice thanks Chris, this seems to have worked:

### Before
Screenshot 2023-01-27 at 15 18 30

### After:
Screenshot 2023-01-27 at 15 17 26

Screenshot 2023-01-27 at 15 16 51

app/views/settings/email/_email_fieldset.html.erb
```diff
@@ -38,7 +50,6 @@
       text: f.object.class.human_attribute_name(:service_email_body),
       for: settings_field_id(f, :service_email_body, deployment_environment)
     },
-    hint: { text: f.object.class.human_attribute_name(:service_email_body_hint) },
     id:  "service-email-body-#{deployment_environment}-field",
```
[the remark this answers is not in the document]
njseeto: 👍 all done

chrispymm (approved):

PR #2037, opened 2023-01-25
Update confirmation page thumbnail background color when payment link…

chrispymm:
When payment links are enabled the application confirmation page has a blue header instead of green.

This PR updates the flow view thumbnail to reflect this and show a blue background to match.

image
Without payment links enabled

image
With payment links enabled


H3ll3m4 (approved):
🚀

PR #2043, opened 2023-01-30
Update brakeman ignore

njseeto:
We have a EOLRuby warning that is preventing us from deploying the app to production.
This warning tells us to upgrade the app to Ruby 3. We have now prioritised this work, so for the time being ignore this warning.
Ignoring this warning will allow us to continue deploying the app to production.


H3ll3m4 (approved):
🚀

PR #2044, opened 2023-01-30
Reduce resize timeout value

chrispymm:
Now that the flow view is 🏎super-fast we don't need the super long timeout on resizing the screen.  Reduced it down to 100ms, which is pretty much instantaneous from a user perspective.


H3ll3m4 (approved):
🚀

PR #2047, opened 2023-01-30
Make user search work in the admin dashboard

chrispymm:
Makes the search work for users in the admin area.

This should just work out-of-the-box with Administrate (the admin gem we use).  A bit of digging uncovered the reason why it doesn't work - the values in our database are encrypted!
This means it's not possible to do an `ILIKE` search of the column values.

Our data is encrypted by rails before insertion and decrypted on attribute access in rails.  this means Postgres has no knowledge of the data and no possible way to search the data before return from the database.

This PR is a solution to the problem that works, it's definitely no elegant, and would have issues with large datasets, but it works for our purposes.

We override the Administrate index method and 

* Force administrate not to search
* If a search term is present, search the returned data using ruby `select`
* Otherwise, paginate

We then have to override the `users#index` view as the local `resources` variable can now be an `Array` after searching, and calling `paginate` on an array throws an error.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -18,6 +18,33 @@ module Admin
     # Overwrite any of the RESTful controller actions to implement custom behavior
     # For example, you may want to send an email after a foo is updated.
     #
+    def index
+      authorize_resource(resource_class)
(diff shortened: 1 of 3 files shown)


H3ll3m4 (approved):
🚀

PR #2050, opened 2023-01-31
Add rake task that removes services in the 'dev' deployment env

njseeto:
This rake task will remove all services in the 'test-dev' and 'live-dev' namespaces, except any automated test services (acceptance and smoke tests).

It is to be run manually, via the command line, this is because we will (hopefully) only need to run this very rarely.
This rake task has been created as we need to remove published services in 'live-dev' and 'test-dev' due to the need to re-publish all forms as a result of the CircleCi key rotation.

To run: `bundle exec rake unpublish:dev_services`

### DIFF
--- a/app/services/unpublish_dev_services.rb
+++ b/app/services/unpublish_dev_services.rb
@@ -1,0 +1,33 @@
+class UnpublishDevServices
+  include MetadataVersionHelper
+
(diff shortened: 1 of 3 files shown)


H3ll3m4 (approved):
🚀

PR #2053, opened 2023-02-01
Only try to verify email address if not already verified on save

StevenLeighton21:


app/services/from_address_creation.rb
```diff
@@ -6,6 +6,7 @@ class FromAddressCreation
 
   def save
     return if from_address.invalid?
+    return if FromAddress.find_by(service_id: from_address.service_id).email_address == from_address.email_address
```
njseeto, line 9:
```suggestion
   def saved_from_address 
     @saved_from_address ||= FromAddress.find_by(
        service_id: from_address.service_id
      ).email_address
    end
```
Should we move the left half to a private method and memoize the call to the DB?

njseeto (commented):
Looks good, just one comment. Also did you want to add the changes to the `ruby-version` and `Gemfile.lock` files to this PR?

njseeto (approved):
🚀

PR #2054, opened 2023-02-01
Add re-publish functionality to admin dashboard

njseeto:
This functionality will re-publish the most recently published version of a service.

This does this by creating a new `PublishServiceCreation` object, which takes in the `ServiceConfiguration` for 'BASIC_AUTH_USER' and 'BASIC_AUTH_PASS' of the service_id, if these exists (ie: if the existing services requires a username/password to access it).

We need to create a new `PublishServiceCreation` as this object understands whether a particular service requires authentication, and it can retrieve the a username/password from the `ServiceConfiguration` table, but it does not hold onto this information after a form is published so we cannot just re-use the last `PublishServiceCreation` object.

Similar to the Unpublish button, the re-publish button will only appear if there is a service that has been published:

Screenshot 2023-02-02 at 09 04 40

### DIFF
--- a/app/controllers/admin/services_controller.rb
+++ b/app/controllers/admin/services_controller.rb
@@ -112,6 +112,35 @@ module Admin
       redirect_to admin_service_path(service_id)
     end
 
+    def republish
+      if queued?
+        flash[:notice] = "Already queued for republishing from #{params[:deployment_environment]}"
+      else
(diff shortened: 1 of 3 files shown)


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2063, opened 2023-02-06
Fix issues with from_address duplication check

StevenLeighton21:
Update tests to test behaviour more fully, add a guard for when no saved record could be found, move the check to prevent setting the status to pending if unchanged

### DIFF
--- a/app/services/from_address_creation.rb
+++ b/app/services/from_address_creation.rb
@@ -7,6 +7,8 @@ class FromAddressCreation
   def save
     return if from_address.invalid?
 
+    return if saved_email_address.present? && saved_email_address == from_address.email
+
     from_address.status = verify_email
     from_address.save!
   rescue ActiveRecord::RecordInvalid
@@ -18,7 +20,6 @@ class FromAddressCreation
 
   def verify_email
     return :default if use_default_email?
-    return if saved_from_address == from_address.email
 
     identity = email_service.get_email_identity(from_address.email)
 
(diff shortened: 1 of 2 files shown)


app/services/from_address_creation.rb
```diff
@@ -78,10 +79,15 @@ def email_identity
...
+  def saved_email_address
     @saved_from_address ||= FromAddress.find_by(
       service_id: from_address.service_id
-    ).email_address
+    )
+    if @saved_from_address.nil?
+      return nil
+    end
+
+    @saved_from_address.email_address
```
njseeto, line 90:
```suggestion
    )&.email_address
```

I think the safe navigation operator should work here.

njseeto: Or we could pull out the db call into it's own method:

```
def saved_email_address
  return nil if saved_from_address.nil?
  
  saved_from_address.email_address
end

def saved_from_address
  @saved_from_address ||= FromAddress.find_by(
    service_id: from_address.service_id
  )
end
```

njseeto (approved):
Looks good, just a minor suggestion but happy to leave it as - whichever is more readable to you

app/services/from_address_creation.rb
StevenLeighton21: Going to merge as is- I like the null coalesce but feels more elegant/less readable to me, I find the method readable and want to only return the email from the helper method rather than splitting it into a helper for the saved object then inspecting it or having two helpers, I just put a nil check in because it seems a remotely possible edge case if the form is old

TY!

PR #2065, opened 2023-02-07
Upgrading Presenter version to 2.17.35

H3ll3m4:
We update the version of the presenter to see the update of the cookie statement expiring from now on to 60 minutes


njseeto (approved):
🚀

PR #2072, opened 2023-02-09
Add tests for EditableCollectionFieldComponent methods

chrispymm:
Not much to say beyond the title really.

These tests should cover all public methods, and effectively test all private methods by proxy as the expectations check the results of those methods.


H3ll3m4 (approved):
🚀

PR #2074, opened 2023-02-09
Add tests for editableComponent class instantiator

chrispymm:
Add tests for the `editableComponent` method that is used to instantiate the correct `EditableComponent` class based on the `type` provided in the content.  

Little bit of metaprogramming to generate the tests, as they're just duplicating the same test/expectation with a different type.  Not the biggest fan of this, and wouldn't want to use too much of it, but in this case I think its acceptable as the result is still pretty readable and obvious what is being tested (I hope).


H3ll3m4 (approved):
🚀

PR #2086, opened 2023-02-16
EditableContent component tests

chrispymm:
Adds tests for the editable content component.

Includes a couple of small refactors:

* rename the pseud-private `_content` variable to `#markdown` private property.  This makes it actually private, and better communicates *what* is actually stored in it compared to the `content` property.
* Extract `convertToHtml`, `convertToMarkdown` and `cleanInput` functions to a utilities file. 

The intent with these tests is to test the class itself, and not its specific output/conversion - we're not trying to test the showdown markdown converter!

However, the `cleanInput()` method is doing transformations specific to our implementation and quirks of showdown and sanitise-html so I have added in some tests to esure we are testing for those corrections working.

### DIFF
--- a/test/editable_components/editable_content/properties_test.js
+++ b/test/editable_components/editable_content/properties_test.js
@@ -2,9 +2,7 @@ require('../../setup');
 
 describe('EditableContent', function() {
   const helpers = require('./helpers');
-  const c = helpers.constants;
   const COMPONENT_ID = 'editable-content-properties-test';
-  const COMPONENT_CLASSNAME = 'EditableContent';
 
   describe('Properties', function() {
     var created;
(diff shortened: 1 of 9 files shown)


test/editable_components/editable_content/properties_test.js
```diff
@@ -0,0 +1,41 @@
+require('../../setup');
+
+describe('EditableContent', function() {
+  const helpers = require('./helpers');
+  const c = helpers.constants;
```
StevenLeighton21, line 5: Couldn't see this get used again in the file? Is it needed?

StevenLeighton21 (dismissed):
LGTM!

test/editable_components/editable_content/properties_test.js
chrispymm: Fixed this now - there were a few of these. I was juts copying and pasting  the top of the file each time!

H3ll3m4 (approved):
🚀

PR #2087, opened 2023-02-17
Editable content include conversion methods

chrispymm:
This PR moves `convertToHtml()`, `convertToMarkdown()` and `cleanInput()` methods into the `EditableContent` class.

While working through the testing I initially moved those methods into a utilities file as it felt like thats what they were.

However on reflection, I reailsed that they only get used by the `EditableContent` class, and `cleanInput()` specifically has transformations that are unique to the implementation within that class.

Given that, it felt much more like these methods belonged within the class itslef - the main function of the `EditableContent` class is to take user input as markdown and convert it for both saving to metadata and displaying as HTML, so it feels right that they should belong in here.


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2088, opened 2023-02-17
EditableContent refactor update() method

chrispymm:
This is a minor refactor of the `update()` method on the `EditableContent` class.  I had thought some lerger changes could be made, however there is a reason we call `cleanInput()` when we do, and with both markdown and HTML.

So this is mostly just a few small tweaks that (I hope) improve the readability and understanding of what is happenign in  the method.


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2089, opened 2023-02-17
Api::BranchesController should return 401 if there is no session

chrispymm:
## Issue
If a users session timed out while on the branching screen, and they then performed an interaction that triggered an api request.  That request was returning the login screen html, which was then being inserted into the page - causing it to look very strange.

## Cause
The `Api::BranchesController` - unlike the other controllers within the `Api` namespace - inherits from the main `BranchesController`.  This mean it was calling the main namespace `require_user!` method, which performs a redirect if the user session is no longer active.

## Solution
Add `require_user!` method into the `Api::BranchesController` which returns a JSON response with an unauthorized status. This is duplicated from the main `ApiController`, not overly DRY, but didn't seem worth another mixin/helper.  It seems unlikely that any other api controllers will need to inherit from a class other than the ApiController.

The `updateDOMByApiRequest` utility method was updated to handle reciving a failure response, and then issue a redirect to the login screen.

Som adjustments to tests were required as they were mocking `$.get` so these mocks needed changing to be able to respond to `.fail`

### DIFF
--- a/app/controllers/api/branches_controller.rb
+++ b/app/controllers/api/branches_controller.rb
@@ -14,4 +14,10 @@ class Api::BranchesController < BranchesController
 
     render layout: false
   end
+
+  def require_user!
+    unless user_signed_in?
+      render json: { message: 'Unauthorised' }, status: :unauthorized
+    end
+  end
 end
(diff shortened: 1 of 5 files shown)


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2096, opened 2023-02-22
Remove From Address link

njseeto:
Now that we have implemented the. Reply-To feature, users will no longer need access to the From Address Settings page.
This PR will hide the link behind feature flags (when we remove the feature flags, we can remove the link itself).

Remove acceptance tests that relate to the From Address settings page.


app/views/settings/submission/index.html.erb
```diff
@@ -6,10 +6,12 @@
 
   <nav class="govuk-navigation" aria-labelledby="page-heading">
     <dl aria-labelledby="submission-settings-navigation-heading" class="fb-settings">
+    <% if ENV['REPLY_TO'] == 'disabled' %>
```
StevenLeighton21, line 9: Can we do if !ENV['REPLY_TO'] == 'enabled'?

StevenLeighton21 (approved):

PR #2097, opened 2023-02-23
Reply-To: Template changes to Collect Information by Email settings page

njseeto:
### Update 'collect information by email' settings page
As part of reply-to, we will now use the default no-reply-moj-forms email address as the 'From address'.
So we no longer want users to access the 'From Address' page via the 'Change from address' link, we also do not want to show any warnings related to from address.
This commit removes access to the link and the warnings associated to the from address verification. We now just present the Form name, not the email address.

### Remove failing acceptance tests
Now that we can no longer show any from address warnings on the 'Collect information by email' page, these tests will fail.

#### Before
Screenshot 2023-02-23 at 11 44 31

#### After
Screenshot 2023-02-23 at 11 44 00

### DIFF
--- a/app/views/settings/email/_email_fieldset.html.erb
+++ b/app/views/settings/email/_email_fieldset.html.erb
@@ -1,4 +1,11 @@
 <%= f.govuk_fieldset(legend: { text: f.object.class.human_attribute_name(:email_fieldset) }) do %>
+
+  <% if ENV['REPLY_TO'] == 'enabled' %>
+    <div class="govuk-form-group">
+      <%= f.object.class.human_attribute_name :service_email_from %>
+      <%=f.object.service.service_name %>
+    </div>
+  <% else %>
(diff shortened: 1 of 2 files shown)


app/views/settings/email/_email_fieldset.html.erb
```diff
@@ -1,4 +1,6 @@
 <%= f.govuk_fieldset(legend: { text: f.object.class.human_attribute_name(:email_fieldset) }) do %>
+
+  <% if ENV['REPLY_TO'] == 'disabled' %>
```
StevenLeighton21, line 3: I'd probably do `if <feature flag>  <new code> else <existing code>` as this covers the case where the feature flag is omitted or set incorrectly

njseeto: all done

StevenLeighton21 (approved):
LGTM!

PR #2098, opened 2023-02-23
Remove from address publish warnings for Test and Live environments

njseeto:
Now that email addresses for reply-to will be validated on save, we no longer require this warning.

Hide warnings in the template and only add the from_address warnings if the environment variable is disabled.

#### Before
Screenshot 2023-02-23 at 15 59 54

#### After
Screenshot 2023-02-23 at 15 59 19

### DIFF
--- a/app/presenters/publishing_page_presenter.rb
+++ b/app/presenters/publishing_page_presenter.rb
@@ -73,7 +73,7 @@ class PublishingPagePresenter
 
     return presenters if deployment_environment == 'dev'
 
-    presenters.push(from_address_warning) if service_output_warning.blank?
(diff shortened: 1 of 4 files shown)


app/presenters/publishing_page_presenter.rb
```diff
@@ -73,7 +73,7 @@ def submission_warning_presenters
 
     return presenters if deployment_environment == 'dev'
 
-    presenters.push(from_address_warning) if service_output_warning.blank?
+    presenters.push(from_address_warning) if service_output_warning.blank? && ENV['REPLY_TO'] == 'disabled'
```
StevenLeighton21, line 76: Might be worth doing something like !ENV['REPLY_TO'] == 'enabled' to cover the empty string case, but it's not a huge deal

StevenLeighton21 (approved):

PR #2100, opened 2023-02-27
bump metadata presenter to 2.17.37

chrispymm:
Bumps metadata presenter gem to 2.17.37 to fix issues with the send body content editable field on the check answers page.


H3ll3m4 (approved):
🚀

PR #2112, opened 2023-03-10
Fix interpolation of hint id attribute

njseeto:
The deployment environment was not interpolating correctly in the template, causing an accessibility issue as the hint id was not unique. This commit fixes this.

### Before
Screenshot 2023-03-10 at 10 37 20

### After
Screenshot 2023-03-10 at 10 36 16


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2113, opened 2023-03-10
Remove feature flags

StevenLeighton21:
We only used feature flags in the editor - the runner and submitter were backwards compatible so this is the only place they need to be removed. Live has been been deployed with the new flag so this is just cleanup.


njseeto:
Looks good, just missed on sneaky flag in the `app/views/publish/_publish_environment.html.erb` file.

njseeto (changes requested):
Looks good, just missed one sneaky flag in the `app/views/publish/_publish_environment.html.erb` file.

njseeto (approved):
🚀

PR #2124, opened 2023-03-20
Change editor session to last during working hours

StevenLeighton21:
Update session clear job to delete all sessions, update cron schedule to fire once per day at 20:00

Users will be logged out at 20:00 and will need to log in again when next using the editor, we aim to allow users to log in at the start of the working day then be logged out at 2000 hrs

This lines up with the auth0 session anyway which lasts 10 hours, and alleviates user pain points (not to mention accessibility issues) around being unable to extend your session and being logged out every 90 mins.

We have future work planned soon (this quarter) to introduce RBAC anyway which will invalidate how we currently manage user identity and sessions.

Cron syntax from crontab
image


H3ll3m4 (approved):
🚀

PR #2173, opened 2023-04-25
Testable save return key

StevenLeighton21:
Adds two env vars to the published forms on create/republish -

SAVED_FORMS_KEY - a generated key used as a platform key when encrypting saved progress at rest, kept separate from the submission key just for ease of rotation in the event of compromise.

SAVE_AND_RETURN - a feature flag used to dark launch the save and return functionality


njseeto (changes requested):
Looks good - just the one comment to remind us to delete the config before we merge into main.

StevenLeighton21:
Superseded

PR #2174, opened 2023-04-26
Testable undo redo error modal

chrispymm:
Adds a pretty generic error message in a modal that is displayed if an undo or a redo action fails to complete.

If the action fails, its likely due to an issue with the metadata, and retrying the action is unlikely to succeed so we remove the button on failure.


H3ll3m4 (dismissed):
🚀

njseeto:
Do we have any / do we want to add any acceptance tests for these new modals?

chrispymm:
> Do we have any / do we want to add any acceptance tests for these new modals?

We don't, and we probably should.  I'll add a couple. 👍

njseeto (approved):
🚀

PR #2178, opened 2023-04-27
Testable save and return

njseeto:
### Save and Return Settings page
This PR will introduce two new service configurations to the Editor:
- `SAVE_AND_RETURN`
- `SAVE_AND_RETURN_EMAIL`

`SAVE_AND_RETURN` will be injected and given a value of `1` and `SAVE_AND_RETURN_EMAIL` is the default email content, which is not configurable by the user.
If the Save and Return checkbox is enabled both these configs will be injected into the Runner on publish.
If the checkbox is not checked, the `SaveAndReturnUpdater` will remove the config if it exists - therefore, this config will not be injected into the Runner on publish.

Screenshot 2023-04-27 at 16 04 31


app/models/save_and_return_settings_updater.rb
```diff
@@ -0,0 +1,56 @@
...
+      save_config_save_and_return
+    end
+  end
+
+  private
+
+  def save_config_save_and_return
+    if save_and_return_settings.save_and_return_enabled?
+      create_or_update_service_configuration(config: 'SAVE_AND_RETURN', deployment_environment: 'dev', value: 'enabled')
+      create_or_update_service_configuration(config: 'SAVE_AND_RETURN', deployment_environment: 'production', value: 'enabled')
```
njseeto, line 21: I had to hardcode the value `enabled` in here as this is what the MDP is expecting currently.

We should check for presence of the `SAVE_AND_RETURN` env var in the MDP to be consistent with other checks in the MDP but this will do for now to get it working e2e without having to bump the MDP.

acceptance/features/save_and_return_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    # when editing a page
+    and_I_return_to_flow_page
+    and_I_edit_the_page(url: 'Page b')
+    then_I_should_not_see_save_and_return_button
+
+    # in preview
+    and_I_return_to_flow_page
+    preview_page = when_I_preview_the_page('Page b')
+    # the_I_should_not_see_save_and_return_button
```
StevenLeighton21, line 49: Guess this line can go?

config/locales/en.yml
```diff
@@ -347,6 +348,14 @@ en:
       url: 'https://www.payments.service.gov.uk/'
       checkbox_label: Add payment links with reference numbers
       url_label: Your GOV.UK Pay payment link URL
+    save_and_return:
+      heading: 'Save for later'
+      lede: 'Give users the option of saving a partially completed form.'
+      description: 'Give users the option of saving a partially completed form for up to 28 days.'
+      page_lede: 'This places a save button at the bottom of each page of your form which will lead users through the save process. For details, see the %{href}'
+      href: <a class="govuk-link" href="%{url}" target="_blank" rel="noopener noreferrer">user guide section on save for later.</a>
+      url: 'https://moj-forms.service.justice.gov.uk/settings/'
```
StevenLeighton21, line 357: We wouldn't localise a url? This could just be hard coded in the view, or even better use <%= settings_path %>

app/views/settings/save_and_return/index.html.erb
```diff
@@ -0,0 +1,31 @@
+<%= mojf_settings_screen(
+  heading:t('settings.save_and_return.heading'),
+  description: t('settings.save_and_return.description')
+  ) do |c| %>
+
+  <% c.with_back_link(href: settings_path) %>
+
+<p><%= t('settings.save_and_return.page_lede', href: t('settings.save_and_return.href', url: t('settings.save_and_return.url'))).html_safe %></p>
```
StevenLeighton21, line 8: if this is a link shouldn't it be an <a> tag? with my other comments about the html not being in the localisation file I'd probably rewrite this to be something like `<a href=<%= settings_path %> ...class ... >t('settings.save_and_return.page_lede')</a>`

njseeto: This is here to embed the url into the paragraph, alternatively, I can remove the localisation and add it all in the template.

Currently we render like this:
```
<p><%= t('settings.save_and_return.page_lede', href: t('settings.save_and_return.href', url: t('settings.save_and_return.url'))).html_safe %></p>
```

But can change to this in the template:
```
<p><%= t('settings.save_and_return.page_lede')%><a class="govuk-link" href="https://moj-forms.service.justice.gov.uk/settings/" target="_blank" rel="noopener noreferrer">user guide section on save for later.</a></p>
```

Both will have the same result, it's just what we'd prefer - I'll leave this comment for now and see if the others have a preference.

config/locales/en.yml
njseeto: can do, will add the url to the above `href` key

app/views/settings/save_and_return/index.html.erb
StevenLeighton21: You can do the second option and still have the `user guide section on save for later.` text as localised. It's just a bit more awkward to inject localised links into localised text but we can't have html in the yaml file IMO, it's not something we could give to a translator and it locks us in to that html

StevenLeighton21 (approved):
LGTM

H3ll3m4 (approved):
🚀

PR #2192, opened 2023-05-09
Api JSON Endpoint Submission Settings in Admin Page

H3ll3m4:
## Context
This is part of the work for the forms migration and the need to add the possibility to send submissions to an API as a JSON. The configuration of this JSON endpoint will be done by our team in the admin page by setting 2 values SERVICE_OUTPUT_JSON_ENDPOINT and SERVICE_OUTPUT_JSON_KEY that will then be sent to the runner for submission

## Change
In this PR we're adding a controller for the API submission from the admin page and for that we need to add an api submission dashboard class. We add 2 routes to show current configuration and to create configuration. We validate the endpoint url by using the uri regex and the key length that is 16 character long. We use flash messages to provide a feedback to the user. We decide here to enter empty fields to reset configuration instead of creating a new delete route. 

## Tests
In this PR there are no integration tests file for the controller but only unit test for the models.


app/models/admin/api_submission_settings.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    def service_output_json_key
+      settings_for(:service_output_json_key)
+    end
+    # rubocop:enable Lint/DuplicateMethods
+
+    private
+
+    def is_key_valid_lenght
```
StevenLeighton21, line 30:
```suggestion
    def is_key_valid_length
```

spec/models/admin/api_submission_settings_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+          expect(subject).to_not be_valid
+        end
+
+        it 'should have errors' do
+          subject.valid?
+          expect(subject.errors.first.type).to eq(expected_error)
+        end
+      end
+
+      context 'key lenght is too long' do
```
StevenLeighton21, line 57:
```suggestion
      context 'key length is too long' do
```

spec/models/admin/api_submission_settings_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+      context 'is empty to reset configuration' do
+        let(:service_output_json_key) { '' }
+
+        it 'is valid' do
+          expect(subject).to be_valid
+        end
+      end
+
+      context 'key lenght is too short' do
```
StevenLeighton21, line 44:
```suggestion
      context 'key length is too short' do
```

app/models/admin/api_submission_settings.rb
```diff
@@ -0,0 +1,46 @@
...
+                  :service,
+                  # required for settings creation in controller:
+                  :service_output_json_endpoint,
+                  :service_output_json_key
+
+    validates :deployment_environment, inclusion: {
+      in: Rails.application.config.deployment_environments
+    }
+    validates :service, presence: true
+    validate :is_key_valid_lenght, :is_url_valid
```
StevenLeighton21, line 13:
```suggestion
    validate :is_key_valid_length, :is_url_valid
```

app/views/admin/services/show.html.erb
```diff
@@ -15,6 +15,10 @@
       Visit service in app
     </a>
 
+<% if ENV['API_SUBMISSION'] == 'enabled' %>
+    <%= link_to "Show API Endpoint Submission Settings", admin_service_api_submission_index_path( :service_id => @service.service_id ), class: "govuk-button fb-govuk-button" %>
+<% end %>
+</div>
```
njseeto, line 21:
```suggestion

```

I think this closing div has snuck in here. Removing it will fix the margin issue on the admin page

njseeto (commented):
Looks good, just one small thing from me

PR #2203, opened 2023-05-18
Api submission

H3ll3m4:
## Change
Endpoint key is actually 17 characters long instead of 16 so instead of updating the adapters key, along with the submitter, the tech docs and other places where it is mentioned or used we remove the length restriction for more flexibility.

## Testing
We were able to receive a submission on the base adapter url: 
https://formbuilder-base-adapter-test.apps.live.cloud-platform.service.justice.gov.uk/submission

### DIFF
--- a/app/models/admin/api_submission_settings.rb
+++ b/app/models/admin/api_submission_settings.rb
@@ -9,7 +9,7 @@ module Admin
     validates :deployment_environment, inclusion: {
       in: Rails.application.config.deployment_environments
     }
-    validates :service, presence: true
+    validates :service, :service_output_json_key, presence: true
     validate :is_url_valid
 
     URL_ERROR = 'Endpoint field should be a valid URL. '.freeze
--- a/app/models/admin/api_submission_settings.rb
+++ b/app/models/admin/api_submission_settings.rb
@@ -30,7 +30,7 @@ module Admin
     def is_key_valid_length
       return if @service_output_json_key == ''
 
-      if @service_output_json_key.nil? || @service_output_json_key.length != 16
+      if @service_output_json_key.nil? || @service_output_json_key.length != 17
         errors.add(:base, KEY_LENGTH_ERROR)
       end
     end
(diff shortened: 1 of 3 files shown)


app/views/admin/api_submission/_form.html.erb
```diff
@@ -36,7 +36,7 @@
     width: 'two-thirds',
     id: settings_field_id(f, :service_output_json_key, deployment_environment),
     hint: {
-      text: 'This is the 16 characters key'
+      text: 'This is the 17 characters key'
```
StevenLeighton21, line 39: Maybe the hint shouldn't specify a length? Seems the key could be any length as defined by the target API

app/models/admin/api_submission_settings.rb
```diff
@@ -10,9 +10,8 @@ class ApiSubmissionSettings < BaseEmailSettings
       in: Rails.application.config.deployment_environments
     }
     validates :service, presence: true
```
StevenLeighton21, line 12:
```suggestion
    validates :service, :service_output_json_key, presence: true
```

spec/models/admin/api_submission_settings_spec.rb
```diff
@@ -19,58 +19,8 @@
...
-
-      context 'has to be 16 characters long' do
-        let(:service_output_json_key) { '98957af1a0424376' }
-
-        it 'is valid' do
-          expect(subject).to be_valid
-        end
-      end
-
-      context 'is empty to reset configuration' do
```
StevenLeighton21, line 36: This test was removed but I suggested adding a validation for presence - it is probably worth adding a test that covers the new behaviour, 

''   = now not valid
nil = now not valid

Could also be that my suggestion was bad and '' should be valid, in which case I'd add that test and remove the `validates presence` for the key

H3ll3m4: I tested the case we remove the config because it's the way to delete but I can add a unit test covering that. Thank you.

StevenLeighton21 (approved):

PR #2204, opened 2023-05-18
Add URL validation on form creation

njseeto:
We are separating the form name and the URL. Since the URL will no longer be derived from the form name, we need to create the URL at the point of form creation to present back to the user on the Form Name settings page.

On form creation, we will use the given form name to create a 'SERVICE_SLUG' service configuration. We will be removing most of the validations on the form name, so we will now need to validate the generated SERVICE_SLUG. 

We ensure the SERVICE_SLUG is:
- 57 chars in length
- Stripped of special characters and parameterised
- Checked against existing SERVICE_SLUG and PREVIOUS_SERVICE_SLUG service configs for uniqueness
- Does not begin with a number

If the SERVICE_SLUG is not unique, we replace the last 3 chars with randomised alphanumeric chars.
There is no check against legacy form URLs as we will keep this validation on form creation anyway.

Note: We have the `SERVICE_SLUG` service config  creation feature flagged so it will only be created for new services in the Test Editor

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -10,6 +10,14 @@ class ServicesController < PermissionsController
     @service_creation = ServiceCreation.new(service_creation_params)
 
     if @service_creation.create
+
+      if ENV['NAME_SLUG'] == 'enabled'
+        FormNameUpdater.new(
+          service_id: @service_creation.service_id,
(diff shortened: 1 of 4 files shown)


H3ll3m4 (approved):
🚀

PR #2206, opened 2023-05-19
Add service_slug_config method

njseeto:
This will use the 'SERVICE_SLUG' value in the ServiceConfiguration table. If it does not exist, we parameterize the service_name.

For now, this is to allow for backwards compatibility.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -126,4 +126,11 @@ class ApplicationController < ActionController::Base
     request.script_name.include?('preview')
   end
   helper_method :editor_preview?
+
+  def service_slug_config
+    @service_slug_config ||= ServiceConfiguration.find_by(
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -126,4 +126,11 @@ def editor_preview?
     request.script_name.include?('preview')
   end
   helper_method :editor_preview?
+
+  def service_slug_config
+    @service_slug_config ||= ServiceConfiguration.find_by(
+      service_id: service.service_id,
+      name: 'SERVICE_SLUG'
+    )&.decrypt_value
```
StevenLeighton21, line 134: Does this need to be stored encrypted? Might be easier for us if they're not and we need to search by them. No big deal if it's hard to change though.

StevenLeighton21 (approved):

PR #2219, opened 2023-05-26
Do not create SERVICE_SLUG config if user is Acceptance Tests

njseeto:
We have many `SERVICE_SLUG` records in our ServiceConfig table. This is because our acceptance tests create new forms on every run, so adding this little caveat to prevent the creation of a SERVICE_SLUG record for acceptance tests.

We use the service name as the service slug by default if there is no `SERVICE_SLUG` record.


StevenLeighton21 (approved):
Is it not deleted when the AT form is deleted during the nightly job?

H3ll3m4 (approved):
🚀

PR #2231, opened 2023-06-01
Change Form Name and URL settings template when form is Live

njseeto:
Once a form has been published to live, we do not allow users to change their URL.
Remove the URL input on the template and show the current live URL instead.
We need to include the hidden field, so as not to trigger the URL validation on save.

### When not Live
Screenshot 2023-06-01 at 11 18 01

### When Live
Screenshot 2023-06-01 at 12 45 55

### DIFF
--- a/app/models/form_name_url_settings.rb
+++ b/app/models/form_name_url_settings.rb
@@ -37,6 +37,10 @@ class FormNameUrlSettings
     service_slug
   end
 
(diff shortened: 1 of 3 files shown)


H3ll3m4 (approved):
🚀

StevenLeighton21 (approved):

PR #2234, opened 2023-06-02
Add backup cipher rotator

StevenLeighton21:
We now default to 'aes-256-gcm', where 'aes-256-cbc' is the old SHA1 digest. This will allow decrypt operations to fall back to the old algorithm, and future encrypt operations will continue with the new default.

There is already a cookie rotator in place, but this should prevent issues we've seen with cronjobs unable to decrypt user emails.

### DIFF
--- a/app/services/encryption_service.rb
+++ b/app/services/encryption_service.rb
@@ -21,6 +21,9 @@ class EncryptionService
   private
 
   def encryptor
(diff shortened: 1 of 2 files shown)


StevenLeighton21:
Closing - we'll come back to this if it seems to be needed but it did not affect our issue

PR #2235, opened 2023-06-02
Fix issues with sprockets sass compilation/compression

chrispymm:
A forthcoming change introduces a gem that adds a dependency that adds assets into the Sprockets asset pipeline that cause issues with sass compression with sassc-rails. '

As no other sass processor seems to work, and the assets are not needed at all this fix porovides an initializer that allows for the assets of specified gems to be prevented from being added into the sprockets asset paths.


config/initializers/remove_unrequired_assets.rb
```diff
@@ -0,0 +1,18 @@
+# Some gems add assets into the sprockets asset pipeline
+# that are not needed or cause issues with compilation / compression. This
+# monkeypatch allows us to remove them from the Sprockets paths
+# 
+# Add the mes of any gems whose assets you do not wish to be compiled
```
StevenLeighton21, line 5:
```suggestion
# Add the names of any gems whose assets you do not wish to be compiled 
```

StevenLeighton21 (approved):

PR #2250, opened 2023-06-14
Increase read timeout

njseeto:
We have an error which only affects the read timeout. This has occurred when a user has tried to publish a form unsuccessfully.

60 seconds is the maximum before rails defaults take over


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2251, opened 2023-06-14
Testable form name url

njseeto:
## Remove Service name validations
Now that we no longer use the service name to create the URL, we are removing the restrictions on the form name.

## Split out the `FormUrlCreation` and `FormUrlUpdater`
There are two very different mechanisms when it comes to creating the service slug on form creation and updating the service slug on an already existing form.
For the service creation, the user has no way to specify a service slug, therefore we need to create one from the service name - to make this possible, in the background we silently create a valid url.
When updating a service, the user is able to input a url and therefore the validations can be run on those inputs.
Therefore, we can assume that anything that passes validations on updating the setting is valid.

## Acceptance tests
Adding tests for happy and unhappy paths on the Form Name and URL settings page.
Note, we can not test the publishing mechanism in this manner and hence, the publishing mechanism was tested manually.

## Remove Form name and URL feature flags
This PR, once merged, will remove the feature flags and be deployed to live production.


H3ll3m4 (approved):
🚀

PR #2263, opened 2023-06-22
Testable multi file upload

StevenLeighton21:
Replaces access to the file upload component with a new, extended component

image

It can be one (optional file) or up to 10 files
image

image

image

Updating the max files to greater than one will show the user how many files they can add

image


config/locales/en.yml
```diff
@@ -312,11 +316,16 @@ en:
       max_string_length: 'Maximum answer length...'
       min_string_length: 'Minimum answer length...'
       upload_options: 'Autocomplete options...'
+      multiupload_options: 'Number of files...'
```
njseeto, line 319:
```suggestion
```
I don't think we use this one

app/controllers/pages_controller.rb
```diff
@@ -98,6 +98,33 @@ def common_params
     }
   end
 
+  def multiupload_files_remaining
+    max_files = @page.components.select { |c| c.type == 'multiupload' }.first['max_files'].to_i
```
njseeto, line 102:
```suggestion
    max_files = @page.components.select(&:multiupload?).first['max_files'].to_i
```

app/controllers/pages_controller.rb
```diff
@@ -98,6 +98,33 @@ def common_params
...
+    answered = false
```
njseeto, line 103: can you help me understand what's happening here? I don't fully understand why we are assigning this variable to a boolean and then using it in the if/else statements

njseeto (commented):
Looks good, just a couple of comments and one question.
I can't comment on the validators as I wrote them and the js files look fine to me but might be worth getting Chris to have a look over them.

app/controllers/pages_controller.rb
StevenLeighton21: It wasn't actually being used, it just needed to exist for the template to work on the edit page. Have removed!

njseeto (approved):

PR #2264, opened 2023-06-22
EditableContent webcomponent

chrispymm:
This PR migrates the EditableContent areas to use a custom element.

This allows the code to be a bit more decalrative and hopefully easier to understand/maintain.

Another major change is to switch our Markdown processing from using showdown.js to marked.js, this is because showdown is bidirectional (adding complexity), old, buggy and not well maintained - we already had several bits of code working around bugs in showdown.  We were also hand-rolling extra additions to support govuk table syntax and govspoeak syntax.

Using marked.js allows us to utilise its plugin architecture, and we are now utilising two x-gov packages to add govuk styles and support govspeak syntax:

https://github.com/x-govuk/govuk-markdown
https://github.com/x-govuk/marked-govspeak

Simpler and more maintained dependencies!

Because we are switching to single-direction markdown coversion, we need to be able to convert Govspeak-flavoured-markdown on the server.  For which we are now using the official alphagov/govspeak gem.  Unfortunately this has the side-effect of bringing the wholww of govuk-publishing-components with it 🙁.  This caused a few issues with sass compilation within the asset pipeline, but has been resolved by exxcluding the assets from the gem from compilation.

Other improvements include making the content areas expand/contract according to the content within them, and not selecting all the content on focus, and moving the cursor to the end on focus.


njseeto (approved):
I've had a read and I can't give an in-depth review but from what I can see it all looks good. Just a small comment about clean up but it's no biggie.

PR #2267, opened 2023-06-26
Added ccrc domain to the allowed list.

mattempty:
This will allow email addresses tp be added as a reply address for that domain,


H3ll3m4 (approved):
🚀

PR #2276, opened 2023-06-27
Use correct service slug for pingdom alerts

njseeto:
### Use the service slug in service configuration to create Pingdom alerts
When creating a pingdom alert, we need to use the service slug saved in the service configuration rather than using the service name.

### Use service configuration when un-publishing services
We now have the service slug saved in the service configuration. We should use this rather than the service name to identify the correct service slug to be unpublished.

Both these mechanisms are only reachable via the Admin dashboard.


H3ll3m4 (approved):
🚀

PR #2277, opened 2023-06-27
Add service_slug method as a helper

njseeto:
Helper method in the application controller didn't work because it it not namespaced with `admin`. Try making the helper method in the UptimeChecks controller to access it in the uptime checks template.


H3ll3m4 (approved):
🚀

PR #2300, opened 2023-07-06
Enable CSP and set to reporting mode only

njseeto:
As part of the recommendations from a recent pen test, we are enabling a CSP. For now, the policy will be on report mode only. We have a Sentry CSP URL that will report any CSP violations. Once we are happy that we have resolved all CSP violations, we can enforce the CSP.

Most of the inline JS scripts have had a nonce added, all inline styles have been moved into a utility class.

### DIFF
--- a/app/javascript/styles/_utilities.scss
+++ b/app/javascript/styles/_utilities.scss
@@ -5,3 +5,15 @@
 .text-success {
   color: govuk-colour('green');
 }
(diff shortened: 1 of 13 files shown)


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2301, opened 2023-07-06
Add hotjar to CSP

njseeto:
We use hotjar on the editor as to analyse user behaviour. We need to add these to our CSP to allow hotjar to function properly.


chrispymm (approved):

H3ll3m4 (approved):
🚀

PR #2311, opened 2023-07-07
Add `unsafe-inline` for hotjar inline scripts

njseeto:
As per hotjar docs, we need to add the `unsafe-inline` keyword for hotjar inline scripts to work.
[HotJar docs](https://help.hotjar.com/hc/en-us/articles/115011640307-Content-Security-Policies)


njseeto:
closing as we have removed hotjar widgets instead of adding this to the CSP.

PR #2327, opened 2023-07-17
Disable save and return on upload components

njseeto:
### Disable save and return on file upload components
This is a temporary measure to prevent a bug that allows files to be uploaded without a reference, which in turn creates an issue with the submission.

### Bump presenter 3.0.14

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'delayed_job_active_record'
 #     github: 'ministryofjustice/fb-metadata-presenter',
 #     branch: 'standalone-pages-notification-banners'
 # gem 'metadata_presenter', path: '../fb-metadata-presenter'
-gem 'metadata_presenter', '3.0.13'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -15,7 +15,7 @@ gem 'delayed_job_active_record'
 #     github: 'ministryofjustice/fb-metadata-presenter',
 #     branch: 'standalone-pages-notification-banners'
 # gem 'metadata_presenter', path: '../fb-metadata-presenter'
-gem 'metadata_presenter', '3.0.13'
+gem 'metadata_presenter', '3.0.14'
```
StevenLeighton21, line 18: It's safe to go to 3.0.15

StevenLeighton21 (dismissed):

Gemfile
njseeto: 👍  have bumped to 3.0.15

PR #2331, opened 2023-07-19
Temporary fix for browser-tools

njseeto:
The CircleCI team are currently working on a fix to make browser-tools compatible with the latest Chromedriver version (115).

Use v114 for now to ensure the pipeline is not blocked.


H3ll3m4 (approved):
🚀

PR #2353, opened 2023-08-02
Replace submitHandler.js with save-button component

chrispymm:
This uses the new save-button custom element used on the settings screens onto the question/edit pages.

* Adds `$form` getter to save-button to return a jQuery object of the form, for compatibility
* Updates `submit()` method to `save()` for better readability.
* `save()` function triggers a submit event, to ensure listeners fire
* Updates pages-controller to replace submitHandler with save-button
* Update the button in the view template

### DIFF
--- a/app/javascript/src/web-components/save-button.js
+++ b/app/javascript/src/web-components/save-button.js
@@ -74,6 +74,10 @@ class SaveButton extends HTMLButtonElement {
   get describedBy() {
     return this.form.querySelector(`#${this.getAttribute('aria-describedby')}`);
   }
+  
+  get $form() {
+    return $(this.form)
+  }
 
   get preventUnload() {
     return this.hasAttribute('prevent-unload');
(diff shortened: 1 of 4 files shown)


app/javascript/src/web-components/save-button.js
```diff
@@ -116,10 +120,10 @@ class SaveButton extends HTMLButtonElement {
 
   // Enables programatically submitting the form, regardless of the state of the
   // button, and bypassing unload prevention.
-  submit() {
+  save() {
     this.#enabled = true;
-    this.#removeBeforeUnloadListener();
-    this.form.submit(); // Does not raise a `submit` event
+    this.form.dispatchEvent(new Event('submit')) // Trigger submit event listeners
```
StevenLeighton21, line 125: Is this now a behaviour change? Looks like the old code didn't trigger the event? Or did it trigger in `submit()` so now we need to do it manually?

chrispymm: Which old code are you referring to? `SubmitHandler`, or the previous `save-button` implementation?

`SubmitHandler` used `$form.submit()` which calls jQuery's `submit()` function, which tirggered event listeners.

We now use `form.submit()` which is the browser native submit method which **does not** trigger event listeners.  (There is a newer `requestSubmit()` method, which calls the validation API and triggers listeners, however it's support only goes back one or two safari versions, which is not ideal so I didn't use it.)

The previous `save-button` implementation wasn't hindered by the event listeners not firing as there was nothing listening for submit events on settings screens.

StevenLeighton21 (approved):

PR #2381, opened 2023-08-24
Filter allowed govspeak components

chrispymm:
This PR limits our support for Govspeak to only this components we want form editors to be able to use.

We do this by only importing the extensions we want from marked-govspeak.

However, editors can still enter valid govspeak syntax into our contenteditable areas - they won't be parsed and rendered by marked however the syntax will remain and could be saved to the DB.  Then the runner **will** be able to render them as it uses the govspeak gem - which we have no control over.  This is fine for most things with the exception of buttons.  **We do not want to allow form editors to add buttons into content**

We have opted to downgrade any button syntax entered into a simple link.

So
```
{button}[I'm a button](http://google.com){/button}
```
will be converted into
```
[I'm a button](http://google.com)
```

To do this we apply filters to both the markdown parsed by marked using the preprocess filter so the preview renders correctly.  

We also use the same filters to pass the content through when saving the content to the backend.

### DIFF
--- a/app/javascript/src/web-components/editable-content.js
+++ b/app/javascript/src/web-components/editable-content.js
@@ -3,17 +3,6 @@ const govspeak = require('@x-govuk/marked-govspeak')
 const GovukHTMLRenderer = require('govuk-markdown')
 const DOMPurify = require('dompurify')
 
-DOMPurify.addHook("afterSanitizeAttributes", function (node, data, config) {
-  if (node.nodeName === "A" && node.classList.contains("govuk-button")) {
-    // Replace it with a link
-    // node.classList.remove("gem-c-button", "govuk-button", "govuk-button--start");
-    // node.classList.add('govuk-link')
-
-    // Or remove it (and its wrapper <p>)
-    node.parentNode.remove();
-  }
-});
-
 class EditableContent extends HTMLElement {
   constructor() {
     super() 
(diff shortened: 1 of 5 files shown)


app/javascript/src/web-components/editable-content.js
```diff
@@ -206,6 +245,23 @@ class EditableContent extends HTMLElement {
...
+    for( const [key, filterFunction] of Object.entries(this.markdownFilters)) {
+      markdown = filterFunction(markdown)
+    }
+    return markdown
+  }
+
+  sanitize(unsafeHTML) {
+    return DOMPurify.sanitize(unsafeHTML, {
+        USE_PROFILES: {html: true}, 
+        FORBID_TAGS: ['style', 'button']
```
StevenLeighton21, line 262: Not 100% sure I am not suggesting silliness here but could we also add 'script' to the FORBID_TAGS for added sanitization goodness?

StevenLeighton21 (approved):

app/javascript/src/web-components/editable-content.js
chrispymm: I'll read the docs again, but I think `<script>` is blocked by default as that is pretty much the amin thing you want to do with an html sanitizer.  TBF, we don't actually need the `<button>` tag in there anymore - that was my first attempt at filtering out buttons, but I forgot they're not actually button elements, just styled links.

chrispymm: Yep.  The profile 'html' includes a whole host of sanitization / xss prevention including removing `<script>` tags.  The `FORBID_TAGS` is then additive on top of that, specifying extra things to filter out.

PR #2403, opened 2023-09-07
Add waf firewall to the runner ingress

mattempty:
This change will be used for all runners


StevenLeighton21 (approved):

StevenLeighton21 (changes requested):
woops - test to update

PR #2422, opened 2023-09-21
Add Conditional Content UI 

chrispymm:
This PR adds in the Controllers and UI for conditional content areas.

However it does not yet use a version of the metadata presenter that supports conditional components.  

All of the UI and behaviours are behind a flag so nothing is exposed to the user, and is safe to ship.


acceptance/support/capybara.rb
chrispymm, line 1: This fixes the Deprecation warning when running acceptance tests.

Means that we can bump rspec/selenium - as the latest version enforces the deprecation, so failed pipelines and we had to revert.

app/javascript/src/index.js
```diff
@@ -89,8 +98,8 @@ switch(controllerAndAction()) {
     break;
 
   default:
-       console.log(controllerAndAction());
-       Controller = DefaultController;
+    //console.log(controllerAndAction());
```
StevenLeighton21, line 101: cheeky comment

app/services/content_visibility_creation.rb
```diff
@@ -0,0 +1,37 @@
...
+  attr_accessor :conditional_content, :latest_metadata
+
+  def save
+    return false if conditional_content.invalid? || conditional_content.any_errors?
+
+    create_version
+  end
+
+  def metadata
+    # conditional_content.conditionals.map{|c| c.to_metadata }
```
StevenLeighton21, line 16: cheeky comment :)

spec/models/conditional_content_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+  let(:previous_page) do
+    service.find_page_by_url('pets')
+  end
+  let(:previous_flow_uuid) do
+    previous_page.uuid
+  end
+  let(:current_page) do
+    service.find_page_by_url('furry')
+  end
+  # let(:conditional_content_uuid) { 'ac914bb1-c4b8-44a0-bacd-2d23a6597f1a' }
```
StevenLeighton21, line 22: 👀

app/javascript/src/index.js
chrispymm: yeah, I lef tthis deliberately as its helpful to just uncomment when you need debugging

StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2435, opened 2023-09-27
Log if grid is being traversed or not

StevenLeighton21:


StevenLeighton21:
Resolved by reducing complexity of the form, there isn't a quick fix and the publish page appears to be memoising what it can effectively, though rails re-renders at least once and appears to traverse twice

PR #2445, opened 2023-10-05
Add an approve to go live button to service admin

StevenLeighton21:
image

Approve for go live will add a service config flagging the service as approved, and revoke will flag a service as unable to publish to live. Either will also remove the other service config and will unpublish the service from the live env


H3ll3m4 (approved):
🚀

PR #2463, opened 2023-10-19
Testable conditional component

chrispymm:
This PR adds conditional content visibility features into the editor.

The ability to specify visibility conditions for content areas on:
* Content pages
* Exit pages
* Multi question pages
* Check your answers pages
* Confirmation pages

image

image

Before deleting a Page, Question or Question option, we check to see if it is used in any content visibility rules.

image

When previewing a single page with conditional content a notice is shown to inform the user that not all items shown on the preview will be seen when completing the form properly.

image


app/views/api/question_options/_delete_option_used_for_conditional_content_modal.html.erb
```diff
@@ -2,9 +2,20 @@
   <h3 data-node="heading" class="govuk-label govuk-label--m">
     <%= t('question_items.delete_modal.can_not_delete_heading') %>
   </h3>
-  <p data-node="content">
-    <%= t('question_items.delete_modal.delete_option_used_for_conditional_content_message') %>
-  </p>
+
+  <% if pages.size > 1 %>
```
StevenLeighton21, line 6: Just to check this logic is meant to be >1 and not >= 1

StevenLeighton21 (approved):

app/views/api/question_options/_delete_option_used_for_conditional_content_modal.html.erb
chrispymm: Yeah, `> 1` is correct.  If there is exactly one then the content is different.

For one:
You need to update {page.heading} first

For multiple:
You need to update the following pages:
* page 1 heading
* page 2 heading

If there are no pages, then this modal will never show, sop the distinction is between 1 and many

chrispymm: In hindsight switching the conditions and having `== 1` and `else` might have been more readable... but thats not a biggie!

PR #2477, opened 2023-10-30
Testable publish to live

StevenLeighton21:
a-cool-form has been published before, so does not see the check
image

a-new-form has not been
image

image

image


app/controllers/publish_controller.rb
```diff
@@ -4,9 +4,12 @@ class PublishController < FormController
   def index
     @published_dev = PublishServicePresenter.new(publishes_dev, service)
     @published_production = PublishServicePresenter.new(publishes_production, service)
+    declarations
   end
 
   def create
+    return unless can_publish_to_live || publish_service_params[:deployment_environment] == 'dev'
```
StevenLeighton21, line 11: Added this just to prevent hitting the route with a payload to bypass the publishing restriction.

zheileman (dismissed):
LGTM (big PR to review, would be best to have smaller ones next time 😄)
Just a few minor code style suggestions, no blockers.

H3ll3m4 (approved):
🚀

PR #2486, opened 2023-11-03
Testable form stats in admin

chrispymm:
Update the stats in the overview section of the admin area to no include forms created by the team, and to split out forms currently published, and a total count of all forms published.

Also fixed the CSV download while I was there

### DIFF
--- a/app/controllers/admin/overviews_controller.rb
+++ b/app/controllers/admin/overviews_controller.rb
@@ -1,6 +1,7 @@
 module Admin
   class OverviewsController < Admin::ApplicationController
     require 'csv'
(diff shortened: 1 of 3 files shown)


app/controllers/admin/overviews_controller.rb
```diff
@@ -1,6 +1,12 @@
 module Admin
   class OverviewsController < Admin::ApplicationController
     require 'csv'
+    ACCEPTANCE_TEST_USER_ID = 'a5833e7a-a210-4447-904c-df050d198e33'.freeze
```
StevenLeighton21, line 4: minor picky comment but can these consts be accessed from one location so we don't have to remember to maintain this list too? I'm sure they're used in the cronjob etc.

StevenLeighton21 (approved):

PR #2501, opened 2023-11-14
Always reference the same deployment env version of the SERVICE_SLUG 

StevenLeighton21:
When we create or update in the settings we always do both dev and production
image

But we don't consistently reference a specific one when we reference this config, which leads to a rare edge case when you've changed the slug value and republished but then attempt to create a pingdom alert.


zheileman (approved):
Good catch!

H3ll3m4 (approved):

PR #2508, opened 2023-11-20
Testable flow view refactor

chrispymm:
This PR is primarily a refactor of the `services/_flow_layout.html.erb` partial.

This single file was responsible for rendering pretty much the entire flow view and contained the markup and logic for displaying every different kind of flow object, as such it was pretty complex and hard to follow, with a bunch of the markup actually residing in the application helper file.

## Headline changes:
1) Refactor to use ViewComponents - this breaks out each type of flow object into its own component all inheriting form a base `FlowItemComponent` this allows for shared logic and behaviours.  
2) Move `flow_branch_link`, `flow_text_link` and `flow_item_thumbnail` from `ApplicationHelper` into their own components. N.B. `flow_item_thumbnail` is still in the helper file as it is used by standalone pages - this is future work to refactor the standalone pages.
3) Change `flow_layout.html.erb` partial to render the flow view and detached flows (i.e. the whole 'flow'.  This matches the name of the partial better (before it actually rendered flow_items) but having this in a partial allows for a future where we can relaod this partial instead the whole page (if we can get turbo drive working!)

### Hints for review
* You should be able to look through the removed code in flow layout and see the equivalent markup/behaviours implemented within the components.
* You should see that the code from the application_helper has fairly directly moved to the `FlowBranchLink`, `FlowItemLink` and `FlowItemThumbnail` components

### Still to do
One advantage of doing this is that it is possible to test the components directly, so it might be good to introduct some component tests to ensure the logic.  Although this should mostly be covered by acceptance tests... so may not bring much benenfit?

Could probably tidy up the branching point template a bit more

Look at the standalone pages, and allow them to use the `FlowItemThumbnail` component


acceptance/features/edit_single_question_autocomplete_page_spec.rb
```diff
@@ -77,7 +77,7 @@ def and_I_go_to_the_page_that_I_edit(preview_form)
   def then_I_should_see_my_changes_in_the_form(preview_form, change)
     within_window(preview_form) do
       page.find('.autocomplete__input').click
-      sleep(1) # allow time for page to load
+      sleep(5) # allow time for page to load
```
StevenLeighton21, line 80: Does this alleviate anything? Just thinking if 1 didn't work I wouldn't _expect_ 5 to, but if it (even anecdotally) does then cool

StevenLeighton21 (approved):

acceptance/features/edit_single_question_autocomplete_page_spec.rb
chrispymm: Hard to say... the test had failed repeatedly until I made this change...  but that could be dumb luck!

I guess we'll find out... this is *the* single flakiest test in the editor suite, so it should be realtively obvious fairly quickly if it helps or not.

I'm also dubious, theoretically the `find` on the previous line should wait up to `max_wait_time` (20s?) for the input to be available...  but maybe the input is available but JS hasn't kicked in yet so we need the extra sleep?  I'll be honest, I didn't think much about this change! 😂 I just got pissed off enough to go and look at the test, saw a `sleep` and increased it! 

Maybe the 'proper' solution would be for the previous line to search for a css class that is only present once the autocomplete JS has done its thing...

StevenLeighton21: Ship it and see if it is anecdotally less flaky then, it won't add too long to the run time

PR #2526, opened 2023-11-30
Flow view accessibility and screenreader improvements

chrispymm:
A whole bunch of improvements to the flow view for accessibility / screen reader users

- Page menu buttons are now labelled ‘Actions for page {title}’
- Branching point menu buttons are labelled ‘Actions for Branching point X’
- Connection menu buttons are labelled ‘Actions for connection between pages {title} and {title}’
- Flow view pages now include types for screenreader - they are announced e.g. ‘What is your name, single question page’
- Flow view item titles are now `<H2>`s, and branch conditions are nested `<H3>`s 
- Branch conditions are now links that jump you to the branch destination

All of the above vastly improves both the tabbing experience, but also allows screenreader users to make use of the rotor for listing links, headings, buttons, form controls for quick navigation around a large flow

Other changes

- Tab order should be logical (up to first branch)
- Using the VO control combo will now open a menu and focus the first item.
- VO control will also open a submenu too now (and focus the first item)
- Improved the mouseover timeouts on the submenu, so it is easier to move the mouse at an angle to get to a submenu item without the submenu disappearing


H3ll3m4 (approved):
🚀

PR #2530, opened 2023-12-06
Handle a nil item passed to FlowItemComponent

chrispymm:
The recent refactor of the flow view removed a `nil` check for items in a flow.  It turns out that there could be nil items in a flow, so this can cause the flow view to break.

Investigation revealed that the `nil` entries were introduced in the `detached_flows` method in the `pages_flow` model. 

This was caused by the code checking for the existence of multiple pointers in a detached flow.  When it found multiple pointers it called `next` within the `map` introducing a `nil` item.  

Using `filter_map` removes out the `nil` entries from the `detached_flows` object before it reaches the view.

### DIFF
--- a/app/components/moj_forms/flow/flow_item_component.rb
+++ b/app/components/moj_forms/flow/flow_item_component.rb
@@ -1,6 +1,7 @@
 module MojForms
   module Flow
     class FlowItemComponent < GovukComponent::Base
+      with_collection_parameter :item
       delegate :service, to: :helpers
       delegate :payment_link_enabled?, to: :helpers
       attr_reader :item, :html_attributes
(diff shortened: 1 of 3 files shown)


app/models/pages_flow.rb
```diff
@@ -18,7 +18,11 @@ def detached_flows
     detached = Detached.new(service:, main_flow_uuids: grid.flow_uuids)
     detached.detached_flows.map do |detached_flow|
       detached_flow.map do |column|
-        column.map do |flow|
+        column.filter_map do |flow|
```
zheileman, line 21: Don't you want to rename the `flow` variable? 😄

app/components/moj_forms/flow/flow_item_component.rb
```diff
@@ -14,6 +14,8 @@ def initialize(item:, classes: [], html_attributes: {})
       end
 
       def call
+        return if item.blank?
```
zheileman, line 17: We probably don't need this anymore?

zheileman (approved):

PR #2531, opened 2023-12-07
Fix detached flow nils and mulitple pointers

chrispymm:
This PR is a resolution to an issue that initially was casued by there being `nil` objects in the flow view - which causeed drendering to fail.

On investagting, there was code in the `detached_flows` method, which called `next` within a `map` introducing nils.

The purpose of this code was to resolve an issue with multiple pointers in detached flows.  It was introduced in this PR https://github.com/ministryofjustice/fb-editor/pull/2117

However, this code worked for the case used in the PR, but didn't work in more complex detached flows containing branches.

The solution to that lay in the metadata_presenter, where the code to trim pointers didn't work correctly for forms lacking branching.  

This PR merges the fixes from the metadata_presenter, and removes the code from the editor attempting to fix the pointers, but introducing `nil`s

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,11 +5,11 @@ ruby '3.1.3'
 
 # Metadata presenter - if you need to be on development you can uncomment
 # one of these lines:
-gem 'metadata_presenter',
-    github: 'ministryofjustice/fb-metadata-presenter',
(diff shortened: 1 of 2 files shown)


zheileman (approved):
🎉

StevenLeighton21 (approved):

PR #2535, opened 2023-12-13
Fix failing periodic rake tasks

zheileman:
There were a couple issues.

First one, these rake tasks for some reason require eager loading for the decryption to work correctly. By default rake tasks will not do eager loading. Console (in production) does eager loading, and that's why the tasks were working when invoked in the console.

The other issue was a performance issue. One rake task (`remove:test_services_configs`) has not been run successfully in such a long time that it accumulated thousands of records (well over 700k+) and the query to return them was not doing it in batches, which killed the process due to memory constraints.

The query has been optimised to use batches.

Both tasks have been run already so next time they run should finish fairly quick.

### DIFF
--- a/app/services/test_services_configs_remover.rb
+++ b/app/services/test_services_configs_remover.rb
@@ -7,7 +7,7 @@ class TestServicesConfigsRemover
 
   def call
     models.each do |model|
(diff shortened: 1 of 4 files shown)


StevenLeighton21 (approved):
Beautiful 😭

PR #2538, opened 2023-12-14
Try different approach for the encryption

zheileman:
~Trying to avoid having to declare encryption variables when they are not necessary (some cronjobs don't need them) by moving from using a constant to a method.~

It seems we must continue using a constant, as for some reason doing lazy assignment (via a method) produces the infamous `ActiveSupport::MessageEncryptor::InvalidMessage` again.

To minimise having to declare these encryption env variables in places where they are not even needed (because no encryption/decryption operations take place) I've resorted to a fetch with fallback.
The rake eager loading is still needed tho.

Tested each of the cronjobs and are working fine, included the sessions trimmer.
Should have no effect on the web containers either.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,8 @@ yarn-debug.log*
 .env
 .env.acceptance_tests
 
+coverage
+
 # Visual Studio Code #
 **/.vscode/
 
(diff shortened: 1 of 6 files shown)


StevenLeighton21 (approved):
Nice - I like this fix

PR #2540, opened 2023-12-20
Add WAF to runner pods 

StevenLeighton21:
Enable WAF for published forms, without rule causing erroneous 403 responses


chrispymm (dismissed):

zheileman (dismissed):
👍

zheileman (approved):
👍

PR #2545, opened 2024-01-16
Allow publishing to live with submission warnings

StevenLeighton21:
We would previously have blocked publishing when missing CYA/Confirmation pages, but we now have both a valid use case to allow a form to publish without and pre-publishing checks in place to prevent cases where this is not valid.


H3ll3m4 (approved):
🚀

PR #2555, opened 2024-02-14
Testable admin users error

StevenLeighton21:


app/views/admin/application/show.html.erb
```diff
@@ -35,18 +35,24 @@ as well as a link to its edit page.
...
-
-      <dd class="attribute-data attribute-data--<%=attribute.html_class%>"
-          > <%= render_field attribute, page: page %></dd>
+    <% page.attributes.each do |title, attributes| %>
+      <fieldset class="<%= "field-unit--nested" if title.present? %>">
+        <% if title.present? %>
+          <legend><%= t "helpers.label.#{page.resource_name}.#{title}", default: title %></legend>
+        <% end %>
+        <% attributes.each do |attribute| %>
+          <dt class="attribute-label" id="<%= attribute.name %>">
```
zheileman, line 44: Do we want the `govuk-!-margin-top-0` class in this `dt` as before?

H3ll3m4 (dismissed):

app/views/admin/application/show.html.erb
StevenLeighton21: I've put it back in for completeness sake

zheileman (approved):

PR #2561, opened 2024-02-21
Export form feature summary as CSV

StevenLeighton21:
Generate a form summary on request for a given environment, outputs as a csv

This can then be used to extrapolate a lot of interesting starts like how many forms use a given component or enable a given feature.

In dev, I generated a report on 29 services in 2.6 seconds

The major expensive operation is that each form needs to have its metadata pulled from the api as a single request for the latest version, in live envs this is likely to be 10x slower, but should still be able to generate a report in under 1 minute.

Report generation is only accessible as a logged in administrator via the admin UI


app/controllers/admin/overviews_controller.rb
```diff
@@ -62,8 +62,52 @@ def export_services
       end
     end
 
+    def export_dev_form_summary
+      respond_to do |format|
+        summary = service_summary('dev')
+
+        format.csv do
+          csv_data = CSV.generate do |csv|
+            csv << ['Service id', 'Service name', 'Locale', 'Start pages', 'Confirmation pages', 'Check your answers pages', 'Single Question pages', 'Multiple Question pages', 'Exit pages', 'Standalone pages', 'Address components', 'Autocomplete components', 'Checkbox components', 'Content components', 'Date components', 'Email components', 'Upload components', 'Multiupload components', 'Number components', 'Radio components', 'Text input components', 'Textarea components', 'Confirmation email enabled', 'Save and return enabled', 'Collect data via email', 'Send to JSON api', 'Receive csv', 'External start page enabled', 'Analytics enabled', 'Reference number enabled', 'Payment link enabled']
```
zheileman, line 71: This array of strings and the one a bit below seems identical to me? Maybe can be extracted to either a constant or a method to DRY hehe.

app/controllers/admin/overviews_controller.rb
```diff
@@ -90,6 +134,81 @@ def services_to_export
...
+
+      published_service_ids.each do |id|
+        metadata = MetadataApiClient::Service.latest_version(id)
+        result << {
+          service_id: id,
+          service_name: metadata['service_name'],
+          locale: metadata['locale'],
+          page_counts: page_type_counts(metadata['pages']),
+          component_counts: component_type_counts(metadata['pages']),
+          confirmation_email_enabled: ServiceConfiguration.find_by(service_id: id, deployment_environment: env, name: 'CONFIRMATION_EMAIL_COMPONENT_ID').present?,
```
zheileman, line 149: A suggestion for performance.

Instead of finding each individual configuration, we can retrieve all of them at once with:
`features = ServiceConfiguration.where(service_id: id, deployment_environment: env).pluck(:name)`

Then we just check if the array of features contains the one we want:
`features.include?('CONFIRMATION_EMAIL_COMPONENT_ID')`

app/controllers/admin/overviews_controller.rb
```diff
@@ -90,6 +134,81 @@ def services_to_export
...
+          save_and_return_enabled: ServiceConfiguration.find_by(service_id: id, deployment_environment: env, name: 'SAVE_AND_RETURN').present?,
+          collect_information_by_email_enabled: ServiceConfiguration.find_by(service_id: id, deployment_environment: env, name: 'SERVICE_EMAIL_BODY').present?,
+          send_to_json_api_enabled: ServiceConfiguration.find_by(service_id: id, deployment_environment: env, name: 'SERVICE_OUTPUT_JSON_ENDPOINT').present?,
+          receive_csv_enabled: ServiceConfiguration.find_by(service_id: id, deployment_environment: env, name: 'SERVICE_CSV_OUTPUT').present?,
+          external_start_page_enabled: ServiceConfiguration.find_by(service_id: id, deployment_environment: 'production', name: 'EXTERNAL_START_PAGE_URL').present?,
```
zheileman, line 154: Here there is a hardcoded value, is it on purpose?
`deployment_environment: 'production'`

app/controllers/admin/overviews_controller.rb
```diff
@@ -90,6 +134,81 @@ def services_to_export
...
+          payment_link_enabled: ServiceConfiguration.find_by(service_id: id, deployment_environment: env, name: 'PAYMENT_LINK').present?
+        }
+      end
+
+      result
+    end
+
+    def page_type_counts(pages)
+      {
+        start_pages: pages.select { |p| p['_type'] == 'page.start' }.count,
```
zheileman, line 166: I believe here we can do a similar strategy, because despite not triggering DB queries, this many select and count will not be optimal.

Idea would be to obtain an array with all types of pages, and do a tally to obtain the count of each type:
Example:
```
['page.start', 'page.singlequestion', 'page.singlequestion', 'page.exit'].tally
=> {"page.start"=>1, "page.singlequestion"=>2, "page.exit"=>1}
```

app/controllers/admin/overviews_controller.rb
```diff
@@ -90,6 +134,81 @@ def services_to_export
...
+      multiuploads = 0
+      numbers = 0
+      radios = 0
+      texts = 0
+      textareas = 0
+
+      pages.each do |page|
+        next if page['components'].blank?
+
+        addresses += page['components'].select { |c| c['_type'] == 'address' }.count
```
zheileman, line 193: Probably same as above, if possible to use tally.

zheileman (commented):
A few suggestions! looking great 🥳

app/controllers/admin/overviews_controller.rb
```diff
@@ -62,8 +62,52 @@ def export_services
       end
     end
 
+    def export_dev_form_summary
```
zheileman, line 65: Not a biggie but these 2 methods feel like could be just one that receives a couple arguments? 😄

zheileman (approved):
🚢

H3ll3m4 (approved):
🚀

PR #2565, opened 2024-02-28
Add command all in one line

zheileman:
Getting `/bin/bash: line 10: --verbose: command not found` in circleci when tests fail, if they pass we don't get the error, so this is weird.

I'm literally following circle ci documentation here: https://circleci.com/docs/use-the-circleci-cli-to-split-tests/#ruby-rspec-tests

Seems like adding all in one line fixes the issue...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,9 +112,7 @@ jobs:
       - run:
           name: Running the tests
           command: |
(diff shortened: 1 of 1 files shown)


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2567, opened 2024-02-28
Add move component function to multiquestion pages

chrispymm:
This PR adds the functionality to move components on a multiquestion page.

Functionality is added using a couple of stimulus controllers to handle adding the behaviour, buttons and events.  An outline of the behavior is as follows:
* Each component on the page has an  orderable-item controller with as stimulsu `order` value
* The orderable item controller injects the move buttons into the component
* The move buttons move the components in the DOM, which triggers a `reorder` event on the parent `orderable-items` controller
* Changing the order value triggers an `orderUpdated` event on each component
* Which is received by the `component-controller` which updates the components data and triggers a `saveRequired` event, updating the state of the save button on the page.

The visiblity of the buttons is handled in css.

Other things in this PR:
* Broke out a few chunks of css into separate files
* Improved the tests for external start page to not be flaky 🤞


Gemfile
```diff
@@ -44,6 +45,7 @@ group :development, :test do
   gem 'brakeman'
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
   gem 'capybara'
+  gem 'capybara_accessible_selectors', git: 'https://github.com/citizensadvice/capybara_accessible_selectors', branch: 'main'
```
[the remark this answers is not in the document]
chrispymm: Sadly, that is how it is used :-(, for whatever reason it hasn't been published as a gem - it is however a citizensadvice repo, so should be relatively trustworthy and unliklely to disappear... 

It's a really helpful bunch of selectors though as it allows lots of nice accessbility assertion within capybara

Can live without it though if we think its not a good idea to include

up_arrow_blue.svg
zheileman, line 1: I think this may be an unused file? 😄

package.json
```diff
@@ -1,45 +1,45 @@
...
-    "@sentry/webpack-plugin": "^1.20.0",
-    "chai": "^4.3.7",
-    "form-data": "^4.0.0",
-    "jquery-ui": "^1.13.2",
-    "jsdom": "^20.0.3",
-    "jsdom-global": "^3.0.2",
-    "mocha": "^10.2.0",
-    "sinon-chai": "^3.7.0"
-  }
+    "name": "fb_editor",
```
zheileman, line 2: Unintentional indentation maybe? Can't really see if something else has changed!

zheileman (approved):
Approving, just 2 minor things I noticed.

acceptance/pages/editor_app.rb
```diff
@@ -131,9 +136,9 @@ def flow_thumbnail(title)
     preview_page_images.find { |p| p.find('.title').text.include?(title) }
   end
 
-  def external_start_page_thumbnail
-    page.find('.external-thumbnail')
-  end
+  # def external_start_page_thumbnail
```
StevenLeighton21, line 139: can delete this now I guess?

app/javascript/packs/application.js
```diff
@@ -19,16 +19,25 @@ import ConditionalsController from "../src/controllers/conditionals-controller.j
...
 import ExpressionsController from "../src/controllers/expressions-controller.js";
 import SelectionRevealController from "../src/controllers/selection-reveal-controller.js";
+import OrderableItemsController from "../src/controllers/orderable-items-controller.js";
+import OrderableItemController from "../src/controllers/orderable-item-controller.js";
+import ComponentController from "../src/controllers/component-controller.js";
 
 Turbo.session.drive = false;
 window.Stimulus = Application.start();
+
+// Stimulus.debug = true;
```
StevenLeighton21, line 29: 👀

Gemfile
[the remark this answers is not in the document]
StevenLeighton21: Is there a release or tag we can pin to? We'll get changes if they merge to main is the only risk, so could be bumped up a major non compatible version

up_arrow_blue.svg
chrispymm: It is used - due to the way the icon buttons work, we need three files for each icon, a black, white and blue one.  

The better solution would be to have the svg inline and use css to change the color, but for *waves hands vaguely* _reasons_ we have to do it this way. (More seriously, I think its roughly to do with alignment and accessibility)

H3ll3m4 (approved):

package.json
chrispymm: This is probably becasue I made a manual edit and my editro has reformatted on save (to use 4 spaces apparently, for some reason)

app/javascript/packs/application.js
chrispymm: Yep - I only discovered this while doing this feature - its fairly handy - stops me having to drop in `console.log('x controller connected')` all over the place to work out the order of things!

Gemfile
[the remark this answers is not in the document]
chrispymm: 👍 pinned it to the latest tag `v0.11.0` from January this year

StevenLeighton21 (approved):

PR #2568, opened 2024-02-29
Revert pointer-events change for invisible buttons

chrispymm:
As part of the move work, the css declaration `pointer-events: none` was added to the three dots and move buttons when invisble in order to prevent the cursor becoming a pointer and the user being able to interact with a 'hidden' button.

However, this change made the buttons unclickable in Safari.  

The buttons display when the editable component title is focused, clicking the button unfocuses the title, which hides the button, which safari now treats as unclickable.  

It appears that Safari is executing the `blur` event before the `click` event, meaning the button gets hidden and rendered non-interactive before the click event is fired.  

This PR simply reverts the behaviour to how it was before, in order to allow the editor to function correctly in Safari.  A more complete fix will hopefully follow

P.S. Also removed an extraneous svg that ended up in the root of the project 😊

### DIFF
--- a/app/javascript/styles/_editable_components.scss
+++ b/app/javascript/styles/_editable_components.scss
@@ -18,7 +18,7 @@ editable-content,
   .ActivatedMenuActivator {
     left: -54px;
     opacity: 0;
-    pointer-events: none;
+    //pointer-events: none;
     position: absolute;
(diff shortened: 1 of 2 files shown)


zheileman (approved):
👍

PR #2569, opened 2024-02-29
Use presenter version 3.3.21

StevenLeighton21:
Resolves a bug where content wouldn't show on pages displaying validation errors

Editable
image

Preview
image


H3ll3m4 (approved):
🚀

PR #2570, opened 2024-02-29
More complete fix for button issue in safari

chrispymm:
This PR implements a fix to the click event issue on component buttons in safari.  

As mentioned in #2568 this was caused by the `blur` event triggering before the `click` event.  It seems that this is per-spec behaviour and Safari was behaving as it should, and Chrome and Firefox shouldn't have worked.

The solution here was to switch from `click` event to `mousedown` event, as the mousedown event will fire before the click event, meaning the button is still visible and interactivew hen the `click` event fires.

Apologies for the diff: I made changes in files that haven't been edited for a while, so they've been reformatted by prettier.

### DIFF
--- a/app/javascript/src/components/menus/activated_menu.js
+++ b/app/javascript/src/components/menus/activated_menu.js
@@ -47,356 +47,354 @@
  **/
 
 const {
-    property,
-    mergeObjects,
-    createElement,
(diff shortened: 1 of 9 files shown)


zheileman (approved):
🚀

PR #2576, opened 2024-03-04
Ensure orderableItemControllers property is optional

chrispymm:
One last tiny bugfix I spotted after the quickfix earlier today.

There are times when the `orderableItemControllers` can be `nil` or `unidentfied` which triggers an error.  This happens when the components area is empty, so it has no user impact, but we should prevent the error anyway.


H3ll3m4 (approved):
🚀

PR #2589, opened 2024-03-25
Add automated accessibility tests

chrispymm:
This PR adds a set of automated accessibility tests for every page in the editor.

A bunch of these are marked as either `pending` or `skip` as they currently fail.  The aim is that I can work through these to ensure every page is passing a basic accessibility audit.

The CircleCI config has been modified to exclude the accessibility test - this is to ensure that it doesn't fail the deploy at this point.  Even when these tests are complete and passing, we may not want an accessibility fail to block deploy, so we may need to aadjust the pipeline config to handle this test spec differently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -323,8 +323,7 @@ jobs:
             echo 'export CI_MODE=true' >> $BASH_ENV
             source $BASH_ENV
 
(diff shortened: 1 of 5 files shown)


H3ll3m4 (approved):
🚀

PR #2593, opened 2024-04-04
Single question page labelling

chrispymm:
This is a rather large PR, that updatest he single question pages to have good accesible labelling for all elements.

The code for this gets a little gnarly as we have to remove the default semantics and labelling asscoiations for the form elements as used in the runner and replace them with new labelling for the editor context.

There are three main things happening:
1. Editable `h1` elements now have a child `<span>` which is used for all the editble semantics in order for the `h1` to retain heading semantics.
2. The question form inputs and buttons are now no longer `disabled` as that makes them invisible to AT, they are now `aria-disabled` so take part in the accessibility tree and can be focused, but not interacted with.
3. All elements now have labelsa nd/or descriptions for screenreaders.

This now means that running AXE checker on the single question pages reports no issues.

### DIFF
--- a/app/javascript/src/question_autocomplete.js
+++ b/app/javascript/src/question_autocomplete.js
@@ -13,26 +13,30 @@
  *
  **/
 
-
-const utilities = require('./utilities');
+const utilities = require("./utilities");
 const mergeObjects = utilities.mergeObjects;
-const Question = require('./question');
+const Question = require("./question");
 
 const SELECTOR_HINT = ".govuk-hint";
-const SELECTOR_LABEL = "label h1, label h2, legend h1, legend h2";
+const SELECTOR_LABEL = "label h1 span, label h2 span";
 
 class AutocompleteQuestion extends Question {
   constructor($node, config) {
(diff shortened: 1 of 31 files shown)


app/javascript/src/question_autocomplete.js
```diff
@@ -13,26 +13,30 @@
...
 
-
-const utilities = require('./utilities');
+const utilities = require("./utilities");
 const mergeObjects = utilities.mergeObjects;
-const Question = require('./question');
+const Question = require("./question");
 
 const SELECTOR_HINT = ".govuk-hint";
-const SELECTOR_LABEL = "label h1, label h2, legend h1, legend h2";
```
StevenLeighton21, line 22: Why did we remove legend here? Because this component type doesn't have them?

StevenLeighton21 (approved):

app/javascript/src/question_autocomplete.js
chrispymm: yep, the autocomplete is a single field component, so it only has a label not a legend

PR #2613, opened 2024-04-18
Fix accessibility issues page editing

chrispymm:
This PR improves the accessible labelling on all other edit screens.  

Bumps to presenter 3.3.33 in order to use the template changes made there.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -58,7 +58,6 @@ Rails.application.configure do
   # Suppress logger output for asset requests.
   config.assets.quiet = true
 
-  config.hosts << 'b6a5-82-0-200-55.ngrok-free.app'
 
   # Raises error for missing translations.
   # config.i18n.raise_on_missing_translations = true
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -58,6 +58,8 @@ Rails.application.configure do
   # Suppress logger output for asset requests.
   config.assets.quiet = true
 
+  config.hosts << 'b6a5-82-0-200-55.ngrok-free.app'
+
   # Raises error for missing translations.
   # config.i18n.raise_on_missing_translations = true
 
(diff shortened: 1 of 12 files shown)


config/environments/development.rb
```diff
@@ -58,6 +58,8 @@
   # Suppress logger output for asset requests.
   config.assets.quiet = true
 
+  config.hosts << 'b6a5-82-0-200-55.ngrok-free.app'
```
StevenLeighton21, line 61: Does this need to stay long term?

spec/spec_helper.rb
```diff
@@ -96,42 +96,42 @@
...
-  enable_coverage(:branch)
-end
-
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new(
-  [
-    SimpleCov::Formatter::HTMLFormatter,
-    SimpleCov::Formatter::Console
-  ]
-)
+# SimpleCov.start do
```
StevenLeighton21, line 99: No more simplecov is a good thing in my book but I guess it's unintentional here? :D

config/environments/development.rb
chrispymm: ah no, sorry, probably me doing an inadvertant commit all 
This was when I was trying to do screenreader testing through a VM, which couldn't access my local network, so was using ngrok as a proxy.
I'll remove it

spec/spec_helper.rb
chrispymm: Haha - yeah I comment it out locally cos it means I can't see the actual test results!
This was proabbly also a result of my inadvertant commit all, but in this case I thought I'd pushed the fix before requesting review.  Apparently not!

StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #2615, opened 2024-04-19
Transfer ownership setting 

H3ll3m4:
[Trello](https://trello.com/c/v9EnzMIb/4035-add-transfer-of-ownership-setting-to-the-editor)

## Context
This is part of the work on transfer of ownership where a form author can change ownership and nominate a new owner.

## Implementation
This PR adds the setting of change form owner in the list of settings, this is only accessible with a feature flag locally and in test. It adds a controller, model and view. We validate the setting to ensure the email is not blank, match a email regex and the user is present in the User table. We then update the metadata and redirect the user to their list of forms. We also show a modal with a confirmation of transfer of ownership.

## Tests
We have added unit tests and manual testing have been performed. Mob testing & acceptance tests will follow.

### DIFF
--- a/app/controllers/settings/form_owner_controller.rb
+++ b/app/controllers/settings/form_owner_controller.rb
@@ -1,0 +1,12 @@
+class Settings::FormOwnerController < FormController
+  def index; end
+
+  def update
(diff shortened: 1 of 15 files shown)


StevenLeighton21 (approved):
Couple of questions in slack but more around design, looks great 👍

PR #2616, opened 2024-04-19
Testable editable elements spans

chrispymm:
This PR modifies the `EditableElement` class to insert a child span inside the element and use that as the `contentEditable` container.  

This is to allow headings to retain their semantics and to fix the fact that in radios and checkboxes making `<label>` elements `contentEditable` fails WCAG as they then have an invalid `role` of `textbox`.

- Manual testing seems to show this as all working
- JS tests have been updated to pass

### To do:
Acceptance tests are still failing

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,11 +5,11 @@ ruby '3.1.3'
 
 # Metadata presenter - if you need to be on development you can uncomment
 # one of these lines:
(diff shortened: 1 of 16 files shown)


StevenLeighton21:
Closed as not progressing to clean up testable editors

PR #2619, opened 2024-04-24
Send email on first publish to test (featured-flagged)

zheileman:
https://trello.com/c/te7rYMDe

Now that we have the Notify integration in place, we can hook the sending of the email and the rules for this to trigger.

Given we really don't want this email to be sent if the service couldn't be published due to some configuration error, or k8s commands error, it will be triggered as an `after_perform` callback on the `PublishServiceJob`, ensuring it only gets called if the job has finalised successfully.

Additionally this new service object has been called `AfterCompletion` as there might be other code we want to run here as well, but for now is only the email.

Note it is still behind feature flag so will not trigger on live editor, only in test/testable editors.

### DIFF
--- a/app/services/publisher/adapters/cloud_platform.rb
+++ b/app/services/publisher/adapters/cloud_platform.rb
@@ -59,14 +59,6 @@ class Publisher
 
       private
 
-      def first_published?
-        PublishService.completed
-        .where(
-          service_id:,
-          deployment_environment:
-        ).count == 1
-      end
-
       def create_config_dir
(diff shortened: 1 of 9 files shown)


app/services/publisher/adapters/cloud_platform.rb
```diff
@@ -59,14 +59,6 @@ def completed; end
 
       private
 
-      def first_published?
```
zheileman, line 62: Removed this method as it was not being used anywhere, but it gave me the clue how to do it to see if the email must trigger or not 😅

PR #2627, opened 2024-05-02
Transfer ownership confirmation email

H3ll3m4:
[Trello](https://trello.com/c/TnIvOouf/4036-add-confirmation-email-of-transfer-of-ownership)
[Trello](https://trello.com/c/gR65jWpd/4095-implement-new-template-for-transfer-of-ownership)

We are adding the confirmation email for the transfer of ownership feature. The form owner controller send information like previous owner, new owner and service name and we are using GovUK Notify as mailer. 

We have decided to add another template for the form owner rather than sending the same email because Notify doesn't allow recipient in cc.

Just added some unit tests, we will do acceptance tests in the next PR and remove the feature flag.


config/locales/en.yml
```diff
@@ -57,6 +57,17 @@ en:
     reference_payment_sentence: "\r\nYour reference number is: {{reference_number}}.\r\nTo complete your submission, you now need to make the required payment at {{payment_link}}."
     service_output_json_endpoint: 'Enter the endpoint URL here'
     service_output_json_key: 'Enter the endpoint key here'
+    transfer_ownership_email:
```
StevenLeighton21, line 60: These localisations aren't used now with Notify I think?

config/initializers/filter_parameter_logging.rb
```diff
@@ -4,5 +4,5 @@
 # sensitive information. See the ActiveSupport::ParameterFilter documentation for supported
 # notations and behaviors.
 Rails.application.config.filter_parameters += [
-  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn, :email, :name, :username, :password, :authenticity_token
+  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn, :email, :name, :username, :password, :authenticity_token, :access_key, :secret_access_key
```
StevenLeighton21, line 7: Probably leave those in there just in case, we could also add the params being sent around for new/previous owner emails

StevenLeighton21 (approved):

app/models/form_owner_settings.rb
```diff
@@ -21,16 +21,19 @@ def update
...
       @version = version
     end
+    true
   end
 
   private
 
   def email_exists?
-    @new_user_id = User.all.map { |user| user.id if user.email == @form_owner }.compact.first
+    @new_user_id = User.all.map { |user|
```
StevenLeighton21, line 34: It probably won't make a noticeable performance difference but this will map over and call a regex match on every row in the user table, I think it would be more performant to offload the match onto the database with something like `User.where('lower(email) = ?', @form_owner.downcase).first`

I don't think that the User table is ever going to be big enough that this is an issue though

StevenLeighton21 (approved):

app/models/form_owner_settings.rb
zheileman: I was about to say the same, and then sadly realised the user email is encrypted so we can't run DB queries 😢 

But I think this alternative might be a bit more performant and easier to read?

```ruby
User.all.detect { |u| u.email.downcase == @form_owner.downcase }.present?
```

PR #2628, opened 2024-05-03
Move back link outside of <main> element

chrispymm:
This PR fixes an issue rails in our DAC accessibility audit regarding the location of the back link on our settings screens in the editor.

The back link was within the `<main>` element meaning it was not bypassed by the 'Skip to content' link. This is annoying for users using that link as they will repeatedly here it announced as the first element.

It has now been moved outside of the `<main>` element.

While working on this PR I realised that we have two layout files for the editor `application.html.erb` and `form.html.erb` and they only differed in their inclusion of the form navigation partial (and the GA tags - which I assume is an error).

I've refactored so the app uses 1 layout and provided a helper method for determining if the form_navigation partial should be included.

This has been created as a testable and the full AT test suite passes.


app/components/moj_forms/settings_screen_component.rb
```diff
@@ -1,6 +1,6 @@
 module MojForms
   class SettingsScreenComponent < GovukComponent::Base
-    renders_one :back_link, MojForms::BackLinkComponent
+    # renders_one :back_link, MojForms::BackLinkComponent
```
StevenLeighton21, line 3: do we still want this?

chrispymm: No, we don't need it anymore.  I'll remove the comment

StevenLeighton21 (dismissed):

StevenLeighton21 (approved):

app/views/settings/confirmation_email/index.html.erb
```diff
@@ -1,10 +1,14 @@
+<% content_for :back_link do %>
+  <%= render MojForms::BackLinkComponent.new(href: settings_submission_index_path) do %>
+    <%=t('actions.back')%><span class="govuk-visually-hidden">to <%=t('settings.submission.heading')%></span>
```
zheileman, line 3: Not sure if it matters for screen readers but there is no space between 'Back' and 'to'?

app/views/settings/email/index.html.erb
```diff
@@ -1,10 +1,13 @@
+<% content_for :back_link do %>
+  <%= render MojForms::BackLinkComponent.new(href: settings_submission_index_path) do %>
```
zheileman, line 2: As there seem to be a bit of repetition with this code, maybe can be extracted to a partial that receives 2 locals as customisation, one for the href of the back link, and another for the visually hidden copy. But no biggie!

app/views/settings/confirmation_email/index.html.erb
chrispymm: 👀 good spot! But spaces actually get added in `::before` and `::after` pseudo elements with the `.govuk-visually-hidden` class.

zheileman (approved):
🚀

PR #2630, opened 2024-05-03
Add visually hidden text to dismiss button of announcement banner

chrispymm:
This PR adds some visually hidden text to aid assistive technology users browsing out of context understand the dismiss button for the announcement banner.


zheileman (approved):
Thanks!

PR #2631, opened 2024-05-03
Testable page titles

StevenLeighton21:
Adds descriptive page titles that match headings to user facing pages


app/controllers/branches_controller.rb
```diff
@@ -70,6 +70,15 @@ def branch_detached_destinations
   helper_method :branch_destinations
   helper_method :branch_detached_destinations
 
+  def page_title
+    if @branch
+      "Edit branch - #{@branch.title} - MoJ Forms Editor"
```
chrispymm, line 75: Isn't `@branch.title` always 'Branch X`?

In which case would this be better as:
`"Edit #{@branch.title} - MoJ Forms"`

zheileman (approved):

chrispymm (approved):
LGTM 🚀

PR #2634, opened 2024-05-09
Show black border and yellow focus ring for ediatble components

chrispymm:
This PR improves color contrast on focus for editable elements.  

The yellow border alone does not have a high enough contrast ratio to pass WCAG AA.

The focus state has been updated to add a black border in addition to the yellow focus ring to match GOV.UK design system form inputs.

**Before:**
image

**After:**
image


zheileman (approved):

H3ll3m4 (approved):
🚀

PR #2638, opened 2024-05-10
Test using default session trim with a threshold of 1 day

StevenLeighton21:


StevenLeighton21:
Closing as I want to restart from a different approach, and closing is a quick easy way to remove the testable and revert the job change

PR #2648, opened 2024-05-15
Skip acceptance tests Notify mail deliveries

zheileman:
https://trello.com/c/OoqBKxm5

Although it is only a nuisance, every time the acceptance tests run there are some tests that will attempt to trigger Notify emails, but the API key for the test environment does not allow sending to the email recipient and returns an error that gets reported to Sentry.

This PR will intercept emails unless we are on `formbuilder-saas-live`. If the recipient is the acceptance tests user, will stop the delivery, meaning we will not call the Notify API so no error will be produced.

### DIFF
--- a/app/lib/host_env.rb
+++ b/app/lib/host_env.rb
@@ -12,7 +12,7 @@ module HostEnv
     def env_name
       return LOCAL if Rails.env.development? || Rails.env.test?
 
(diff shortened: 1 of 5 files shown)


StevenLeighton21 (approved):
I like this approach 👍

StevenLeighton21 (approved):

PR #2670, opened 2024-07-04
Test handle double quotes in user entered config values such as email…

StevenLeighton21:
… body

image


danielglen-moj (approved):

StevenLeighton21:
This PR raises alerts around incomplete substitution -

 - The values being substituted are already sanitized in the markdown -> html pipeline, and come from form owners.
 - The substitution is done with gsub so will substitute values globally
 - It's done to prevent double quotes in the user entered values from breaking the yaml file they're injected into

PR #2671, opened 2024-07-11
Re-enable dependabot

StevenLeighton21:
While the platform was actively developed dependabot tended to create more noise than value, but now we're no longer in active feature development it may be helpful to keep on top of version updates.

Revert in the event of excessive noise!


mattempty (approved):
🚀

PR #2743, opened 2024-09-04
Sort the publish records by created at for currently published status

StevenLeighton21:
I think the order isn't guaranteed, or can end up with records returned by 'last' after being put in an array that isn't actually the _latest_


mattempty (approved):
🚀

PR #2767, opened 2025-03-27
Stop CSP from reporting violation from inline click handlers

StevenLeighton21:
Adds unsafe inline to script src - we know this is generally better avoided, but is now reporting on all inline onclick handler events which is every page edit.

It's so error event noisy as to cause more problems in missed real events without turning this off.
Also bump MDP as it contains a bugfix worth applying.


mattempty (approved):
🚀

PR #2776, opened 2025-04-22
Bump metadata-presenter for govuk crest

HettieS:


MeenaModhvadia (approved):
LGTM - bumping up versions  and branch name amended to new crest branch.

MeenaModhvadia (approved):
LGTM - adding sleeps to allow extra time for tests to complete

HettieS:
> LGTM - bumping up versions and branch name amended to new crest branch.

@MeenaModhvadia yes that's right, we set the version of the metadata data presenter to be the one with the updated crest. The preview section inside the editor can then access the new crest. I had to add some additional CSS styling to the editor to work in conjuction with the metadata presenter code.

I think the branch renaming you're referring to is where I pointed the editor to a test branch version of the metadata presenter - I did this to diagnose some issues. That code is commented out because it is only for testing purposes.

PR #2777, opened 2025-05-06
CDPT-2705: Increase CPU to reduce high pod count

HettieS:
Relates to https://dsdmoj.atlassian.net/browse/CDPT-2705

We are seeing a high resource count which is hitting our resource quota.

Try increasing CPU to prevent sensitive scaler.


njpreddy (approved):
LGTM

PR #2793, opened 2025-06-05
Capitalise justice addresses

HettieS:
EntraID has a nice quirk where it capitalises justice account email addresses. 

You need to mirror this capitalisation when you add the email to the admins array.


mattempty (approved):
🚀

PR #2815, opened 2025-08-14
CDPT-2970: Update rake task time to investigate alerts frequency

ripan:
- ```KubePodCrashLooping```  alert triggers everyday around from 5:00am to 5:30am. 
- Updating the cronjob time to check if rake task is causing the  ```KubePodCrashLooping```  alert


wilson1000 (approved):
Lovely 😄

PR #2820, opened 2025-09-15
CDPT-3082: Enable and fix acceptance tests

ripan:
- Acceptance tests are failing because of missing "Create a new form" button. To make it pass we need to display this button conditionally based on environment variable CREATE_FORM_ENABLED.
- "Create a new form" button needs to be enabled in test environment only.


mattempty (approved):
🚀

PR #2965, opened 2026-06-11
CDPTKAN-1049 MoJ Forms:  Rails upgrade to 8.1.3

tmckernan:
Upgrades the application from Rails 7.2.3.1 to Rails 8.1.3, along with several dependent gem upgrades required to maintain compatibility.                            
                                                                                                                                                                       
### Dependency changes                                                                                                                                           
                                                                                                                                                                       
  - Rails 7.2.3.1 → 8.1.3                                                                                                                                         
  - Administrate ~> 0.20.1 → ~> 1.0
  - govuk-components < 5.0.2 → ~> 5.13                                                                                                                                 
  - metadata_presenter 3.4.14 → 3.4.15                                                                                                                                 
  - view_component pinned to >= 4.0, < 4.2 (newly explicit)                                                                                                            
                                                                                                                                                                       
### Compatibility fixes                                                                                                                                            
                                                                                                                                                                       
  **ViewComponent initialize / super forwarding**                                                                                                                                                                                                                                                                                   

-   Rails 8.1.3 changes how keyword arguments are forwarded via super. All ViewComponent initialize methods that called bare super were passing their keyword args to    
-   ViewComponent::Base#initialize, which no longer accepts them. Fixed by changing super → super() across all affected components.                                                                                                                                                                                                          
-   ViewComponent Webpacker helper access   

**image_pack_tag**                                                                                                                              
                                                                                                                                                                  
-   image_pack_tag is no longer automatically available as an instance method inside ViewComponents. Added delegate :image_pack_tag, to: :helpers to the components that 
-   use it (ItemThumbnailComponent, ExternalStartPageThumbnailComponent, FlowItemComponent).                                                                             
                                                                                                                                                                
  **AnnouncementComponent template extraction**                                                                                                                                                                                                                                                                                              

-   Extracted the inline `erb_template` from AnnouncementComponent into a standalone `announcement_component.html.erb` file, which is the preferred approach in ViewComponent 4.x.

  **Administrate filters local**                                                                                                                                           
                                                                                                                                                                       

-   Administrate 1.0 templates now expect a filters local variable. The custom Admin::UsersController#index was missing this — added `filters = Administrate::Search.new(...).valid_filters` to bring it in line with the base controller.                                                                            
                                                                                                                                                              
  **Hidden field autocomplete attribute**                                                                                                                                  
                                                                                                                                                                       

-   Rails 8.1.3's f.hidden_field no longer adds `autocomplete="off"` automatically. Added it explicitly to the hidden fields in the move dialog partial. Updated the corresponding request spec expectations to reflect the new attribute order (value → autocomplete → type → name → id).                                                

                                                                                                                                                                       
  **GovukVisuallyHiddenHelper and GovukLinkHelper**                                                                                                                        
                                                                                                                                                                       

-   Explicitly included GovukVisuallyHiddenHelper in presenters that reference visually-hidden helpers, required by the govuk-components 5.x API.


ripan (approved):
LGTM 🚀

PR #3006, opened 2026-08-14
CDPTKAN-239: Retire webpacker and replace with shakapacker

ripan:
- Update node version to 22.18.0
- Update yarn version 1.22.22
- Replace webpacker with shakapacker for assets compilation


mattempty (approved):
🚀

